PR #10, opened 2020-06-16
change CONF_SEPARATOR from the default value to the key.

knbedgm:
I got the error:
Invalid config for [waste_collection_schedule]: [separator] is an invalid option for [waste_collection_schedule]. 
I looked into the code and noticed that you had the key accidentally set as default value instead of the key.


mampfes:
Oops. Thanks for sharing!

PR #42, opened 2021-04-02
Using MyMuell with jumomind sources

bjoernwolter:
I found out that you can also use the data of the mymuell.de App with the junomind source.

to get the needed city and area id i have modified the jumomind_de.py script to use it with mymuell
during my testing i got some cities which does not deliver street-data. The script will now quit with a message if you are using mymuell and answer of the "street-data" request is empty.


mampfes:
Thanks for your contribution. I found out how to deal with the empty street data requests: The city request already contains the area-id in this cases, therefore there is no need to query the street data. So we should now have a fully working MyMuell integration :-)

PR #47, opened 2021-04-18
Add attribute to choose GET or POST for ICS parameters

coccyx00:
I want to use a garbage service that is only accessible with POST parameters.  
So I added a new attribute in yaml to choose whether to use GET or POST.  GET is the default so there are no adaptions needed for other services.  

Additionally there is bit more debug logging when using in hass.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 *.swo
 .vscode/
 .mypy_cache/
+venv/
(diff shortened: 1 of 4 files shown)


mampfes:
Thanks for your contribution.

Small enhancement: You can set the option `year_field` to automatically replace the current year. So you don't have to adjust your configuration in 2022. See also the TEST_CASE example in the .py file.

PR #59, opened 2021-07-05
use wildcard for year in munich testcase

phixion:
added wildcard {%Y} for munich testcase removed unneeded bits from the url.

Not a biggie but a bit cleaner I think


mampfes:
Thanks for your contribution. I missed that point when I updated the Munich test case.

PR #60, opened 2021-07-09
Add cubefour awido

autinerd:
This adds the source "Cubefour AWIDO", which contains the schedules from about 30 cities and districts.


mampfes:
Thanks for your contribution. I removed the support to filter for fractions (because a source should always return all waste types) and also did some minor refactorings and renaming.

PR #78, opened 2021-09-17
Change shebang line to respect venvs

danielrheinbay:
To run this script in a virtual environment (venv), the shebang line needs to invoke /usr/bin/env instead of /usr/bin/python3 to make sure the python3 version from the venv (including its modules) is executed.


mampfes:
Thanks for your contribution.

PR #90, opened 2021-10-23
Add support for hygea.be

xorob0:
I'm not a python dev, feel free to criticise my code as much as necessary.


mampfes:
Thanks for your contribution. This is the 1st source for Belgium, which is great!

I refactored some minor things, esp. cities without a street index didn't work so far.

However, due to the fact that a big part of the logic to display the correct icons in the calendar of hygea.be is implemented as css class selectors, this source can't be perfect. This is really strange.

PR #100, opened 2021-11-05
Simple fix for Seattle Public Utilities

atkulp:
I stepped through and found the issues were just down to two URL's that had changed


mampfes:
Wow, thank you very much for your contribution. Very much appreciated.

PR #108, opened 2021-11-18
Add support for directly specifying prem_code for seattle_gov 

buckbanzai:
Adds support for manually specifying a `prem_code` in the configuration for the seattle_gov source, as automatic lookup can struggle with multi-tenant addresses sometimes (as can the city collection calendar website).

Documentation was updated to mention how to look this up manually, a test case with a manually specified prem_code was added and test_sources passed.


mampfes:
Thanks for your contribution.

PR #115, opened 2021-12-02
fix problem in ics source

jberrenberg:
When trying to prefetch the next year the component crashed when no
year_field was provided.
Additionally there was a syntax error in the fetching of next years ics.


mampfes:
Thanks :-)

PR #123, opened 2021-12-13
remove source oberhausen.de

buffcode:
City moved to AbfallNavi.de which is already a supported source.


buffcode:
New configuration template if anyone is interested:

```yaml
waste_collection_schedule:
  sources:
    - name: abfallnavi_de
      args:
        service: oberhausen
        ort: Oberhausen
        # fill out
        strasse: ""
        hausnummer: ""
```

mampfes:
Thanks for your contribution.

PR #130, opened 2021-12-30
add Landkreis Heilbronn in abfall.io wizard

dergeberl:
Thanks for that cool integration. 
I added Landkreis Heilbronn in abfall.io wizard.


mampfes:
Thanks for your contribution.

PR #133, opened 2022-01-02
Add NAWMA

TheNoctambulist:
Add support for the North Adelaide Waste Management Authority.
This services the northern suburbs of Adelaide, South Australia.


mampfes:
Thanks for your contribution.

PR #135, opened 2022-01-03
Add EGN-Abfallkalender

audacity363:


custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+
+HEADERS = {
+    "content-type": "application/x-www-form-urlencoded; charset=UTF-8",
+    "x-requested-with": "XMLHttpRequest",
+}
+
+_LOGGER = logging.getLogger(__name__)
+
+class Source:
+    def __init__(self, city, district, street, housenumber, days=20):
```
mampfes, line 32: Could you remove the `days` option and just return the entire period? Filtering is a functionality of the framework and shall not be done by a source.

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+        tag = soup.find("meta", {"name": "csrf-token"})
+        if tag == None:
+            return []
+        HEADERS["x-csrf-token"] = tag["content"]
+
+        post_data = urllib.parse.urlencode({"city": self._city, "district": self._district,
+            "street": self._street, "street_number": self._housenumber})
+        r = s.post(URL, data=post_data, headers=HEADERS)
+        data = r.json()
+        if "error" in data and data["error"] == True:
```
mampfes, line 63: `if data.get("error"):` ?

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+
+        for year, months in data["waste_discharge"].items():
+            for month, days in months.items():
+                for day, types in days.items():
+                    date = datetime.datetime(year=int(year), month=int(month), day=int(day)).date()
+                    if date > goal_date:
+                        break
+                    for type in types:
+                        color = data["trash_type_colors"][str(type).lower()] if data["trash_type_colors"][str(type).lower()] != None else f"Unkown ({type})"
+                        icon = self._iconMap[color] if color in self._iconMap else None
```
mampfes, line 77: icon = self._iconMap.get(color)

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+                        color = color.capitalize()
+                        entries.append(
+                            Collection(
+                                date, color, icon=icon
+                            )
+                        )
+                else:
```
mampfes, line 84: I think all these `else`, `continue`, `break `statement can be removed.

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+        
+
+        for year, months in data["waste_discharge"].items():
+            for month, days in months.items():
+                for day, types in days.items():
+                    date = datetime.datetime(year=int(year), month=int(month), day=int(day)).date()
+                    if date > goal_date:
+                        break
+                    for type in types:
+                        color = data["trash_type_colors"][str(type).lower()] if data["trash_type_colors"][str(type).lower()] != None else f"Unkown ({type})"
```
mampfes, line 76: `color = data["trash_type_colors"].get(type.lower(), type)` ?

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+
+import requests
+from bs4 import BeautifulSoup
+from requests.sessions import dispatch_hook
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "EGN Abfallkalender"
+DESCRIPTION = "Source for EGN Abfallkalender"
+URL = "https://www.egn-abfallkalender.de/kalender"
+TEST_CASES: Dict[str, Dict[str, str]] = {
```
mampfes, line 14: I'm not very familiar with typing, but the dict value you use for `days` is not a str. Is this data typing correct, therefore? And it would be nice if housenumber could be also specified as an int.

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
```diff
@@ -0,0 +1,91 @@
...
+        s = requests.session()
+        r = s.get(URL)
+
+        soup = BeautifulSoup(r.text, features="html.parser")
+        tag = soup.find("meta", {"name": "csrf-token"})
+        if tag == None:
+            return []
+        HEADERS["x-csrf-token"] = tag["content"]
+
+        post_data = urllib.parse.urlencode({"city": self._city, "district": self._district,
```
mampfes, line 59: Is `urllib.parse.urlencode` really needed?

audacity363: Addresses can contain special character which needs to be url encoded.

custom_components/waste_collection_schedule/waste_collection_schedule/source/egn_abfallkalender_de.py
audacity363: Implemented in the next commit

mampfes:
Thanks for your contribution.

PR #139, opened 2022-01-11
add Ludwigshafen am Rhein to abfall.io

FHeilmann:
Thanks for writing this integration! Here's a PR that adds my hometown (Ludwigshafen am Rhein).

Also added to wizard, one small detail I noticed is that the wizard terminates after querying `f_id_kommune` and `f_id_strasse`. Using only those parameters works, however it results in less events returned (164) compared to when `f_id_strasse_hnr` is added to the query as well (198). 

Bests,

Florian

Signed-off-by: Florian Heilmann <Florian.Heilmann@gmx.net>


mampfes:
Thanks for your contribution.

PR #142, opened 2022-01-12
Add wuerzburg_de

rikroe:
Adds waste collection for the city of Würzburg.

I know that `days` will be requiring a discussion: I have to call the API with both `ev[start]` and `ev[end]` dates.
Without them, nothing is returned. With just the start date, only event for this day are returned.
The API by default used approx. 1 year to look into the future.

I could also remove the parameter completely and hard-code the +1 year time interval.


mampfes:
Hi, thanks for the nice explanation regarding the days argument. Do you know the maximum time span of the provided data? In my experience, the schedules are provided on a per year basis. Therefore I would hardcode the one year time span to avoid bothering the user with a service provider specific parameter.

mampfes:
Thanks for your contribution.

PR #144, opened 2022-01-12
Add missing brackets

rikroe:
When looking through the code, I stumbled across this. Pretty sure, that it is just missing curly brackets.
Would be relevant if multiple scrapers are used with the same date & collection types.

Thanks for the cool integration, very much liked how easy it was to implement an additional source!


mampfes:
Good catch. Thanks!

PR #146, opened 2022-01-16
Add support for Christchurch City Council, NZ

codyc1515:
Also makes the other New Zealand plugins in the documentation more readable


mampfes:
Thanks for your contribution.

codyc1515:
Thank you for making it so easy to contribute. I haven't used Python in years but your developer documentation was so easy to follow I was done end-to-end in under 2 hours.

PR #152, opened 2022-01-20
Add support for Colchester.gov.uk for United Kingdom

Canis-L-Sapien:
Add support for waste removal the borough of Colchester, United Kingdom


Canis-L-Sapien:
Noticed bug with future collections

mampfes:
Hi, your service provider uses an interesting data format. Could you please add an description (either here or better directly in the source code) this magic with `WeekOne` true/false works?

Canis-L-Sapien:
@mampfes added some explanation in comments in the code

mampfes:
Thanks for your contribution.

PR #154, opened 2022-01-21
Update abfall_io.py

ranseyer:
Ostablkreis added... (Works for me!)


mampfes:
Thanks for your contribution.

ranseyer:
Thank you for the Project and the quick response...

PR #155, opened 2022-01-21
Update for ics.md

d3claes:
Hi,

I've added documentation for using the ics-file from service provider Limburg.net in Belgium with your HA-extension.

Greetings,

d3claes


mampfes:
Thanks for your contribution.

Issue #153, opened 2022-01-21
Debugging 'no components' error


geNAZt:
Hello,

i try to add "aha" as waste collection source. I use the ICS directly provided (they use a POST form to generate them) configured as such:

```yaml
# Waste management
waste_collection_schedule:
  sources:
    - name: ics
      customize:
        - type: Abfuhr Restabfall # from calendar event name
          alias: Restabfall # matches type in sensor configuration above
      args:
        url: https://www.aha-region.de/abholtermine/abfuhrkalender
        method: POST
        params:
          hausnr: "49"
          ladeort: "01916-0049+"
          strasse: "01916%40Marienstr.%40"
          ical_android: "10"
```

When i issue the same request from the host (with cUrl) via `curl 'https://www.aha-region.de/abholtermine/abfuhrkalender' --data-raw 'hausnr=49&ladeort=01916-0049+&strasse=01916%40Marienstr.%40&ical_android=10'` i get a valid? ICS back. I have attached the response ICS.

Currently i don't know how to debug the issue since i can't see the request in logs, can't see the response, can't see the parser state. I'm a bit lost and need help to solve this :)

Thanks
[waste.txt](https://github.com/mampfes/hacs_waste_collection_schedule/files/7912974/waste.txt)

mampfes:
I think the problem is caused by duplicate URL encoding. Try:
```yaml
      args:
        url: https://www.aha-region.de/abholtermine/abfuhrkalender
        method: POST
        params:
          hausnr: 49
          ladeort: "01916-0049 "
          strasse: "01916@Marienstr.@"
          ical_android: 10
```

Note the space at the end of `ladeort` and the `@` instead of `%40` for `strasse`.

geNAZt:
yeah that was it but it was not obvious and hard to debug since there are no requests/responses logged. having some sort of debug log would be neat (i ended up enabling httpclient debug and urllib3 loglevel to debug) to see requests/response to see those issues.

mampfes:
I always use the [test_sources.py](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py) which is part of the waste_collection_schedule component for debugging.

mampfes:
I have a request for you: Could you please describe in the ics.md how you got the source arguments? I think this would be very helpful for other people in your district. Thanks in advance.

PR #156, opened 2022-01-21
Clarify that params are urlencoded

geNAZt:
Params are in all cases urlencoded and passed to either the URL or HTTP body with additional headers. This documentation change now reflects this to avoid confustion if the configuration needs to be urlencoded or not. 

This should help removing confusion like happend in https://github.com/mampfes/hacs_waste_collection_schedule/issues/153


mampfes:
Thanks for your contribution.

PR #101, opened 2021-11-06
Add source for stonnington.vic.gov.au

tathamoddie:
This is a major council area in inner city Melbourne, Victoria, Australia.

PR #160, opened 2022-01-28
Added source for banyule.vic.gov.au

ravngr:
Adds a source for Banyule City Council, covering an area of North-East Melbourne, Australia.

Based quite heavily upon the implementation for Stonnington Council in #101; they use almost identical mechanisms for fetching upcoming dates. Only major change is a mechanism to bypass one of the lookups.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/banyule_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/banyule_vic_gov_au.py
@@ -1,0 +1,124 @@
+import logging
+import typing
+from datetime import datetime
+
+import requests
+from bs4 import BeautifulSoup
+from waste_collection_schedule import Collection
+
+
+TITLE = 'Banyule City Council'
+DESCRIPTION = 'Source for Banyule City Council rubbish collection.'
+URL = 'https://www.banyule.vic.gov.au/binday'
+TEST_CASES = {
+    'Monday A': {'street_address': '6 Mandall Avenue, IVANHOE'},
+    'Monday A Geolocation ID': {'geolocation_id': '4f7ebfca-1526-4363-8b87-df3103a10a87'},
+    'Monday B': {'street_address': '10 Burke Road North, IVANHOE EAST'},
+    'Thursday A': {'street_address': '255 St Helena Road, GREENSBOROUGH'},
+    'Thursday B': {'street_address': '35 Para Road, MONTMORENCY'}
+}
+
+_LOGGER = logging.getLogger(__name__)
+
+ICON_MAP = {
+    'green waste': 'mdi:leaf',
+    'recycling': 'mdi:recycle'
+}
+
+
+class SourceConfigurationError(ValueError):
+    pass
+
+
+class SourceParseError(ValueError):
+    pass
+
+
+class Source:
+    def __init__(self, street_address: typing.Optional[str] = None, geolocation_id: typing.Optional[str] = None):
+        if street_address is None and geolocation_id is None:
+            raise SourceConfigurationError('Either street_address or geolocation_id must have a value')
+
+        self._street_address = street_address
+        self._geolocation_id = geolocation_id
+
+    @property
+    def geolocation_id(self) -> str:
+        if self._geolocation_id is None:
+            # Search for geolocation ID
+            geolocation_response = requests.get(
+                'https://www.banyule.vic.gov.au/api/v1/myarea/search',
+                params={
+                    'keywords': self._street_address,
+                    'maxresults': 1
+                }
+            )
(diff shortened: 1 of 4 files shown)


ravngr:
Now @tathamoddie has completed an implementation for MRSC (#162) there are obviously a number of councils all using the same approach. I did some searching to see if they utilised some well-known software on the back-end, and they're all outsourcing to [Open Cities](https://www.opencities.com/). A quick search for "Powered by OpenCities" throws up many council websites in Victoria and NSW (at least). I was actually hoping for a nice API to be documented somewhere (parsing HTML wrapped in JSON is a little gross...), but had no such luck.

In the future it might be worth re-factoring the underlying parser (which I used quite extensively) into a separate service or abstract source, then passing the relevant URLs in the concrete implementation. It will make adding support for new councils much simpler. The only caveat might be parsing services in different regions, for example in Banyule there is booked hard rubbish rather than regular collections in other areas.

custom_components/waste_collection_schedule/waste_collection_schedule/source/banyule_vic_gov_au.py
```diff
@@ -0,0 +1,124 @@
...
+        pickup_entries = []
+
+        for element in calendar_parser.find_all('article'):
+            _LOGGER.debug(f"Parsing collection: {element!r}")
+
+            waste_type = element.h3.string
+
+            # Extract and parse collection date
+            waste_date_str = element.find(class_='next-service').string.strip()
+            waste_date = datetime.strptime(waste_date_str.partition(' ')[2], '%d/%m/%Y')
```
tathamoddie, line 116: I think you'll want a tighter check here.

See #161 which is an open PR of mine to fix a bug that I started hitting with Stonnington's implementation this week.

tathamoddie:
> I did some searching to see if they utilised some well-known software on the back-end, and they're all outsourcing to [Open Cities](https://www.opencities.com/). A quick search for "Powered by OpenCities" throws up many council websites in Victoria and NSW (at least).

Great find!

> In the future it might be worth re-factoring the underlying parser (which I used quite extensively) into a separate service or abstract source, then passing the relevant URLs in the concrete implementation.

Yep, sounds like a good next step for us to massively expand support here.

> The only caveat might be parsing services in different regions, for example in Banyule there is booked hard rubbish rather than regular collections in other areas.

In the differences between Stonnington and MRSC, I found that names were different (e.g. `Green waste` vs. `FOGO bin`), so that required a different icon map between the providers. Stonnington currently has no upcoming regularly scheduled hard waste, so it's basically on-demand right now, and my fix in #161 deals with this.

tathamoddie:
> I did some searching to see if they utilised some well-known software on the back-end, and they're all outsourcing to [Open Cities](https://www.opencities.com/). A quick search for "Powered by OpenCities" throws up many council websites in Victoria and NSW (at least).

Great find!

> In the future it might be worth re-factoring the underlying parser (which I used quite extensively) into a separate service or abstract source, then passing the relevant URLs in the concrete implementation.

Yep, sounds like a good next step for us to massively expand support here.

> The only caveat might be parsing services in different regions, for example in Banyule there is booked hard rubbish rather than regular collections in other areas.

In the differences between Stonnington and MRSC, I found that the names of collections were different (e.g. `Green waste` vs. `FOGO bin`, so that required a different icon map)

custom_components/waste_collection_schedule/waste_collection_schedule/source/banyule_vic_gov_au.py
ravngr: Yeah I saw your PR. For the short-term I don't think it's an issue since we don't have collections that haven't got a set date (at the moment) but it is a good sanity check regardless.

tathamoddie: That's how Stonnington worked until they broke it last week :wink:

ravngr:
> In the differences between Stonnington and MRSC, I found that names were different (e.g. `Green waste` vs. `FOGO bin`), so that required a different icon map between the providers. Stonnington currently has no upcoming regularly scheduled hard waste, so it's basically on-demand right now, and my fix in #161 deals with this.

Something I think a quick look up table can solve 😁 

The other corner case to consider is when public holidays come up. I think Banyule only kick the date forward for Good Friday and ANZAC day, but that may differ by region. I'm dreading them changing the text on those dates (eg. "pickup on the day after <date>"), but we unfortunately can't look into the future on those pages to check.

I was really hoping that by changing a magic value somewhere in the request you'd get JSON back from the schedule query, but it seems particularly fussy about the ```ocsvclang``` parameter being set and some combination of cookies.

mampfes:
Thanks for your contribution.

PR #162, opened 2022-01-30
add source for mrsc.vic.gov.au

### LINKED ISSUE #159: New Data source - Macedon Ranges, Victoria, Australia
Hoping there's an easy way to add my local council waste days as a data source.
There's an address lookup page here.
https://www.mrsc.vic.gov.au/Live-Work/Bins-Rubbish-Recycling/Bins-and-collection-days/Bin-collection-days

I've just got no idea about how to even start looking at being able to get the data to connect.

Thanks
Simon

tathamoddie:
Fixes #159 

This new provider is based on the [existing provider for Stonnington City Council](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/source/stonnington_vic_gov_au.py) as it looks like both councils are using the same publishing software, albeit with their own instances deployed.


mampfes:
Thanks for your contribution.

PR #181, opened 2022-03-20
Update ics.md by adding Bodenseekreis in Germany

snowstoked:
Splitted list of German counties by Bundesland and added Bodenseekreis (list and example configuration)


mampfes:
Thanks

PR #187, opened 2022-03-23
Added new data source for Ludwigsburg Germany

percidae:
I added the waste source for Ludwigsburg in Baden-Württemberg Germany.
The source is found here: https://www.avl-ludwigsburg.de/privatkunden/termine/abfallkalender/

I hope I did everything according to the instructions and it can be merged.


custom_components/waste_collection_schedule/waste_collection_schedule/source/avl_ludwigsburg.py
```diff
@@ -0,0 +1,143 @@
...
+        data.pop('tx_kesearch_pi1[sortByField]', None)
+        data.pop('tx_kesearch_pi1[sortByDir]', None)
+
+        # Depending on the city remove the street from the data set
+        if self._city.lower() not in cities_with_streets:
+            data.pop('tx_avlcollections_pi5[wasteCalendarStreetItem]', None)
+
+        # Creat the headers and cookies
+        headers = {
+            'Connection': 'keep-alive',
```
mampfes, line 62: Do you really need all these header? Please try to use only the absolute minimum.

custom_components/waste_collection_schedule/waste_collection_schedule/source/avl_ludwigsburg.py
```diff
@@ -0,0 +1,143 @@
...
+
+        if not r.ok:
+            raise Exception(
+                "Error: failed to fetch url: {}".format(
+                    url
+                )
+            )
+        
+        # Parse ics file, fix broken encoding
+        if r.encoding=="ISO-8859-1":
```
mampfes, line 135: Would r.encoding = "utf-8" also work? (Instead of the if else). I used this many times for other sources.

doc/source/avl_ludwigsburg_de.md
```diff
@@ -0,0 +1,151 @@
...
+          show: true
+        - type: Biomüll
+          alias: Biomüll
+          show: true
+        - type: Glas
+          alias: Glas
+          show: true
+```
+
+Use `sources.customize` to filter or rename the waste types:
```
mampfes, line 60: Is there a difference in the 2nd example compared to the first? If not, please remove it.

doc/source/avl_ludwigsburg_de.md
```diff
@@ -0,0 +1,151 @@
...
+**city**<br>
+*(string) (required)*
+
+**street**<br>
+*(string) (optional - depending on city)*
+
+## Example
+
+```yaml
+waste_collection_schedule:
```
mampfes, line 27: Please remove the customize part.

mampfes (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/avl_ludwigsburg.py
percidae: I tried it without that, didn't work. For some reason the webservice or the python requests module received ISO-8859-1 encoded data.

custom_components/waste_collection_schedule/waste_collection_schedule/source/avl_ludwigsburg.py
percidae: Fixed with 95428a33a4c24c68deb81e84c3c33d251a3cb123

doc/source/avl_ludwigsburg_de.md
percidae: Fixed with 99f76ee391ace6bda6a79fd79bc9672f1e57c4a8

doc/source/avl_ludwigsburg_de.md
percidae: Fixed with 99f76ee391ace6bda6a79fd79bc9672f1e57c4a8

custom_components/waste_collection_schedule/waste_collection_schedule/source/avl_ludwigsburg.py
mampfes: I'm not sure whether there is a misunderstanding. Therefore I repeat my point:

I know that python-requests sometimes has problems to detect the right encoding. But instead of the if/else statements, for me always a simple `r.encoding = "utf-8"` worked.

There are some examples in other sources, e.g.:

https://github.com/mampfes/hacs_waste_collection_schedule/blob/96f536ac74dc5d4acc1e3dbcb05879e47fcf23dd/custom_components/waste_collection_schedule/waste_collection_schedule/source/abfall_io.py#L127

Would that also work for you?

percidae: This might fix it for some cases but it assumes the format of input data which are in someone elses reponsibility. Therefore I used a more flexibel approach with 570fafb8d6ee72e27db6fa6178d10df7f9914348 to be prepared for different scenarios, without assuming the encoding of the data.

mampfes:
Thanks for your contribution.

PR #202, opened 2022-04-06
Update brisbane_qld_gov_au.py

trstns:
New event types have started being used, so this update makes sure to only consider recycle and organic events and ignore others (eg special information events).


mampfes:
Thanks for your contribution.

PR #203, opened 2022-04-06
Ipswich City Council, QLD, Australia Source

tonymyatt:
Add source for ICC.


custom_components/waste_collection_schedule/waste_collection_schedule/source/ipswich_qld_gov_au.py
```diff
@@ -0,0 +1,116 @@
...
+            self._load_date = False
+
+            items = data.strip().split("-")
+            self._loaded_date = datetime.date(int(items[0]), int(items[1]), int(items[2]))
+
+        if self._load_bin:
+            self._load_bin = False
+
+            if data == "Waste Bin":
+                self._entries.append(Collection(self._loaded_date, "Rubbish", icon="mdi:trash-can"))
```
mampfes, line 84: Please do not rename the original waste types. Aliases for waste types can be configured by user.
Use a dict to get the icon for known waste types.
Then we don't need the if statements any more :-)

PR #208, opened 2022-04-19
Add source wsz-moosburg.at

dweinber:
Add support for WSZ-Moosburg.at, including wizard to create the configuration.


mampfes:
Hi, thanks for your contribution, especially for providing a wizard. At the beginning of the project I thought it is a good idea to provide a wizard to 1. support users to get the right source parameter and 2. to relieve the source provides from unnecessary HTTP requests.
However, in the meantime I had to learn that a lot of users have problems with the wizard - mainly because the need a pure Linux box which they often do not have. Therefore I do not create wizards for new sources any more, but let the users configure the city, street, house number, etc. directly. This causes additional HTTP requests, but is much easier for a lot of users.

Therefore I would to ask if you could merge the wizard logic into the source code, so that the user only has to copy the parameters from the website (e.g. municipality) and don't have to use the separate wizard.

Thanks

dweinber:
Thanks for the feedback! In fact, I had everything in the source before and only later noticed the wizards and split it off 😅 Your reasoning definitely makes sense, I'll change that 👍

dweinber:
I removed the wizard script and incorporated all the requests into the source @mampfes

mampfes:
Thanks for your contribution. Please note that I had to rename - to _ in the file name to be consistent with other sources.

PR #217, opened 2022-04-26
Add source Campbelltown.nsw.gov.au

jameshirka:
Added basic support for Campbelltown NSW Australia.


mampfes:
Thanks for your contribution.

PR #229, opened 2022-05-05
Add calendar_title to unique_id

ejjoman:
This allows to use the same source for multiple calendars.


mampfes:
Could you please explain why you need the calendar title in addition? The unique-id already contains all the source arguments, therefore I supposed that this generates a unique-id, even if you use the same source multiple time (because the source-arguments differ).

ejjoman:
In my case, I have configured the source muellmax_de three times with the same arguments, but with different customization in order to have all different waste types in it's own calendar:


```yaml
waste_collection_schedule:
  sources:
    - name: muellmax_de
      calendar_title: Restmüll
      args:
        service: SERVICE
        mm_frm_ort_sel: CITY
        mm_frm_str_sel: STREET
      customize:
        - type: Restabfalltonne
          alias: Restmüll
          show: true # <-- only "Restabfalltonne" active
        - type: Biotonne
          alias: Biomüll
          show: false
        - type: Gelbe Tonne
          show: false

    - name: muellmax_de
      calendar_title: Biomüll
      args:
        service: SERVICE
        mm_frm_ort_sel: CITY
        mm_frm_str_sel: STREET
      customize:
        - type: Restabfalltonne
          alias: Restmüll
          show: false
        - type: Biotonne
          alias: Biomüll
          show: true # <-- only "Biotonne" active
        - type: Gelbe Tonne
          show: false

    - name: muellmax_de
      calendar_title: Gelbe Tonne
      args:
        service: SERVICE
        mm_frm_ort_sel: CITY
        mm_frm_str_sel: STREET
      customize:
        - type: Restabfalltonne
          alias: Restmüll
          show: false
        - type: Biotonne
          alias: Biomüll
          show: false
        - type: Gelbe Tonne
          show: true # <-- only "Gelbe Tonne" active

    - name: ics
      calendar_title: Altpapier
      args:
        file: "altpapier_2022.ics"

```

This produces this result in HA:
grafik

With a single source, all those different waste types would be in the same calendar.
Maybe there is another way to achieve this behavior?

mampfes:
That is an interesting use case...

Currently this use case is not supported, by I think this would be a useful extension to add to `waste_collection_schedule`. Adding calendar_title to the unique is is however not the right way because of multiple reasons:

1. Unique-ids should not contain names (see HA documentation).
2. Fetching is the same data multiple times is not nice.

I added an enhancement label and will look into it later.

PR #234, opened 2022-05-10
South Cambs service added

dt215git:
Added South Cambridgeshire UK service (scambs_gov_uk).
Added iCal example for South Cambs.


mampfes:
Thanks for your contribution.

PR #242, opened 2022-05-17
Add source belmont_wa_gov_au

robbo600:
Added source belmont_wa_gov_au.py
Added docs belmont_wa_gov_au.md
Updated info.md and README.md with Belmont City Council details

Apologies if this is not up to scratch, I'm not a developer by any stretch of the imagination.


mampfes:
Thanks for your contribution. I replaced the 4 address parameters by a single one, so that you can now just copy the address string from the web page.

PR #246, opened 2022-06-01
Fix Source banyule_vic_gov_au

ravngr:
The source I created accidentally passed datetime objects to the Collection constructor instead of Date objects. A quick fix for that.

While I wrote the source a while ago I didn't actually use it until recently. It "passed" ```test_sources.py``` since it quite happily dumps datetimes as strings 🙃


mampfes:
Thanks for your contribution.

mampfes:
Fixed test_sources.py in #247

PR #251, opened 2022-06-07
Add new ICS source Gemeinde Zorneding

FaserF:
I was looking everywhere to see my local waste collection in HA, thanks to your ICal variant I can give support for "Gemeinde Zorneding" and confirm its working.

Also to decrease searching for new users if their collection is supported via ICS/iCal I recommend adding a short list about all ICS/Ical supported variants on the main readme (done with this pull request).

Thank you for your great addon!


chrfritsch:
Awesome, thanks for this and greetings from Pöring ;)

PR #252, opened 2022-06-07
Add North Somerset council

andrewbeaton:
Adding of collection data for North Somerset council, covering Clevedon, Nailsea, Portishead and Weston super Mare.


mampfes:
Thanks for your contribution.

PR #260, opened 2022-06-21
Add ALW Wolfenbüttel as source

envy:
This adds support for https://www.alw-wf.de/.

They don't have a public API but they do have an iOS and Android app. They have "authentication" with the API there, but the credentials are not really secure und hardcoded into the app, so whatever.


mampfes:
Thanks for your contribution

PR #272, opened 2022-07-02
fix: bradford_gov_uk has an old TLS config not supported by openssl 3…

medains:
… without some special context options setting

Using an answer from here:
https://stackoverflow.com/questions/71603314/ssl-error-unsafe-legacy-renegotiation-disabled

This is potentially vulnerable to a MITM attack, but for this application it wouldn't seem like a big issue.


mampfes:
Thanks for your contribution.

Issue #185, opened 2022-03-23
Not Able to Pull data for Auckland


kiwilarka:
Hi, What a wonderful idea to automatically update it. Great concept. I think Documentation is lacking for new users. Probably would be ideal to have more step by step guides.

I have got the following config in YAML, took ages to generate it without errors. My sensors show up but they are blank, no information is getting pulled. Any help will be much appreciated.

`waste_collection_schedule:
  sources:
    - name: aucklandcouncil_govt_nz
      args:
        area_number: 12342560766
      customize:
        - type: brown bin
          show: no
        - type: Rubbish
        - type: Recycling
sensor:
  # next collection
  - platform: waste_collection_schedule
    name: next_collection
    
  # next yard waste collection
  - platform: waste_collection_schedule
    name: next_yard_collection
    types:
      - Rubbish
    
  # next garbage collection
  - platform: waste_collection_schedule
    name: next_garbage_collection
    types:
      - Recycling`

mampfes:
Try (note that yaml is case sensitive and that indentation counts!):

```yaml
waste_collection_schedule:
  sources:
    - name: aucklandcouncil_govt_nz
      args:
        area_number: 12342560766
      customize:
        - type: brown bin
          show: no

sensor:
  - platform: waste_collection_schedule
    name: next_collection
  - platform: waste_collection_schedule
    name: next_yard_collection
    types:
      - rubbish
  - platform: waste_collection_schedule
    name: next_garbage_collection
    types:
      - recycle
```

Could you explain the problems you had? This would be helpful to improve to documentation. Or even better: Provide a suggestion :-)

kiwilarka:
Sorry my code was right the formatting got stuffed when i pasted here. So its exactly same as you have mentioned above, configurations check out.
But when i insert the sensor in a button , it shows as Next Garbage collection (Unknown)
it means its not picking up the values from the website, how do i test or confirm that.
I will create a video for step by step to help you out. instead of suggestions.

mampfes:
I had to adjust `rubbish `and `recycle` to get it working! Please check if you use exactly the same values.

kiwilarka:
Hi, Thanks for the prompt reply, I used it exactly as mentioned above but when i put the sensor to home screen it has unknown value. How to Troubleshoot ?
how do i confirm if its picked up the value from that site ? and what variables or names its using.

Below is a copy paste from my Configuration.Yaml and attached is my button

````
waste_collection_schedule:
[...]

PR #278, opened 2022-07-06
Update aucklandcouncil_govt_nz.md

kellyatkinson:
Added an additional example as requested in https://github.com/mampfes/hacs_waste_collection_schedule/issues/185#issuecomment-1176028662

### DIFF
--- a/doc/source/aucklandcouncil_govt_nz.md
+++ b/doc/source/aucklandcouncil_govt_nz.md
@@ -27,10 +27,28 @@ waste_collection_schedule:
         area_number: 12342306525
 ```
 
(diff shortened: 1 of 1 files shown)


mampfes:
Sorry, I don't want to add individual customization to the generic examples.

mampfes:
Hm, maybe I was to fast with closing this PR. I just checked the discussion in #185. But I don't understand why this `customize` example is of generic interest. Even if you omit it, you should get valid data.

kellyatkinson:
No, I think it's fair to push back on this - happy to keep the documentation to the valid data only. Thanks!

PR #285, opened 2022-07-09
Fix wuerzburg_de

rikroe:
Fixes for `wuerzburg_de`:
- New URL
- HTML `SELECT`s now may contain line breaks, which the dict comprehension didn't like. Now checking for existence before accessing it in the dict comprehension.


mampfes:
Thanks for your contribution

PR #287, opened 2022-07-09
Add support for multiple calendars per source

ejjoman:


custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py
```diff
@@ -1,5 +1,6 @@
 #!/usr/bin/env python3
 
+from curses import use_default_colors
```
mampfes, line 3: Do you need this import or is this a debugging leftover?

mampfes:
Thanks for your contribution.

PR #289, opened 2022-07-11
Add source for Cochem-Zell

mirkolenz:
Add a new source for "Landkreis Cochem-Zell" based on their ICS calendar feed.


mampfes:
Thanks for your contribution.

PR #290, opened 2022-07-11
Add types to ICS parser and check for None values

mirkolenz:
During development of #289, I noticed that the `ICS` class currently has no type annotations. I made the following changes:

- To make it easier in the future to create new ICS-based sources, I added the correct types to the class.
- As a result of this, I was warned by VSCode that the value `dtstart` may be None. Thus, I added corresponding checks and made sure that only entries with a valid start date are returned by the `convert` function.
- I replaced the conditions of `type(e.start)` with corresponding `isinstance(e.start, ...)` calls to enable type inference in modern IDEs. Due to that, I had to change the order of the conditions because `datetime.datetime` inherits from `datetime.date`.
- `split_at` and `regex` were handled differently despite both representing a regular expression pattern. Thus, both are compiled to such a pattern in the init function.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
@@ -1,6 +1,7 @@
 import datetime
 import logging
 import re
+from typing import Any, List, Optional, Tuple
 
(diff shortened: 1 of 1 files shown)


mampfes:
Thanks for your contribution.

Issue #296, opened 2022-07-14
Error collecting data for Chesterfield.gov.uk


nathanael-pearson:
Hi

I'm getting the following error with the chesterfield.gov.uk integration...

fetch failed for source chesterfield.gov.uk: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py", line 122, in fetch entries = self._source.fetch() File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/chesterfield_gov_uk.py", line 54, in fetch "message": "{\"actions\":[{\"id\":\"4;a\",\"descriptor\":\"aura://ApexActionController/ACTION$execute\",\"callingDescriptor\":\"UNKNOWN\",\"params\":{\"namespace\":\"\",\"classname\":\"CBC_VE_CollectionDays\",\"method\":\"getServicesByUPRN\",\"params\":{\"propertyUprn\":\"" + self._uprn +"\",\"executedFrom\":\"Main Website\"},\"cacheable\":false,\"isContinuation\":false}}]}", TypeError: can only concatenate str (not "int") to str

Is this config or code?
Kind regards

dt215git:
Probably the code. I had this error a couple of times when composing the post request but thought I'd resolved it as it didn't happen during the subsequent testing.

It may be because on your system it's interpreting the uprn as a number rather than a string, but will take another look. 

If anyone else has any bright ideas feel free to chip in...

dt215git:
Yep, that'd do it.
Easy enough to update the .md to include the quote marks and fix my typo.

PR #300, opened 2022-07-15
typo corrected in chesterfield_gov_uk.md

dt215git:
Updated based on  #296


mampfes:
I added a conversion to string instead, to be tolarant for numerical uprn's. Therefore this PR is not needed any more.

PR #304, opened 2022-07-18
Added static source

ejjoman:
The static source allows to define static dates or a recurrence.
This is useful if the service provider does not offer an API, nor an ICS file.
It can also simplify the testing of automations by, for example, configuring a daily recurrence.


mampfes:
Thanks for your contribution.

PR #335, opened 2022-08-18
Source SSAM; Fixed problem with certain addresses

vasaru:
Certain addresses were not working properly. Updated code to reflect these different waste types and schedules.


mampfes:
Thanks for your contribution.

Issue #337, opened 2022-08-22
[bug] [recycleapp_be] UnboundLocalError: local variable 'streetId' referenced before assignment


algra4:
Hello, 

I have a problem with the recycleapp_be source.

I have the error into the log : 

```
Logger: waste_collection_schedule.scraper
Source: custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py:143
Integration: waste_collection_schedule ([documentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
First occurred: 15:24:43 (1 occurrences)
Last logged: 15:24:43

fetch failed for source Recycle!: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py", line 141, in fetch entries = self._source.fetch() File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/recycleapp_be.py", line 66, in fetch if streetId is None: UnboundLocalError: local variable 'streetId' referenced before assignment
` ``

PR #339, opened 2022-08-23
Resolve the bug #337

algra4:
This code resolve this problem #337


mampfes:
Thanks for your contribution.

PR #342, opened 2022-08-24
Add source for manchester city council.

thinkl33t:
Adds support for manchester city council's page.

https://www.manchester.gov.uk/bincollections/


mampfes:
Thanks for your contribution.

Issue #316, opened 2022-07-31
Calendar Error after 1.24.0 Update


VDRainer:
```txt
2022-07-31 12:19:17 ERROR (MainThread) [homeassistant.loader] Unexpected exception importing platform custom_components.waste_collection_schedule.calendar
Traceback (most recent call last):
  File "/srv/ha/lib/python3.9/site-packages/homeassistant/loader.py", line 618, in get_platform
    cache[full_name] = self._import_platform(platform_name)
  File "/srv/ha/lib/python3.9/site-packages/homeassistant/loader.py", line 635, in _import_platform
    return importlib.import_module(f"{self.pkg_path}.{platform_name}")
  File "/usr/lib/python3.9/importlib/__init__.py", line 127, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
  File "<frozen importlib._bootstrap>", line 1030, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1007, in _find_and_load
  File "<frozen importlib._bootstrap>", line 986, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 680, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 790, in exec_module
  File "<frozen importlib._bootstrap>", line 228, in _call_with_frames_removed
  File "/home/homeassi/.homeassistant/custom_components/waste_collection_schedule/calendar.py", line 118, in <module>
    def calc_unique_calendar_id(scraper: Scraper, type: str | None = None):
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
2022-07-31 12:19:17 ERROR (MainThread) [homeassistant.setup] Unable to prepare setup for platform waste_collection_schedule.calendar: Platform not found (Exception importing custom_components.waste_collection_schedule.calendar).
```

mampfes:
Please add your configuration and context information.

VDRainer:
```
waste_collection_schedule:
  sources:
    - name: abfall_io
      args:
        key: 3701fd1ff111f63996ab46a448669ea3
        f_id_kommune: 1069
        f_id_bezirk: 979
        f_id_strasse: 2890

sensor:
  - platform: waste_collection_schedule
    source_index: 0
    name: Abfall WaBue
    details_format: "appointment_types"
[...]

PR #346, opened 2022-08-27
remove python 3.10 type hinting

### LINKED ISSUE #345: Not using HA supported python version
HA supports the last two minor versions of Python, that is currently 3.10 and 3.9.
[calendar.py](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/calendar.py#L118) makes use of Python 3.10 only type hinting features for optional arguments via unions:
`def calc_unique_calendar_id(scraper: Scraper, type: str | None = None):`
The union str | None is not supported as type hint by Python 3.9, hence the waste collection schedule fails to load albeit HA runs on a supported installation.

mampfes:
fix #345

### DIFF
--- a/custom_components/waste_collection_schedule/calendar.py
+++ b/custom_components/waste_collection_schedule/calendar.py
@@ -1,15 +1,12 @@
 """Calendar platform support for Waste Collection Schedule."""
 
 import logging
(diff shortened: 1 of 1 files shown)


mampfes:
#316

PR #348, opened 2022-08-27
Update ics.md

iAmYves:
Minimal change to A.R.T to support Landkreis Trier-Saarburg (as Landkreis Trier-Saarburg has Gelber Sack & Altpapier combined on the same day)


mampfes:
Thanks for your contribution.

PR #354, opened 2022-09-04
adds ecoharmonogram

pawelhulek:
Signed-off-by: Pawel Hulek <pawelhulek@gmail.com>


custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
```diff
@@ -0,0 +1,85 @@
...
+}
+
+towns_url = "https://ecoharmonogram.pl/api/api.php?action=getTowns"
+scheduled_periods_url = "https://ecoharmonogram.pl/api/api.php?action=getSchedulePeriods"
+streets_url = "https://ecoharmonogram.pl/api/api.php?action=getStreets"
+schedules_url = "https://ecoharmonogram.pl/api/api.php?action=getSchedules"
+
+
+class Source:
+    def __init__(self, town_input, street_input, house_number_input):
```
mampfes, line 24: I would omit `_input` in the arguments.

doc/source/echoharmonogram_pl.md
```diff
@@ -0,0 +1,33 @@
...
+  sources:
+    - name: ecoharmonogram_pl
+      args:
+        town_input: Krzeszowice
+        street_input: Wyki
+        house_number_input: 
+```
+
+## Keep in mind
+Ecoharmonogram data is filled by each city, so that some towns doesn't require street name while some does.
```
mampfes, line 32: What should a user enter as street name if it not required? Could you make an optional argument out of it?

mampfes (changes requested):
Please add the new source to README.md and info.md in the top-level directory.

doc/source/echoharmonogram_pl.md
pawelhulek: If it will be null it will work just fine.
I would rather like users to fill all the fields as it will give them a guarantee that the integration works.
Also If they will leave for example street name empty the integration will work fine. Also worth mention that leaving street name empty is absolutely valid as some villages doesn't have streets names just a house numbers -

custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
pawelhulek: I would rather not like to remove `_input` prefix as in fetch method I am using variable names like `town`, `streets`, that may hard later on to figure out which is input and which is town object from the API response.

mampfes: Your argumentation is ok from developer perspective, but as a user I don't want to be bothered with this details. Also other sources don't use this convention. Therefore, please remove the _input from the arguments in the constructur (you can leave the internal variable names as they are - we don't have strict conventions for that - although most python code uses the prefix underscore).

doc/source/echoharmonogram_pl.md
mampfes: `null` means that the user has to enter the variable name, but leave the value empty, right? I think this is not what most users have in mind. I think they want to omit the variable name also, this is what I call an optional argument. Other sources also do it that way.

pawelhulek: ok, applied the fix

custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
pawelhulek: Ok, I can see it now. The issue was about consistent API between sources. Applied the fix. Thanks.

pawelhulek:
@mampfes Applied your comments - can't reask for review so I allowed myself to ping you directly

mampfes:
Thanks for your contribution.

PR #356, opened 2022-09-06
Added "LK Traunstein" to abfall.io script

abbe79:
Added the key to the wizard and a test case.
Hopefully I haven't forgotten anything else.

Thanks for this great project, finally I can throw away this "AbfallApp" 😁


mampfes:
Thanks for your contribution.

PR #360, opened 2022-09-07
Update info.md

tj551955:


mampfes:
Please do only one pull request per source, not one PR per file!

PR #365, opened 2022-09-09
Update minrenovasjon_no.py

randriksen:
Found some fractions in Rælingen kommune that gave an error because it couldn't find the correct icon.  Added the fractions to the icon map.


mampfes:
Thanks for your contribution.

PR #389, opened 2022-10-07
Add Newcastle-Upon-Tyne

SplinterHead:
Add a new source for Newcastle-Upon-Tyne, UK

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,3 +2,5 @@ beautifulsoup4==4.11.1
 icalendar==4.0.9
 icalevents==0.1.25
 recurring_ical_events==1.0.2b0
+requests==2.28.1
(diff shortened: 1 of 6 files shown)


requirements.txt
```diff
@@ -2,3 +2,5 @@ beautifulsoup4==4.11.1
 icalendar==4.0.9
 icalevents==0.1.25
 recurring_ical_events==1.0.2b0
+requests==2.28.1
```
mampfes, line 5: Sorry - I'm not the export for requirements.txt. Could it cause trouble with other integrations if we tag frequently used Python modules like `requests` with a fixed number?

mampfes:
Thanks for your contribution. I really appreciate it if you could support me with the Dependabot.

PR #394, opened 2022-10-13
Fix wuerzburg_de

tschlueter:
Fixed HTTP 403 for wuerzburg_de source-url: added missing HTTP header (user-agent)


mampfes:
Thanks for your contribution.

PR #397, opened 2022-10-17
Fix up for Wiltshire council to consider non-glass recycling

eli-randi:
Provides a small fix for Wiltshire Council, where the string doesn't always exactly match.


mampfes:
Thanks for your contribution.

PR #398, opened 2022-10-18
Add "Ostallgäu" to abfall_io.py

ShadNex:


mampfes:
Thanks for your contribution.

PR #400, opened 2022-10-20
Fix for request returning no data for warszawa19115

FliesWithWind:
It appears there was an update that broke this integration.
This fixes it.


mampfes:
Thanks for your contribution.

PR #362, opened 2022-09-08
Add source for Richmondshire, North Yorkshire

theanorak-keith:
Apologies if this is problematic, my first pull request.

PR #401, opened 2022-10-22
Missing richmondshire_gov_uk info added

dt215git:
I'm sure there's a better way of doing this, but here's the missing info for richmondshire_gov_uk:

- README.md
- info.md
- richmondshire_gov_uk.md

These, and the `richmondshire_gov_uk.py` file in #362, should complete that source request.


mampfes:
Thanks for your contribution.

Issue #386, opened 2022-10-06
Fault with Integration Chesterfield.gov.uk 


nathanael-pearson:
hi

This integration was working but is now showing the following error:Logger: waste_collection_schedule.scraper
I'm running the latest version of HA and the integration, I've tried rewriting the config and confirmed the UPRN is right.

--------------------------------------------------------------------------------------------------------------------------------------------------
Source: custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py:143

Integration: waste_collection_schedule ([documentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
First occurred: 12:04:19 PM (2 occurrences)
Last logged: 1:16:23 PM

fetch failed for source chesterfield.gov.uk: Traceback (most recent call last): File
"/config/custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py", line 141, in fetch entries = self._source.fetch() File 
"/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/chesterfield_gov_uk.py", line 74, in fetch waste_type = item["serviceTasks"][0]["taskTypeName"] IndexError: list index out of range
--------------------------------------------------------------------------------------------------------------------------------------------------

I've tried to run the debug test_sources.py but am getting access denied over SSH this script.

Kind regards

dt215git:
The query now seems to also be returning an entry for Commercial waste collection schedules, but for a residential property it's empty. That's probably triggering the IndexError, and can be caught with _try except_ statement.
After a quick tinker, here's the test output having inserted the _try except_, you can see the "Collection entry contains no data" message being printed when the IndexError is caught.

```
Testing source chesterfield_gov_uk ...
Collection entry contains no data
  found 3 entries for Test_001
    2022-10-21: Domestic Recycling [mdi:recycle]
    2022-10-21: Domestic Organic [mdi:leaf]
    2022-10-14: Domestic Refuse [mdi:trash-can]
Collection entry contains no data
  found 3 entries for Test_002
    2022-10-19: Domestic Recycling [mdi:recycle]
    2022-10-19: Domestic Organic [mdi:leaf]
    2022-10-12: Domestic Refuse [mdi:trash-can]
Collection entry contains no data
  found 3 entries for Test_003
    2022-10-17: Domestic Refuse [mdi:trash-can]
    2022-10-10: Domestic Recycling [mdi:recycle]
    2022-10-10: Domestic Organic [mdi:leaf]
[...]

PR #402, opened 2022-10-22
Further update to chesterfield_gov_uk

dt215git:
Pre-emptive fix for continued discussion in #386
Attempts to deal with periodically change in fwuid parameter


mampfes:
Thanks for your contribution.

nathanael-pearson:
Hi

I've just tested the pending change and it is, at least for now, working properly and I can see the next two collections scheduled in HA!

Thanks @dt215git

Issue #407, opened 2022-10-24
Issue to get the date for Recycle! / RecycleApp.be


Gunth:
Hi all,

Since today i have an issue when getting the dates for this service providers Recycle! / RecycleApp.be.

Here is the logs :

```
2022-10-24 15:28:18.281 ERROR (SyncWorker_4) [waste_collection_schedule.scraper] fetch failed for source Recycle!:
Traceback (most recent call last):
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py", line 141, in fetch
    entries = self._source.fetch()
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/recycleapp_be.py", line 54, in fetch
    headers["Authorization"] = r.json()["accessToken"]
KeyError: 'accessToken'
```

Version : 1.28.0

dt215git:
Looks like the API address has changed, the x-secret parameter is now different, and a POST is needed in one step rather than a GET.
I get this response have changed a few things in the recycleapp_be.py file, but don't know how it compares to what it used to produce - 366 results for one address seems a bit odd, but maybe they genuinely get a collection every few days.

```
Testing source recycleapp_be ...
  found 366 entries for 1140 Evere, Bazellaan 1
    2022-10-25: Household waste [None]
    2022-10-25: PMD [None]
    2022-10-28: Food waste [None]
    2022-10-28: Pruning waste [None]
    2022-10-28: Green waste [None]
    2022-10-28: Household waste [None]
    2022-10-28: Paper-cardboard [None]
    2022-11-01: Household waste [None]
    2022-11-01: PMD [None]
    2022-11-04: Food waste [None]
    2022-11-04: Pruning waste [None]
    2022-11-04: Green waste [None]
    2022-11-04: Household waste [None]
    2022-11-04: Paper-cardboard [None]
    2022-11-08: Household waste [None]
    2022-11-08: PMD [None]
    2022-11-11: Food waste [None]
    2022-11-11: Pruning waste [None]
    2022-11-11: Green waste [None]
    2022-11-11: Household waste [None]
    2022-11-11: Paper-cardboard [None]
    2022-11-15: Household waste [None]
    2022-11-15: PMD [None]
    2022-11-18: Food waste [None]
    2022-11-18: Pruning waste [None]
    2022-11-18: Green waste [None]
    2022-11-18: Household waste [None]
    2022-11-18: Paper-cardboard [None]
[...]

PR #408, opened 2022-10-24
recycleapp_be update

dt215git:
Updates as commented in #407:

API url updated to `https://api.fostplus.be/recycle-public/app/v1`
x-secret parameter updated
`{url}/streets` query change from GET to POST

Results for first test case seem odd, where there was always that many?


mampfes:
Thanks for your contribution.

jodewee:
thanks, works as a charm!

PR #411, opened 2022-10-26
Expanded search to cover next month and added garden waste as collection type

mcbalston:
Currently the source only returns collections for the current month. This fix adds collections for the next month when the current date is within 14 days of the month end.  This is done by utilising a couple of optional parameters in the API (specifying the month and year required).

Also adds the chargeable garden waste collection which is available for some properties.


mampfes:
There is a lot of code duplication. Could you move the fetching and appending part into a separate function (e.g. `fetch_month`) which takes a date (actually month + year) as argument?

mcbalston:
Good idea, thanks.  I'll make that change and resubmit.

mampfes:
Thanks for your contribution. I did the requested changes.

Issue #417, opened 2022-11-04
awb_es_de returning 0 results because of update for next year


meilon:
Since November 1st awb_es_de isn't returning any dates because AWB-ES is in the process of adding the ICS for next year which also is empty right now.

I switched to HA this year and I don't know how the ICS files were generated at the end of last year, but the current fix in my opinion would be to add a `break` after finding the first ICS link after line 44:
https://github.com/mampfes/hacs_waste_collection_schedule/blob/38b33c41cb9fee56f4657f467ade9faf476f2923/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py#L41-L44

This way there are at least dates until the 31.12.2022 available, maybe we need another fix if the homepage still lists the old 2022 ICS on 01.01.2023.

What are your opinions on this?

PR #418, opened 2022-11-04
Update awb_es_de.py

meilon:
Fixes the empty results in issue #417, might need a new fix when the year rolled over


mampfes:
Thanks for your contribution.

PR #420, opened 2022-11-04
Fix sample configuration for Recollect ics config

tobiasb:
Sample config wasn't valid yaml.

Offtopic: Great integration, thanks for building/maintaining this!!!


mampfes:
Thanks for your contribution.

Issue #423, opened 2022-11-08
How do I figure out why the collection is failing?


NZRobNZ:
I'm getting the following from the waste_collection_schedule.scraper - it appears to not like the way things are getting requested? "Max retries exceeded with url..."
Note, this is NOT for my address, I'm using the sample code for now until I can get the thing working...
```
This error originated from a custom integration.

Logger: waste_collection_schedule.scraper
Source: custom_components/waste_collection_schedule/waste_collection_schedule/scraper.py:143
Integration: waste_collection_schedule (documentation)
First occurred: 5:10:11 PM (1 occurrences)
Last logged: 5:10:11 PM

[...]

PR #424, opened 2022-11-08
Updated aucklandcouncil_govt_nz

dt215git:
Addresses Issue #423
Code snippet added to work around **SSL error unsafe legacy renegotiation disabled** server error.
Original code commented out, but not removed, just in case there is a need to revert at some point.


mampfes:
Thanks for your contribution.

Issue #427, opened 2022-11-12
I want to add support for infeo.at


dm82m:
Hey guys,

infeo.at provides a iCal file that changes from time to time. Till now I am using this component and manually change the iCal URL if the service changes (a few times a year). All this manual overhead is not what I want so I created a small python script that gets the current iCal URL based on the correct year and city. Is there any reference provider in in this plugin where a similar way (accessing a simple iCal source) is made? Don't want to have too much manual research, would be great if one can point me there.

Best,
Dirk

dm82m:
found everything and added first draft of PR

PR #428, opened 2022-11-12
#427: added support for infeo.at and tested with customer bogenschuetz-entsorgung.de

dm82m:


custom_components/waste_collection_schedule/waste_collection_schedule/source/infeo_at.py
```diff
@@ -0,0 +1,89 @@
...
+        entries = []
+
+        # loop over each calendar year entry and extract data
+        for calendarYear in response:
+            calendarYearId = calendarYear["id"]
+            calendarYearName = calendarYear["name"]
+
+            # get available zones for calendar year
+            url = f"{baseUrl}/zones/?calendarId={calendarYearId}"
+            response = requests.request("GET", url)
```
mampfes, line 48: Please pass URL parameters using the params argument instead of string concatenation: https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

Same is true for the following requests.

mampfes, line 48: Please use `requests.get` instead of `requests.request("GET", ...`

README.md
```diff
@@ -155,6 +158,7 @@ Currently the following service providers are supported:
 ### Switzerland
 
 - [A-Region.ch](./doc/source/a_region_ch.md)
+- [Infeo.at](./doc/source/infeo_at.md)
```
mampfes, line 161: Does infeo.at offer its services also in Germany and Switzerland? It looks a little bit strange if we have a `.at` domain in the Germany and Switzerland section. Therefore I would rather remove them here and just add a link once a concrete service provider is using their services - as you did with bogenschuetz-entsorgung.de

mampfes (changes requested):

mampfes:
fix #427

README.md
dm82m: yes they absolutely do: Bogenschütz is the only customer in Germany. All their other customers are in Austria and Switzerland, you can see them here: https://infeo.at/index.php/referenzen/

But okay, I will remove the global ones. My extension should work for them aswell but we can then if one wants to use another customer of infeo.at test everything out and put it officially online.

custom_components/waste_collection_schedule/waste_collection_schedule/source/infeo_at.py
dm82m: okay, I switched to params even if I don't like it that much cause it tends to more loc needed and for me personally the readability is not that good. but anyways: you own the repo, you make the decision.

dm82m:
@mampfes all requested changes done and PR is updated

dm82m:
@mampfes any idea when it will be merged and released?

mampfes:
Thanks for your contribution.

PR #431, opened 2022-11-13
adds districts to ecoharmogram

pawelhulek:
This MR allows user to narrow down town by district. Seems that there are towns with the same names supported by ecoharmonogram.

Signed-off-by: pawel.hulek <pawelhulek@gmail.com>


mampfes:
Thanks for your contribution.

PR #434, opened 2022-11-19
Added source for Sheffield City Council

NSF12345:
Added source for Sheffield City Council with support for all bin types.

I never got Dev Containers working to properly test, however I have manually added this as a source to my instance of Home Assistant/HACS and tested fine.


mampfes:
Thanks for your contribution.

PR #438, opened 2022-11-22
Fix example for Märkisch-Oderland

Rocka84:


mampfes:
Thanks for your contribution.

PR #444, opened 2022-11-25
Add source Erlangen-Höchstadt, Germany

fwinn:


custom_components/waste_collection_schedule/waste_collection_schedule/source/erlangen_hoechstadt_de.py
```diff
@@ -0,0 +1,38 @@
...
+        self._city = city
+        self._street = street
+        self._ics = ICS()
+
+    def fetch(self):
+        city = urllib.parse.quote(self._city.upper())
+        street = urllib.parse.quote(self._street)
+        today = datetime.date.today()
+        year = today.year
+        r = requests.get(
```
mampfes, line 29: Please pass arguments in URLs via the `params` argument: https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

I assume you can then also remove the urllib.parse.quote calls.

doc/source/erlangen_hoechstadt_de.md
```diff
@@ -0,0 +1,27 @@
...
+```yaml
+waste_collection_schedule:
+  sources:
+    - name: erlangen_hoechstadt_de
+      args:
+        city: CITY
+        street: STREET
+  separator: " / "
+```
+Note: Set the `separator` to `" / "` as it's the way different waste types are separated in one calendar event.
```
mampfes, line 15: I don't understand this not. `separator `is used to join waste types for the user interface. Therefore it should be irrelevant in the example. Why do you need it here?

mampfes (changes requested):

doc/source/erlangen_hoechstadt_de.md
fwinn: Because in the ICS Calendar, multiple waste types are displayed e. g. as "Restmülltonne / Biotonne" in one event. So you'll have to split them by configuring it via the `seperator`. Or did I understand it wrong?

custom_components/waste_collection_schedule/waste_collection_schedule/source/erlangen_hoechstadt_de.py
fwinn: Done, removing `urllib.parse.quote` worked as well.

doc/source/erlangen_hoechstadt_de.md
mampfes: `separator` is used to join waste types. In your case you have to split waste types. Please check the `split_at` argument of class `ICS`:
https://github.com/mampfes/hacs_waste_collection_schedule/blob/0f6d7b63cb35c63f7725c1d17b7ebae86c12a25c/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py#L16

fwinn: Okay, thanks, I understood and fixed it.

mampfes:
Thanks for your contribution.

PR #453, opened 2022-11-30
Update wiltshire_gov_uk.py to include Garden Waste

kitleyn:
Addition of Garden Waste collection schedule


kitleyn:
Closing this since I see that Marc has already added this and more in #411

PR #456, opened 2022-12-05
Add schedule for Horowhenua District, New Zealand

craSH:
* This largely takes previous work from @dt215git for the kuringgai_nsw_gov_au source which seems to use the same backend API as this district (among other AU districts also present in this repository)
* This provider currently only provides Rubbish/Garbage and Recycling services.


mampfes (changes requested):
Please add source also to info.md

craSH:
Done, thanks @mampfes

mampfes:
Thanks for your contribution.

PR #457, opened 2022-12-05
#427 reworked the documentation to make it clear and gave a few examples

dm82m:


mampfes:
Thanks for your contribution.

I have another suggestion: Many users have problems to find the correct source in a case like this where a web service provider (= infeo.at) is hosting the service for a waste collection company (= Bogenschuetz). I therefore tend to add the waste collection companies also in the documentation (README.md and info.md) and just link them to the markdown of web service provider. Is this also applicable here? If yes, could you please create a PR?

PR #464, opened 2022-12-11
added Source for Korneuburg [Austria]

53RT:
Added support Korneuburg in Austria

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
@@ -3,7 +3,6 @@ from urllib.parse import urljoin
 
 import requests
 from bs4 import BeautifulSoup
-from requests.adapters import HTTPAdapter, Retry
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 from waste_collection_schedule.service.ICS import ICS  # type: ignore[attr-defined]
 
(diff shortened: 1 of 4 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
```diff
@@ -0,0 +1,181 @@
...
+        entries = [Collection(d[0], d[1]) for d in dates]
+
+        return entries
+
+    def fetch(self):
+
+        ical_urls = self.get_region_links()
+
+        for u in ical_urls:
+            print(u)
```
mampfes, line 173: Remove debug output

custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
```diff
@@ -0,0 +1,181 @@
...
+    "Rathaus using Teilgebiet": {"street_name": "Some Street", "street_number": "1A", "teilgebiet": "4"},  # Teilgebiet 4
+    "Werft": {"street_name": "Am Hafen", "street_number": 6}                                               # Teilgebiet 2
+}
+
+
+def http_get_retry(params, n_retries=5):
+    """HTTP GET with n retries """
+    s = requests.Session()
+
+    retries = Retry(total=n_retries,
```
mampfes, line 33: Just to be curious: Is their service so unreliable that a special retry handling is required? I have never seen a need for this in any other source.

mampfes (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
53RT: Thanks. I have overlooked this one.

custom_components/waste_collection_schedule/waste_collection_schedule/source/korneuburg_stadtservice_at.py
53RT: When I was writing the code I often got 502 responses and this helped so I could run all the tests. 

Today the tests ran quite fast so I think it was a temporary issue. If you prefer I can remove this function.

mampfes: Yes, I would prefer to start without this special code. Please remove it. Thanks.

53RT: Okay. Done

mampfes:
Thanks for your contribution.

PR #465, opened 2022-12-12
Added Städteservice Raunheim Rüsselsheim

ubl8:
Added the staedteservice_de.py and staedteservice_de.md
Updated the README.md and info.md

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
@@ -46,6 +46,7 @@ class Source:
         calendar = self.clean_ics(current_calendar)
         dates = self._ics.convert(calendar)
 
+        # in december the calendar for the next year is available
         if self.month == 12:
             self.year += 1
             next_calendar = self.get_calendar_from_site(session)
@@ -65,9 +66,9 @@ class Source:
         return r.text
     
     def get_city_code(self) -> str:
-        # returns the citycode based on the city input
-
+        # returns the city_code based on the city input
         city_code = ""
+
         if self.city == "Rüsselsheim":
             city_code = "1"
         elif self.city == "Raunheim":
@@ -80,26 +81,28 @@ class Source:
 
         currentDateTime = datetime.datetime.now()
         date = currentDateTime.date()
-        
-        return int(date.strftime("%Y"))
+        year = int(date.strftime("%Y"))
+
+        return year
 
     def get_month(self) -> int:
-        # returns the current year
+        # returns the current month
 
         currentDateTime = datetime.datetime.now()
         date = currentDateTime.date()
-        
-        return int(date.strftime("%m"))
+        month = int(date.strftime("%m"))
+
+        return month
 
     def clean_ics(self, calendar: str) -> str:
         # clean ics from problematic lines
-
-        # split calendar and remove problematic lines
-        split_calendar = calendar.split("\n")
-        clean_calendar = ""
-
+        
         # lines to be removed from ics
         remove_tuple = ("BEGIN:VALARM", "TRIGGER", "ACTION:DISPLAY", "DESCRIPTION", "END:VALARM")
+
+        # split lines
+        split_calendar = calendar.split("\n")
+        clean_calendar = ""
 
         # do cleanup
         for index in range(0,len(split_calendar)):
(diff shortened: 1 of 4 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+            city_code = "2"
+
+        return city_code
+
+    def get_year(self) -> int:
+        # returns the current year
+
+        currentDateTime = datetime.datetime.now()
+        date = currentDateTime.date()
+        year = int(date.strftime("%Y"))
```
mampfes, line 84: Why not simply: `currentDateTime.date().year` ?

custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+        r.raise_for_status()
+        r.encoding = "utf-8"  # enshure it is the right encoding
+
+        return r.text
+    
+    def get_city_code(self) -> str:
+        # returns the city_code based on the city input
+        city_code = ""
+
+        if self.city == "Rüsselsheim":
```
mampfes, line 72: Better use a dict for this:
```
# add this as global variable at the beginning
CITY_CODE_MAP = {  
"Rüsselsheim": 1,
"Raunheim": 2
}

# do this in __init__
self.city_code = CITY_CODE_MAP[city]
```
Then this function becomes obsolete.

custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+        # returns the current month
+
+        currentDateTime = datetime.datetime.now()
+        date = currentDateTime.date()
+        month = int(date.strftime("%m"))
+
+        return month
+
+    def clean_ics(self, calendar: str) -> str:
+        # clean ics from problematic lines
```
mampfes, line 98: Just to be curious: Why are these lines problematic? Would it also work without cleaning???
If this "cleaning" is really required, please add a comment which describes the problem.

custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+            self.year += 1
+            next_calendar = self.get_calendar_from_site(session)
+            calendar = self.clean_ics(next_calendar)
+            dates += self._ics.convert(calendar)
+
+        return dates
+
+    def get_calendar_from_site(self, session: requests.Session) -> str:
+        # example format: https://www.staedteservice.de/abfallkalender_1_477_2023.ics
+        URL = BASE_URL+"_"+self.city_code+"_"+self.street_number+"_"+str(self.year)+".ics"
```
mampfes, line 60: better use a f-string instead of string concatenation

custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+}
+
+BASE_URL = "https://www.staedteservice.de/abfallkalender"
+
+class Source:
+    def __init__(self, city, street_number):
+        self.city = str(city)
+        self.city_code = self.get_city_code()
+        self.street_number = str(street_number)
+        self.year = self.get_year()
```
mampfes, line 29: Why are you assigning year and month here? `__init__ `will be executed only once if Home Assistant starts.
I would do this all in fetch()

custom_components/waste_collection_schedule/waste_collection_schedule/source/staedteservice_de.py
```diff
@@ -0,0 +1,116 @@
...
+        return entries
+
+    def get_dates(self, session: requests.Session) -> list:
+        current_calendar = self.get_calendar_from_site(session)
+        calendar = self.clean_ics(current_calendar)
+        dates = self._ics.convert(calendar)
+
+        # in december the calendar for the next year is available
+        if self.month == 12:
+            self.year += 1
```
mampfes, line 51: I guess this will stop working in day 2 after starting Home Assistant...

mampfes (changes requested):

ubl8:
I updated the code according to the suggestions.

mampfes:
Thanks for your contribution.

PR #467, opened 2022-12-13
canadabay_nsw_gov_au: retrieve 1 year of data

calvinbui:
To match the website also retrieving 1 year of data.

image

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/canadabay_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/canadabay_nsw_gov_au.py
@@ -22,6 +22,7 @@ TEST_CASES = {
         "suburb": "Mortlake",
         "street_name": "Tennyson Road",
         "street_number": "76",
+        "days": 60,
     },
 }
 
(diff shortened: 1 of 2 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/canadabay_nsw_gov_au.py
```diff
@@ -22,25 +22,27 @@
...
+        "days": 60,
     },
 }
 
 HEADERS = {"user-agent": "Mozilla/5.0"}
 API_URL = "https://canada-bay.waste-info.com.au/api/v1"
 
 class Source:
-    def __init__(self, suburb, street_name, street_number):
+    def __init__(self, suburb, street_name, street_number, days=183):
```
mampfes, line 33: There is no need to limit the information while retrieving it. Just retrieve everything - in your case I would also use 1 year. The amount of data can be limited on the sensor side if necessary.

Therefore please remove this configuration option again.

calvinbui: PR updated

mampfes:
Thanks for your contribution.

PR #469, opened 2022-12-14
canterbury_gov_uk: Fix Broken API requests

GigaByte4711:
Canterbury Gov have changed the responses given by their API. This fixes that change


mampfes:
Thanks for your contribution.

PR #471, opened 2022-12-16
Added support for waste collection for Tewkesbury Borough Council (UK)

Rohaq:
Added source for Tewkesbury Borough Council using the public API endpoint used by their website.


mampfes:
Thanks for your contribution.

PR #472, opened 2022-12-17
Added support for Middlesbrough Council - UK

ximon:
Added source and documentation for Middlesbrough council - UK
Currently uses the UPRN.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
@@ -1,0 +1,109 @@
+import logging
+import requests
+import time
+
+from datetime import datetime
+from waste_collection_schedule import Collection
+
+TITLE = 'middlesbrough.gov.uk'
+DESCRIPTION = (
+    'Source for waste collection services for Middlesbrough Council'
+)
+URL = 'https://www.middlesbrough.gov.uk/bin-collection-dates'
+
+
+HEADERS = {
+    "user-agent": "Mozilla/5.0",
+}
+
+COOKIES = {
+
+}
+
+TEST_CASES = {
+    "Tollesby Road - number" : {"uprn": 100110140843},
+    "Tollesby Road - string" : {"uprn": "100110140843"},
+    "Victoria Road - number" : {"uprn": 100110774949},
+    "Victoria Road - string" : {"uprn": "100110774949"},
+}
+
+API_URLS = {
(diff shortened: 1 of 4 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
```diff
@@ -0,0 +1,109 @@
...
+
+        #now query using the uprn
+        payload = {
+            "formValues": { "Find My Collection Dates": { "uprn_search": { "value": self._uprn } } }
+        }
+        scheduleRequest = s.post(API_URLS['schedule'] + '&_' + str(now) + '&sid=' + sessionKey , headers=HEADERS, json=payload)
+        #print(scheduleRequest.json())
+        data = scheduleRequest.json()['integration']['transformed']['rows_data']['0']
+
+        print(data)
```
mampfes, line 77: Remove debug output

custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
```diff
@@ -0,0 +1,109 @@
...
+}
+
+
+_LOGGER = logging.getLogger(__name__)
+
+
+class Source:
+    def __init__(self, uprn: str = None):
+        self._uprn = str(uprn)
+        if not self._uprn:
```
mampfes, line 50: If you remove `= None` from the uprn argument then you can also discard this if statement.

custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
```diff
@@ -0,0 +1,109 @@
...
+        greenDates = data['Green'].split('<br />')
+
+        entries = []
+
+        for date in refuseDates:
+            if len(date) > 0:
+                entries.append(Collection(
+                    date = datetime.strptime(date, '%d/%m/%Y').date(),
+                    t = 'refuse bin',
+                    icon = ICONS.get('REFUSE')
```
mampfes, line 90: Just wondering why you use a `ICONS` dict here. Wouldn't it be simpler to just name the icon here directly (same as for `t`)?

mampfes (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/middlesbrough_gov_uk.py
ximon: Very true, i was just going with the style of the source i based this upon.

ximon:
I've made the suggested changes and tested 👍

mampfes:
Thanks for your contribution.

PR #489, opened 2022-12-19
Add support for ART Trier

### LINKED ISSUE #383: Add support for Trier/Irrel Rheinland Pfalz
Hey friends, hope you are doing well.

I want just too ask if it could be possible to add:

https://www.art-trier.de/eo/cms?_sprache=de&_bereich=artikel&_aktion=detail&idartikel=408001

Irrel and maybe more from the ART. 

This would be awesome. 

Thanks very much. 

Can I also buy a beer? Can we donate to this little project?

mirkolenz:
Fixes #383

I added multiple test cases that cover different issues for creating the URL. However, it may be the case that some edge cases are not correctly handled as of now. In the current implementation, an error is thrown if the ICS parser is unable to parse the response.

If desired, I could wrap line 74 with a `try` statement and fail silently. I am not sure how the other sources handle such scenarios (I only dealt with Cochem-Zell so far), but I would tend to keep it this way and throw an error.


mampfes:
Thanks for your contribution.

PR #502, opened 2022-12-22
Added manual way to retrieve "street_code" for AWB Köln

scharmach:
This might be useful for people who don't have Python installed or don't want to install it.


mampfes:
Thnaks for your contribution.

PR #504, opened 2022-12-22
Added maribyrnong.py/.md to support source for Maribyrnong Council (V…

dpeluso:
…IC) rubbish collection, https://www.maribyrnong.vic.gov.au/Residents/Bins-and-recycling

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/maribyrnong.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/maribyrnong.py
@@ -1,0 +1,117 @@
+import json
+from datetime import date, timedelta
+
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Maribyrnong Council (VIC)"
+DESCRIPTION = "Source for Maribyrnong Council (VIC) rubbish collection."
+URL = "https://www.maribyrnong.vic.gov.au/Residents/Bins-and-recycling"
+TEST_CASES = {
+    "Random address": {
+        "suburb": "Footscray",
+        "street_name": "Ballarat Rd",
+        "street_number": "70-100",
+    }
+}
+
+HEADERS = {"user-agent": "Mozilla/5.0"}
+
+
+class Source:
+    def __init__(self, suburb, street_name, street_number):
+        self.suburb = suburb
+        self.street_name = street_name
+        self.street_number = street_number
+
+    def fetch(self):
+
+        suburb_id = 0
+        street_id = 0
+        property_id = 0
+        today = date.today()
+        nextmonth = today + timedelta(30)
+
+        # Retrieve suburbs
+        r = requests.get(
+            "https://maribyrnong.waste-info.com.au/api/v1/localities.json", headers=HEADERS
+        )
+        data = json.loads(r.text)
+
+        # Find the ID for our suburb
+        for item in data["localities"]:
+            if item["name"] == self.suburb:
+                suburb_id = item["id"]
+                break
+
+        if suburb_id == 0:
+            return []
+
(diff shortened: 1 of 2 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/maribyrnong.py
```diff
@@ -0,0 +1,117 @@
...
+            if item["name"] == f"{self.street_number} {self.street_name} {self.suburb}":
+                property_id = item["id"]
+                break
+
+        if property_id == 0:
+            return []
+
+        # Retrieve the upcoming collections for our property
+        r = requests.get(
+            f"https://maribyrnong.waste-info.com.au/api/v1/properties/{property_id}.json?start={today}&end={nextmonth}",
```
mampfes, line 84: Is there more data available than one month? If yes, please fetch all data (up to one year).

custom_components/waste_collection_schedule/waste_collection_schedule/source/maribyrnong.py
```diff
@@ -0,0 +1,117 @@
...
+            if item["name"] == self.suburb:
+                suburb_id = item["id"]
+                break
+
+        if suburb_id == 0:
+            return []
+
+        # Retrieve the streets in our suburb
+        r = requests.get(
+            f"https://maribyrnong.waste-info.com.au/api/v1/streets.json?locality={suburb_id}",
```
mampfes, line 52: Please use `params` argument to pass parameters in URLs: https://requests.readthedocs.io/en/latest/user/quickstart/#passing-parameters-in-urls

mampfes (changes requested):
- Please rename source (and docu) to maribyrnong_vic_gov_au
- Please add links to new source to README.md and info.md

DanAE111:
Looks like Maribyrnong also uses OpenCities (you can see it mentioned on the website footer), you might find you can reference the Melton source, it could be as simple as changing the URLs to scrape the data from the Maribyrnong city council.

mampfes:
> Looks like Maribyrnong also uses OpenCities (you can see it mentioned on the website footer), you might find you can reference the Melton source, it could be as simple as changing the URLs to scrape the data from the Maribyrnong city council.

@DanAE111: Interesting! Do you have an idea how many councils using that service? Maybe it is worth to create a single source which could handle all OpenCities.

DanAE111:
@mampfes As far as i know there are at least 3 but Granicus are touting they have over 170 government agencies using OpenCities (https://granicus.com/solution/govaccess/opencities/). If the implementation across the different agencies is the same we could certainly scrape the data with a single source.

The Maribyrnong site is different visually to Melton and Nillumbik which are the other two i know are using OpenCities in Australia. So maybe that means the source would be different in this case?

I had a quick Google and it would seem that many are similar to Melton. But it looks like the sites can change the implementation of how the waste collection lookup works, i would assume those ones are not compatible. I've listed some examples of those similar below and a few outliers i have found, it would be interesting to see if different countries using OpenCities share the address lookup which seems to be the trend here.

OpenCities Address Lookup:
https://www.murrayriver.nsw.gov.au/Our-Services/Waste-and-recycling/Find-my-bin-day
https://www.tweed.nsw.gov.au/property-rates/waste-recycling/bin-services-collection/bin-collection-days
https://www.melton.vic.gov.au/Services/Waste-and-recycling/Find-my-bin-day
https://www.nillumbik.vic.gov.au/Residents/Waste-and-recycling/Bin-collection/Check-my-bin-day

Alternative Waste Collection Lookup:
https://www.wdrc.qld.gov.au/Business-Development/Maps/Interactive-Maps
https://www.maribyrnong.vic.gov.au/Residents/Bins-and-recycling

mampfes:
I did the requested changes in #621

PR #505, opened 2022-12-23
Add support for Bürgerportal

mirkolenz:
I have discovered today that my local waste collector abandoned its existing infrastructure for the next year. This is currently available here via the `cochem_zell_online_de` source.

Instead, they will migrate to a new "Bürgerportal" developed by c-trace. There is already an existing c-trace source in this integration `c_trace_de`, but this only supports their older system. This new one does not have a canonical URL (i.e., it can be hosted on any server). Thus, I am not sure how to call this new source. As a result, this is only a draft PR and no documentation has been added so far. 

The code itself is finished and tested. While my local service is currently in maintenance (thus the test is commented), I was able to find two other districts that use the same software. This Bürgerportal is developed in JavaScript/Nuxt, so I had to reverse-engineer the API. In its current state, the user only has to provide the address the same way it is entered in the web-based form and the code will automatically retrieve the correct ID's.

This leads to a relevant issue: I assume that I am basically accessing the internal Nuxt API here that may be subject to change at any time. I am also not sure whether the API will stay the same between the different instances of this portal. So we should maybe discuss whether it makes sense to keep these servers combined or if an individual source for each operator should be created.

I tried to keep my code modular: I have included typed dicts of the API results and designed the icon assignment in a dynamic fashion (by searching for keywords like "bio" or "papier" in the waste types).

I am happy for any feedback on this pull request and will adjust the code as needed 😃


custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_buergerportal.py
```diff
@@ -0,0 +1,233 @@
...
+        district_id = self.fetch_district_id(session)
+        street_id = self.fetch_street_id(session, district_id)
+        # Eventually verify house number in the future
+
+        params = {
+            "$expand": "Abfuhrplan,Abfuhrplan/GefaesstarifArt/Abfallart,Abfuhrplan/GefaesstarifArt/Volumen",
+            "$orderby": "Abfuhrplan/GefaesstarifArt/Abfallart/Name,Abfuhrplan/GefaesstarifArt/Volumen/VolumenWert",
+            "orteId": district_id,
+            "strassenId": street_id,
+            "jahr": year,
```
mampfes, line 91: Do your really need year as argument? Some service providers return all available data is you don't specify a date. If not (= year is required), could you please fetch the data for the current year and in December also the data for the next year?

mampfes (commented):
Nice job and description - thanks for that!

Regarding naming: c-trace obviously doesn't appear in an URL and is not used in the source code. Therefore I would remove it from the source name. Proposal: `buergerportal_de`?

> may be subject to change at any time

Yes, that is the risk that we always have as long as the service providers are not forced to offer a open data API. Fortunately the APIs do not change very often.

> servers combined or if an individual source 

I would prefer to create one source for now. We do not know so far whether there will be differences between instances. Once we see that there are differences we can better adapt because we known the differences.

I would also prefer to hide the url behind a map and just offer a "customer" or "municipality" or whatever argument. The possible values should be added to the documentation.

custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_buergerportal.py
mirkolenz: Yes, we need the year as an argument. The API endpoint `AbfuhrtermineAbJahr` returns all upcoming waste collections starting with a certain `year`. In my tests, I sometimes had the scenario that only collections in 2023 were returned if leaving this parameter out. If given the year 2022, then we will also receive all collections from 2023: The parameter `year` is only meant to specify the *starting point*.

mirkolenz:
Thank you for your comments! I will implement your suggestions and update this pull request. Also, some minor adjustments are needed for the "Cochem-Zell" district: They just published the new website today and handle `subdistrict` more strictly than the other sources.

mirkolenz:
I have implemented the suggested changes and updated the docs accordingly. One question is left: How should I proceed with the `cochem_zell_online_de` Python source? It will not receive new entries for 2023. Should I just delete the file or is there a recommended approach to "deprecating" such sources? I have already updated the links in `info.md` and `README.md` to the new Bürgerportal.

mampfes:
> How should I proceed with the cochem_zell_online_de Python source?

Yes, please remove it - within this PR. I will add a comment to the release notes.

mirkolenz:
Done. I also resolved a merge conflict resulting of #517

mampfes:
Thanks for your contribution.

PR #506, opened 2022-12-23
Improve code quality of art_trier_de

mirkolenz:
During my work on #505, I have come up with some improvements for the new `art_trier_de` source: The logic for parsing the correct district has been simplified and the source will now raise an exception if there is an HTTP error so that the user is aware of potential issues.


mampfes:
Thanks for your contribution.

PR #508, opened 2022-12-23
awb_es_de: now using all available ics

meilon:
AWB-ES uses one ics file per year, now every linked ics for a given location gets pulled


mampfes:
Thanks for your contribution.

PR #518, opened 2022-12-24
Add Source SRV atervinning, SE

bbr111:
Add Source SRV atervinning, SE


custom_components/waste_collection_schedule/waste_collection_schedule/source/srv_se.py
```diff
@@ -0,0 +1,39 @@
...
+            return []
+
+        data = r.json()
+
+        entries = []
+
+        for container in data["results"][0]["containers"]:
+            type=container["contentType"]
+            for calentry in container["calendars"]:
+                dates =  {calentry["startDate"],type}
```
mampfes, line 35: `dates` is not used

mampfes (changes requested):
Please rename .py and .md files to srvatervinning_se

custom_components/waste_collection_schedule/waste_collection_schedule/source/srv_se.py
bbr111: changed in last commit

mampfes:
Thanks for your contribution.

PR #520, opened 2022-12-26
Add fetch_data Service

### LINKED ISSUE #382: Please expose service for manual schedule refresh
As per my understanding current setup  allows refresh of the schedule to happen only once a day at the time configured in `fetch_time`.
This may cause issues if for some reason the source is not available at the given time, there is an issue with connectivity or a schedule change has been announced via different channels and update needs to happen on-demand.

Please expose `waste_collection_schedule.reload` service that would call the same routing that is normally executed at `fetch_time`, but on demand.

bbr111:
Add an Service .refresh_source to manual update source

This a first suggestion. to be discussed.

This can resolve #382


mampfes (commented):
Nice!

The only thing that is really missing is documentation :-) Regarding documentation: dtgit215 is currently restructuring the whole documentation. Therefore any additions to the docu may subject of change until he has finished.

And I would rename the service to `fetch_data` because we use `fetch` everywhere in the source code and it actually fetches data from multiple sources.

The next question is then if we need an option to stop the automatic fetching? In my opinion we don't need this and I would first start without it!

It may be also interesting to add an additional service which allows to trigger the `day_switch_time`, which basically elapses entries for the current day.

bbr111:
@mampfes 

it agree in renaming.
Will try with a service vor day_switch_time.

I don't think, that we need to stop the automatic fetching. In my opinion this is only an additional option, if you want a faster refresh without restarting HA.

PR #534, opened 2022-12-30
added service and more help

dabenzel:
added a (imo) better help, how to get the service-name and other parameters for c-trace


mampfes:
Thanks for your contribution.

PR #535, opened 2022-12-30
Fix duplicate entries for buergerportal_de & some other enhancements

mirkolenz:
When adding `buergerportal_de`, I somehow missed that the source returns duplicate entries: It differentiates between different container sizes. This PR uses a set to remove such duplicates since in most cases, these different sizes will still be collected on the same day. If this is not the case for certain areas, it also adds a new option `show_volume` that includes the size of the container in the title of the entry. This way, users can use the built-in `customize` option of the sources configuration to adjust the collection types to their needs. These two changes give maximum freedom while still making the source easy to set up for most use cases. The corresponding documentation has also been updated to explain this new parameter and point users to the `customize` option in case they need it.

As part of this PR, I also removed the default icon of art_trier_de (since this integration provides one anyway) and added the correct `Optional[str]` type to some parameters of the `Collection` class.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/collection.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/collection.py
@@ -1,8 +1,14 @@
 import datetime
+from typing import Optional
 
 
 class CollectionBase(dict):  # inherit from dict to enable JSON serialization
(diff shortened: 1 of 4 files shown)


mampfes:
Thanks for your contribution.

PR #542, opened 2023-01-01
Add source for Ashfield District Council, UK

ConfusedTA:
Thanks for the great framework - adding this didn't take much time at all!

The only thing that's not tested so far is the brown/garden waste collection - this doesn't start until March/April-ish around here - I did kind of investigate this with openHAB last year, but didn't get very far, so _from memory_ the garden/brown follows a different naming scheme... though I will monitor this and make sure that it works as expected - but thought that the current stuff is working for me, it'd be good to share it.

If there are any modifications required, please just ask!


mampfes:
Thanks for your contribution.

PR #544, opened 2023-01-02
changed URLs, after KWU changed calendardownload

maverickme:


mampfes:
Thanks for your contribution.

PR #546, opened 2023-01-03
Fix documentation for berlin_recycling_de

schmaun:
as the example was wrong


mampfes:
Thanks for your contribution.

PR #549, opened 2023-01-03
Added a new source for Abfallwirtschaft Nürnberger Land

jodli:
Hey!

I've added a new source for the Abfallwirtschaft Nürnberger Land.
The URL is this one: https://abfuhrkalender.nuernberger-land.de/#waste_calendar/calendar

This is just the first iteration where the user has to get the id from the webpage. :(
I want to change this in a future iteration to also support setting the city, district and street name via the config.

Also thanks for the hard plugin work!


mampfes:
Thanks for your contribution.

mampfes:
I renamed the source because only the top-level domain name is necessary (naming convention) and fixed an issue with non-ascii characters in the UID field. Otherwise it crashed on my machine.
And it also splits the waste types into separate entries now.

mahe112:
Hi jodli, 
great work adding Abfallwirtschaft Nürnberger Land to the waste collection schedule!
But some entries are missing in the ics, meaning dates where Biotonne is collected without Restmüll in summer time.
By adding the filter "&filter=rm:bio:p:dsd" to the url after the id such entries are included in the resulting ics.
Without this filter such entries are missing in the resulting ics.
Can you please verify?

jodli:
> Hi jodli, great work adding Abfallwirtschaft Nürnberger Land to the waste collection schedule! But some entries are missing in the ics, meaning dates where Biotonne is collected without Restmüll in summer time. By adding the filter "&filter=rm:bio:p:dsd" to the url after the id such entries are included in the resulting ics. Without this filter such entries are missing in the resulting ics. Can you please verify?

Hey!
I was of the impression when I omit the `filter` argument I would get every entry in the table, but maybe I was wrong with that assumption.
So after comparing the file with the filter and without it, I can confirm that you are right.
I will prepare a PR that fixes this!

5ila5:
> Hi jodli, great work adding Abfallwirtschaft Nürnberger Land to the waste collection schedule! But some entries are missing in the ics, meaning dates where Biotonne is collected without Restmüll in summer time. By adding the filter "&filter=rm:bio:p:dsd" to the url after the id such entries are included in the resulting ics. Without this filter such entries are missing in the resulting ics. Can you please verify?

fixed by #982

Issue #559, opened 2023-01-04
Bug: goldcoast_qld_gov_au definition for street_address is incorrect - Line 27


TheBlackMini:
2023-01-04 11:37:42.519 ERROR (MainThread) [homeassistant.setup] Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 253, in _async_setup_component
    result = await task
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 109, in async_setup
    api.add_source_shell(
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 200, in add_source_shell
    SourceShell.create(
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 196, in create
    source = source_module.Source(**source_args)  # type: ignore
TypeError: Source.__init__() got an unexpected keyword argument 'street_address'

PR #560, opened 2023-01-04
Fix: #559 - goldcoast_qld_gov_au definition for street_address was incorrect

TheBlackMini:
Fix for #559 - Line 27 had the wrong parameters, not sure how this made it through the tests!


mampfes:
Thanks for your contribution.

PR #579, opened 2023-01-07
Added a new source for Stadt Fürth

raaaf:
Hey @mampfes!

New source coming up for Stadt Fürth.

I can't do the magical python work, so I just prepared everything I could.
Would you try it, please? Eager to see how you solve that..

URL: https://abfallwirtschaft.fuerth.eu

**PS: Danke dir für das geniale Plugin!**

Cheers,
Rafael


jodli:
I will take a look later 😊

dt215git:
Once you can see your collection shcedule, if you click on `Abfehrtermine importieren` and then hover over the `Kalender-Datei herunterladen` button you can see the url for the ics calendar download. It'll be something like: `https://abfallwirtschaft.fuerth.eu/termine.php?icalexport=<some_number>`.

You may be able to use the [ics](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/ics.md) option to process this rather than scripting around the street/house search. The search seems to derive the number that's  referenced in the calendar url.

image

sir106:
was also working on adding Stadt Fürth as a service provider. have a look at pull request #602 if it fits your needs...

mampfes:
already implemented in #602

PR #581, opened 2023-01-08
Update AVL Ludwigsburg ics link example

WolfMu:
Update AVL Ludwigsburg ics link example, but please create your own file on avl homepage.

The issue recommended to use the ics implemenation, but the link was outdated.


mampfes:
Thanks for your contribution.

PR #591, opened 2023-01-09
Add Wyre Forest District Council, UK

ConfusedTA:
A friend lives in this area, and has started using HA, so thought I'd do this on his behalf!

The only thing that's not tested so far is the brown/garden waste collection - this doesn't start until March in his area - I will monitor and amend if necessary once it starts showing dates.


mampfes:
Thanks for your contribution.

PR #593, opened 2023-01-09
Add Harlow, UK

ConfusedTA:
A friend lives in this area, and has started using HA, so thought I'd do this on his behalf!

Used a small selection of random addresses from the area for testing.


ConfusedTA:
@mampfes I've just looked at another friend's address, and they appear to use the same type of data provider, but probably just with a different request URL.

I wasn't sure from the documentation if this is a supported scenario, or whether each one should be a separate source file?

I'm happy to hold fire on this one for the moment if it'll be better to join the two together?

Just let me know how you'd prefer to proceed!

mampfes:
Thanks for your contribution.

mampfes:
This source is so simple that I will accept it as a separate source file. For more complex ones I prefer to have it only once.

PR #604, opened 2023-01-11
Added Regensburg as source to awido documentation

Jaegerstefan:


mampfes:
Thanks for your contribution.

PR #609, opened 2023-01-13
Adding Waverley Borough Council

gingemonster:
Adding Waverley Borough Council, should this be accepted can clone to do East Herts Council as same service but slight differences in text


mampfes:
Thanks for your contribution.

mampfes:
I would really appreciate if you could add other councils, too. I had to do a minor change in your PR, otherwise the auto-generated README.md would be broken with the next run of `update_docu_links.py`. (The other changes are just automatic code beautifications by black/mypy/...). Therefore please use the latest version from the master branch for the clone.

gingemonster:
Thanks, I for some reason dont seem to be able to run update_docu_links.py without errors on windows so have to do the change manually

Issue #586, opened 2023-01-08
provide custom HEADERS in ical source


immenz:
Hello, 

I'm trying to download an ics schedule from my providers homepage. It only works when a "Referer":"<url>" Header is set. Otherwise a 403 Error occure. I manually changed the fixed Header (currently the user-agent) in the ics.py, but that will be overridden in future updates, i guess.

My question (or feature request, i suppose) is, is there a way to provide such a custom Header in the source configuration for ics?

bbr111:
@mampfes 
Could we just add a new argument headers and extend the existing HEADERS variable?

```python
class Source:
    def __init__(
        self,
        url=None,
        file=None,
        offset=None,
        params=None,
        year_field=None,
        method="GET",
        regex=None,
        split_at=None,
        version=2,
        verify_ssl=True,
---     headers=None,   #add this line
    ):
        self._url = url
        self._file = file
        if bool(self._url is not None) == bool(self._file is not None):
            raise RuntimeError("Specify either url or file")
        if version == 1:
            self._ics = ICS_v1(offset=offset, split_at=split_at, regex=regex)
        else:
            self._ics = ICS(offset=offset, split_at=split_at, regex=regex)
        self._params = params
        self._year_field = year_field  # replace this field in params with current year
        self._method = method  # The method to send the params
        self._verify_ssl = verify_ssl
---     if headers is not None:  #add this line
---         HEADERS.update(headers)  #add this line
```

mampfes:
@immenz I created a PR with the proposed extensions. Could you please test if it is working for your service provider. If yes, please provide an example in the ICS source docu, so that everyone can reuse it.

immenz:
I can confirm its working nicely. Thanks for the quick response.

PR #610, opened 2023-01-14
Ics custom header use case documentation

immenz:
documentation for custom header use case "Abfallwirtschaftsbetrieb Ilm-Kreis"

See https://github.com/mampfes/hacs_waste_collection_schedule/issues/586


mampfes:
Thanks for your contribution.

PR #613, opened 2023-01-15
Fix example configuration for South Cambs

quentinsf:
Fix example configuration for South Cambs, which includes an unneeded 'version' parameter (which gives an error).


mampfes:
Thanks for your contribution.

PR #614, opened 2023-01-15
Renaming of "Colchester Borough Council"

thewillwilson:
Updating documentation following rename from Colchester Borough Council to Colchester City Council


mampfes (changes requested):
Please update also (at least) the `TITLE` variable in the corresponding `.py` file because the entries in README.md and info.md are generated from that.

mampfes:
Thanks for your contribution.

mampfes:
I will do the requested change.

PR #617, opened 2023-01-16
Add Newport, UK

ConfusedTA:
Another source for another friend, this time for Newport, which is in south Wales, UK.


mampfes:
Thanks for your contribution.

mampfes:
I removed the name mapping, because this can be done via aliases individually.

PR #620, opened 2023-01-18
Added source for Region Gotland

oskarannas:
This adds a source for Region Gotland, serving the municipality of Gotland, Sweden.

Searching by address is not available since Region Gotland doesn't provide this service.

So the user has to login in to the service and find the property reference number and the enter this into the YAML config.

This procedure is explained in the docs.

Because of this there is no publicly available reference number to use for the test cases, so i've just added a dummy value.

However i have tested the script with my own unique number and it works as expected.


custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
```diff
@@ -0,0 +1,40 @@
...
+from urllib.parse import urlencode
+
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Region Gotland"
+DESCRIPTION = "Source for Region Gotland waste collection."
+URL = "https://edpfuture.gotland.se"
+TEST_CASES = {
+    "Dummy": {"uprn": "0000000000"},
```
mampfes, line 13: This doesn't look like a valid uprn. Please provide at least one valid test case which can be test with `test_sources.py`

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
```diff
@@ -0,0 +1,40 @@
...
+}
+
+class Source:
+    def __init__(self, uprn):
+        self._uprn = uprn
+
+    def fetch(self):
+        query_params = urlencode({"address": '(' + self._uprn + ')'})
+        response = requests.get(
```
mampfes, line 22: Please pass arguments in URLs via the `params` argument. This will probably obsolete the `urlencode` above.

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
```diff
@@ -0,0 +1,40 @@
...
+from datetime import datetime
+import json
+from urllib.parse import urlencode
+
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Region Gotland"
+DESCRIPTION = "Source for Region Gotland waste collection."
+URL = "https://edpfuture.gotland.se"
```
mampfes, line 11: Please remove `edpfuture` from URL.

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
```diff
@@ -0,0 +1,40 @@
...
+            "https://edpfuture.gotland.se/FutureWeb/SimpleWastePickup/GetWastePickupSchedule?{}"
+            .format(query_params)
+        )
+        data = json.loads(response.text)
+
+        entries = []
+        for item in data["RhServices"]:
+            waste_type = item["WasteType"]
+            icon = "mdi:trash-can"
+            if waste_type == "Matavfall":
```
mampfes, line 32: Please use an ICON_MAP (see other sources) to map icon types.

mampfes (changes requested):
Please rename to gotland_se.py (same as the primary URL), Same for .md file.

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
oskarannas: That's correct. Since the uprns are personal there's no way for me to provide one for the test case. I've tested it with my own personal one to make sure the code works.

oskarannas: This is the result from running the test script with my personal uprn:

```
λ python custom_components\waste_collection_schedule\waste_collection_schedule\test\test_sources.py -s region_gotland_se -i -l
Testing source region_gotland_se ...
  found 2 entries for Dummy
    2023-01-25: Restavfall [mdi:trash-can]
    2023-01-25: Matavfall [mdi:leaf]
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
oskarannas: Yes, fixing this

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
oskarannas: Yes, fixing this

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
mampfes: You don't have to provide your personal uprn. Is there a public webpage which can be used to get the uprn of e.g. a public building or a restaurant? 
Without a valid test case the code cannot be maintained because we don't know whether it is broken. Maybe you can guess another uprn???

mampfes:
We defined that, by default, the source file name is derived from the url that is used to access the service - because this is a unique identifier and also well known to the users. If I see it correctly than the county of Gotland is using the url `gotland.se`, therefore the default name for the source would be gotland_se.

If there is another provider in the county of Gotland, it would probably use a different url, right?

oskarannas:
> We defined that, by default, the source file name is derived from the url that is used to access the service - because this is a unique identifier and also well known to the users. If I see it correctly than the county of Gotland is using the url `gotland.se`, therefore the default name for the source would be gotland_se.
> 
> If there is another provider in the county of Gotland, it would probably use a different url, right?

Yes, that makes sense! :)

custom_components/waste_collection_schedule/waste_collection_schedule/source/region_gotland_se.py
oskarannas: I managed to get a hold of a working test uprn so that solves the issue! :)

mampfes:
Thanks for your contribution.

Issue #650, opened 2023-01-29
Request: South Derbyshire District Council, United Kingdom


Novem30:
Great project, thank you all very much.

It would be great to add South Derbyshire District Council in the United Kingdom bin collection schedule.

Bin collection dates can be found here:
https://www.southderbyshire.gov.uk/our-services/recycling-bins-and-waste/bin-collection-dates

They also have a site here that gives bins collection dates, planning applications (that would be a cool addon, hehe), councillors, and references the UPRN of the house:
https://maps.southderbyshire.gov.uk/iShareLive.Web/atmyresponsivecouncil.aspx

Example postcode: DE65 6FT

Also, according to this website the CMS South Derbyshire are using is 'web-labs' used on other council websites so maybe it'll be useful for them too. I'm not knowledgeable enough to know though: https://www.viko.co.uk/blog/local-government-cms/

Thanks in advance!

mampfes:
fixed by #651

PR #651, opened 2023-01-30
adding South Derbyshire District Council

gingemonster:
implementing https://github.com/mampfes/hacs_waste_collection_schedule/issues/650 but this organisation has something I have not seen before and recycle coffee pods so I have implemented as a bin type with the "mdi:coffee" icon, let me know if that is the right thing to do as I couldn't find any other provider with this to follow.


mampfes:
Thanks for your contribution.

Issue #649, opened 2023-01-29
Calender is Blank and Error when trying to call service


jb123567:
Hi trying to configure for Wiltshire Council in the UK. I am getting this error when I manually call the service and the calender is blank. Am I missing something?

`websocket_api script: Error executing script. Unexpected error for call_service at pos 1: async_setup.<locals>.fetch_data() takes 0 positional arguments but 1 was given`

`Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 451, in _async_step
    await getattr(self, handler)()
  File "/usr/src/homeassistant/homeassistant/helpers/script.py", line 684, in _async_call_service_step
    await service_task
  File "/usr/src/homeassistant/homeassistant/core.py", line 1755, in async_call
    task.result()
  File "/usr/src/homeassistant/homeassistant/core.py", line 1798, in _execute_service
    await self._hass.async_add_executor_job(
  File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
TypeError: async_setup.<locals>.fetch_data() takes 0 positional arguments but 1 was given`

mampfes:
These are definitely 2 different problems.

1. How do you call `fetch_data`? It doesn't take an argument! But normally you don't have to call this service, esp. if you have problems with missing data.
2. Missing data: Please share your configuration. Did you set `postcode` and `uprn`?

jb123567:
I get the area by simply selecting the fetch data service and calling it in developer tools? I only tried this as the calendar was blank.

My config is as below taken from the readme instructions. I removed my exact details but the post code didn't have spaces and UPRN is correct.

`waste_collection_schedule:
  sources:
    - name: wiltshire_gov_uk
      args:
        postcode: SP****
        uprn: 1009******* `

Thanks,

J

> These are definitely 2 different problems.
> 
> 
> 
> 1. How do you call `fetch_data`? It doesn't take an argument! But normally you don't have to call this service, esp. if you have problems with missing data.
> 
> 2. Missing data: Please share your configuration. Did you set `postcode` and `uprn`?

dt215git:
Using a random SP2 postcode and corresponding uprn it seems to work for me.

```yaml
waste_collection_schedule:
  sources:
    - name: wiltshire_gov_uk
      args:
        postcode: "SP2 ..."
        uprn: "100121......"
      calendar_title: Wilts

[...]

PR #656, opened 2023-02-01
updated wiltshire_gov_uk

dt215git:
uprn now automatically padded to 12 characters to address issue identified in #649
Icon map also added


mampfes:
Thanks for your contribution.

PR #667, opened 2023-02-07
Added Innherred Renovasjon, Norway

bendikrb:


custom_components/waste_collection_schedule/waste_collection_schedule/source/irnt_no.py
```diff
@@ -0,0 +1,118 @@
...
+        """Fetch fetch"""
+        response = requests.get(
+            URL_CALENDAR.format(premise_id=self._premise_id),
+            headers=self.default_headers,
+        )
+        soup = BeautifulSoup(response.content, "html.parser")
+
+        items = (
+            jq.compile(
+                "[. as $m|range(0;$m|length;2)|[{date:($m[.+1]),type: $m[(.)]}]]|add"
```
mampfes, line 84: Wow, this looks really complex :-).

 jq is another dependency and I prefer to use built-in packages as much as possible. Therefore, could this be done also with built-in json?

custom_components/waste_collection_schedule/waste_collection_schedule/wizard/irnt_no.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
+
+import urllib.parse
+
+import inquirer
+import requests
+
+URL_ADDRESS_SEARCH = "https://innherredrenovasjon.no/wp-json/ir/v1/addresses/{search}"
```
mampfes, line 8: I really appreciate that you created a wizard to get the magic premise_id. But from my experience a lot of users have problems running such a wizards. Nowadays I would prefer to specify the address as source argument and let the source query the premise_id internally.

mampfes (changes requested):

mampfes:
Closed because no response from originator.

PR #672, opened 2023-02-09
Add Support for App that are Developed by cmcitymedia.de

Newspicel:
cmcitymedia.de distributes custom apps for German cities and municipalities, all using the same backend. This is the implementation of the waste backend, which you can find here: https://sslslim.cmcitymedia.de/v1


mampfes:
Thanks, great job! However, from my experience, the solution with the wizard will not work well - users are overstrained with that. I was wondering if we could add all information to the .md file? A script based on your wizard should be easy to develop. I did s.th. similar with the update_docu_links.py script in the top level directory. The new script should add all hpid mappings and the district mapping (if needed). That's all, right?
What do you think?

mampfes:
Thanks for your contribution.

mampfes:
@Newspicel: I still see an SSL error on my Ubuntu test machine (with Python 3.10.6):

```
$ ./test_sources.py -s cmcitymedia_de
Testing source cmcitymedia_de ...
  Blankenheim - Bezirk F: Lindweiler, Rohr failed: HTTPSConnectionPool(host='sslslim.cmcitymedia.de', port=443): Max retries exceeded with url: /v1/415/waste/41500/dates?district=1371 (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:997)')))
  Blankenheim failed: HTTPSConnectionPool(host='sslslim.cmcitymedia.de', port=443): Max retries exceeded with url: /v1/415/waste/41500/dates (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:997)')))
  Oberstadion failed: HTTPSConnectionPool(host='sslslim.cmcitymedia.de', port=443): Max retries exceeded with url: /v1/447/waste/44700/dates?district=1349 (Caused by SSLError(SSLError(1, '[SSL: WRONG_SIGNATURE_TYPE] wrong signature type (_ssl.c:997)')))
```

I already did an Google search, but didn't find a solution (except the one you already have). Any ideas?

Newspicel:
@mampfes Sorry for replying so late, wanted to write a comment but I forgot. 
I know the bug, their API uses a old SSL configuration. Actually, I thought I had fixed it with the `TLSAdapter` class. https://stackoverflow.com/a/69580908 https://github.com/psf/requests/issues/4775
I can't get the bug, it works on my MacBook and on my Raspberry Pi fine. 
Which openssl version do you have on your ubuntu system?

Newspicel:
@mampfes Hey, I fixed the bug in #1089. So I don't know what the error was or how to fix it, but I find a way around it.

PR #673, opened 2023-02-10
added source for zaw.online.de

5ila5:
added source for zaw.online.de the serving Darmstadt-Dieburg Hesse, Germany.
Unoriginality you cannot search directly for your address but you have to get a unique identifier for your city and street from the HTML of their website


mampfes:
Thanks for your contribution.

Issue #697, opened 2023-02-22
Request: Entsorgungs- und Wirtschaftsbetrieb Landau (in der Pfalz)


Kurisutian:
After looking to integrate the waste schedule for this provider a friend and I kinda started using the ICS integration since the website creates a unique link to what to download. However, part of that link seems to be a random session id which we are not sure of if this remains or changes over time so at one point the download no longer works. 
Not being a programmer I wanted to ask if someone more skilled can take a look at it. 
The website can be found here: https://www.ew-landau.de/Kurzmen%C3%BC/Entsorgungskalender/

When choosing a street and building number a link like this will be created: https://apps.c-trace.de/web.landau/(S(db5sn4tg0wstgioyfdcvhg0v))/abfallkalender/downloadcal?Strasse=Am%20Kindergarten&Hausnr=1

So it contains the street name but also that session id part (S(<SessionID>)) which we are not sure of whether it can be random so we can stick to the ICS download path of if it would be worth having a distinct integration for the provider.

If somebody can advise me if we should stick to the ICS download way or could alternatively look into coding a integration for this that would be awesome. Thanks in advance!

5ila5:
there is already an existing integration for these c-trace.de urls:
[c-trace](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/c_trace_de.md)

but it seems like it only works for web.c-trace.de urls at the moment I made a pull request so that this might be possible:
```yaml
waste_collection_schedule:
  sources:
  - name: c_trace_de
    args:
      strasse: Am Kindergarten
      hausnummer: 1
      service: web.landau
      subdomain: apps
      ical_url_file: downloadcal
```

but I'm not the biggest fan of the ical_url_file so maybe there will be some changes

PR #701, opened 2023-02-22
updated c_trace_de to support other than web.ctrace.de urls

5ila5:
I updated c_trace_de to also allow other subdomains (see #697) and formats because there where only a few changes to be made  and I personally think it would not make sense to make a new integration, there are a few more like this one out there (googling: `site:apps.c-trace.de`)

but I'm not sure if this are to many possible arguments. (especially ical_url_file feels kind of janky)  I'm open to changes or other Idears

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_de.py
@@ -27,6 +27,13 @@ TEST_CASES = {
         "hausnummer": 7,
         "service": "augsburglandkreis",
     },
+    "landau": {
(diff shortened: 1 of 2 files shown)


mampfes:
I also don't really like the extra `ical_url_file` argument. My proposal would be to add a mapping table which maps the service name to the ical file name. And I would also remove the `web.` from Landau (maybe using the same mapping table).

5ila5:
- I added a Mapping table
- added all apps.c-trace.de municipalities in the mapping table
-  and added all municipalities I could find by googling into the table in the md file (apps.`...` und web.`...`)

mampfes:
Great. Thanks for your contribution.

PR #706, opened 2023-02-25
Hume City Council

seedzero:
Add Hume City Council waste collection schedule


seedzero:
Just noticed that Whittlesea city council has changed in the Readme, this is from running ./update_docu_links.py

mampfes:
Thanks for your contribution.

PR #709, opened 2023-02-26
Corrected Recycling & Green Waste

seedzero:
Corrected Recycling & Green Waste and added additional test case


mampfes:
@seedzero  Hi, did you close this PR by intend? Because thhe change makes sense to me.

seedzero:
Hi @mampfes, I did, but only to open another due to your refactors which I hadn't seen. Sorry.

Issue #413, opened 2022-10-27
Request for provider!  Cardinia, Victoria, Australia


vytux-com:
Can we please add this to the source list?

https://www.cardinia.vic.gov.au/info/20002/rubbish_and_recycling/385/bin_collection_days_and_putting_your_bins_out/2

skipishere:
Looks like the initial search using "Daly Rd, Nar Nar Goon Vic" as an example hits the page https://geocoder.cit.api.here.com/6.2/search.json?app_id=pYZXmzEqjmR2DG66DRIr&app_code=T-Z-VT6e6I7IXGuqBfF_vQ&country=AUS&state=VIC&&searchtext=Daly%20Rd%2C%20Nar%20Nar%20Goon%20Vic
(the App id and app code seem to be needed, unsure if they change with usage)

```json
{
    "Response": {
        "MetaInfo": {
            "Timestamp": "2022-11-08T23:02:48.131+0000"
        },
        "View": [
            {
                "_type": "SearchResultsViewType",
                "ViewId": 0,
                "Result": [
                    {
                        "Relevance": 0.86,
                        "Distance": 2886.4,
                        "MatchLevel": "street",
                        "MatchQuality": {
                            "Country": 1.0,
                            "State": 1.0,
                            "City": 1.0,
                            "Street": [
                                1.0
                            ]
                        },
                        "Location": {
                            "LocationId": "NT_CwL.5zhRWw6bCkcZtyEzDB",
                            "LocationType": "address",
                            "DisplayPosition": {
                                "Latitude": -38.10658,
                                "Longitude": 145.60191
                            },
                            "NavigationPosition": [
                                {
                                    "Latitude": -38.10658,
                                    "Longitude": 145.60191
                                }
                            ],
                            "MapView": {
                                "TopLeft": {
                                    "Latitude": -38.10543,
                                    "Longitude": 145.58621
                                },
                                "BottomRight": {
                                    "Latitude": -38.11707,
                                    "Longitude": 145.61092
                                }
                            },
                            "Address": {
                                "Label": "Daly Rd, Nar Nar Goon VIC 3812, Australia",
                                "Country": "AUS",
                                "State": "VIC",
                                "City": "Nar Nar Goon",
                                "Street": "Daly Rd",
                                "PostalCode": "3812",
                                "AdditionalData": [
                                    {
                                        "value": "Australia",
                                        "key": "CountryName"
                                    },
                                    {
[...]

PR #712, opened 2023-02-27
Add Cardinia Shire Council cardinia_vic_gov_au

seedzero:


mampfes:
Thanks for your contribution.

mampfes:
> The Cardinia Shire council website only provides the next available dates. Due to the way the code works on their site, more dates could easily be provided. [The feature request](https://github.com/mampfes/hacs_waste_collection_schedule/issues/413) provides information on how this works, basically a start date is provided for each waste type along with the number of weeks (every week or every second). Should the data be extrapolated further?

Hm, interesting question. If the data is reliable I think is could be helpful to extrapolate the new 1-3 months.

PR #713, opened 2023-02-27
Correct mdi icon for "Rubbish"

seedzero:
alias listed on mdi icons doesn't resolve, so corrected icon name


mampfes:
Thanks for your contribution.

PR #719, opened 2023-02-28
hume_vic_gov_au: Optional prediction of future collections

seedzero:
Allow optional prediction of future collections using the predict boolean


mampfes:
Thanks for your contribution.

PR #736, opened 2023-03-05
MidSussex update

dt215git:
Update to midsussex_gov_uk.py: Empty string default values supplied to correct issues flagged by @thor9898


mampfes:
Thanks

mampfes:
@dt215git Hi, is there a way to contact you via private message?

PR #742, opened 2023-03-06
Added new source samiljo_se and updated readme.md+info.md

Lance0095:
Added new source script and markdown for samiljo_se. Updated README.md+info.md using update_docu_links.py. Output from test_sources.py appended below.

gustaf@DESKTOP-VQOFJKP:/mnt/c/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/test$ python3 test_sources.py -s samiljo_se
Testing source samiljo_se ...
  found 30 entries for Test1
  found 46 entries for Test2
  found 43 entries for Test3


mampfes:
Thanks for your contribution.

After merging, I discovered 2 minor issues:

1. `Fyrfackskärl 1 (Helgvecka - hämtningsdagen kan avika)` is quite long and and will be a pain if someone wants to set an alias for it. Can you come up with a shorter name?
2. For test case 3 is get the following error on my machine. Do you see the same error?

```
./test_sources.py -s samiljo_se -t
Testing source samiljo_se ...
  found 30 entries for Test1
  found 46 entries for Test2
  Test3 failed: 'NoneType' object has no attribute 'contents'
      File "/home/steffenz/tmp/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/test/./test_sources.py", line 84, in main
        result = source.fetch()
      File "/home/steffenz/tmp/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/source/samiljo_se.py", line 104, in fetch
        str(wasteday_wastetype.find("div", "styleInteIdag").contents).strip(
    AttributeError: 'NoneType' object has no attribute 'contents'
```

Lance0095:
Yes, today I have the same error. Apparently the tag changes on collection
day. Will issue a pull request with a fix.

For the naming I see two approaches. Using the short coded name without the
NAME_MAP (e.g. HKARL1 and HKARL1-H) and letting the user use an alias in
Home Assistant. Or as you suggested a shorter naming scheme e.g.
"Fyrfackskärl 1 - Helgvecka".

The long name is due holidays "Fyrfackskärl 1 (Helgvecka - hämtningsdagen
kan avvika)" translates to "Four compartment bin 1 (Week containing a
public holiday - collection day might deviate"


Den tis 7 mars 2023 kl 07:36 skrev Steffen Zimmermann <
***@***.***>:

mampfes:
Ok, I see.

In general I prefer that the users set their favorite names via the alias option - and there is no name mapping in the sources. But the precondition is that the users are able to understand the original names and can map them to something else. And with "HKARL1" and "HKARL1-H" I have my doubts. Or what do you think?
So in that case (in which the user can't map the original name provided by the source) cannot map, it is ok for me to add a simple name mapping in the source. "Fyrfackskärl 1 - Helgvecka" would be ok for me. Maybe it is worth to add a note to the documentation aso.

Lance0095:
The short coded name are for sure more difficult to understand. Will add a note in documentation regarding "Helgvecka" and the meaning of it.

Another problem is the unknown waste/bin types. There are several waste/bin types that i don't have a more common name (42 to be exact, list appended below). As long as I don't have a test case were these are included I won't be able to decode them to a more common name. 
I have made a solution that would in the case of a bin not matching the NAME_MAP output the short coded name instead. I don't know if this is the best solution. 
+Works for all bin types, with the caveat that some will have the short coded name.
-Breaking change for anyone using the short coded name if one would add one or more of these to the NAME_MAP.

if waste

Another solution would be to output nothing for unknown types and then add the new types when a user issues a complaint. One could in this case have a print statement that would show when running test_sources.py
+Won't work for any types not in NAME_MAP.
-No breaking changes.

if waste print


HKARL1-F
HKARL1-HK
HKARL1-HK-H
HKARL1-HK-F
HKARL2-F
HREST-F
HREST-HK
HREST-HK-H
HREST-HK-F
HOSORT-F
HMAT-F
FKARL1
FKARL1-F
FKARL1-H
FKARL1-HK
FKARL1-HK-H
FKARL1-HK-F
FKARL2
FKARL2-H
FREST
FREST-F
FREST-H
FREST-HK
FREST-HK-H
FREST-HK-F
FOSORT
FOSORT-H
FOSORT-F
FMAT
FMAT-H
FMAT-F
TRG
TRG-H
TRG-F
MAT
MAT-H
MAT-F
SKARL1
SKARL2
SREST
SMAT
SOSORT

Lance0095:
May have found a solution to the unmapped names. I'm searching through the address database for possible test cases not matching the NAME_MAP. Hoping to not get banned during this.

Lance0095:
I have now searched through all addresses and mapped the new bin types I found with a common name. There are still a few missing (24) but there are currently no street addresses that are using them.

Lance0095:
Do you want me to add the search script I made for mapping the different bin types? If so where?

mampfes:
Yes, this would be great. You can add it to the services folder.

PR #757, opened 2023-03-12
adding swindon_gov_uk

joshwillcock:
New source for Swindon Borough Council.
Provides Waste, Plastic, Recycling & Green bin (if subscription exists).
4 test cases currently have 2 with green waste subscription & 2 without.


mampfes:
Thanks for your contribution.

Issue #727, opened 2023-03-02
Request: Mein-Abfallkalender.de


TheBassfly:
I would love to see an integration for the "mein-abfallmalender.de"

I couldn't make the ICS Import to work correct.

Thank you:

https://eppstein.mein-abfallkalender.de

mampfes:
What was the problem with ics?

I did a short test with mein-abfallkalender.de and the ics source and it worked:
Select your adress in the web page, then go to "Termine iCal / WebCal". Set "Erinnerung" to "keine Erinnerung" and set "Verwendung" to "Termine via WebCal Protokoll abonnieren (Allgemein)". You then get a webcal and a http link. Use the http link with the ics source.

Example:
http://eppstein.mein-abfallkalender.de/ical.ics?sid=19799&cd=inline&ft=noalarm&fu=webcal_other&fp=next_30&wids=494,495,496,498,497,499,502,500,513,501&uid=263834&pwid=d9caede3b1&cid=80

TheBassfly:
Maybe I need some advanced help.

The Integration got an error in the log:

2023-03-06 18:51:23.153 WARNING (SyncWorker_3) [homeassistant.loader] We found a custom integration waste_collection_schedule which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2023-03-06 18:51:23.157 WARNING (SyncWorker_3) [homeassistant.loader] We found a custom integration govee which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2023-03-06 18:51:23.159 WARNING (SyncWorker_3) [homeassistant.loader] We found a custom integration browser_mod which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2023-03-06 18:51:23.161 WARNING (SyncWorker_3) [homeassistant.loader] We found a custom integration eufy_security which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
2023-03-06 18:51:23.163 WARNING (SyncWorker_3) [homeassistant.loader] We found a custom integration ics_calendar which has not been tested by Home Assistant. This component might cause stability problems, be sure to disable it if you experience issues with Home Assistant
[...]

PR #761, opened 2023-03-14
Adding mein abfallkalender

deese:
I've added support for mein-abfallkalender.de source. 
This providers provides iCal for different cities. I've tested it with bad-vilbel and hofheim.

Thanks!


mampfes:
Hi, I really appreciate all the effort you spent in the integration of mein-abfallkalender. I have seen that they also provide a webcal link (which could be used with the ics source or with a dedicated source). So I was wondering if it wouldn't be the better choic to let the users generate a webcal link and use this as an argument for the source instead of using the wizard or the hardcore variant. From my experience, even the wizard is too complex for many users.

mampfes:
I tried the webcal link with the ICS source and it works perfect. There are 2 important things to mention:

1. Don't forget to add your email address, otherwise the generated webcal links don't work.
2. Replace webcal:// with https:// when the adding to the configuration.

For me, this approach looks much simpler than using a wizard of extracting 5(?) argument from the webcal URL.

t-pohl:
Just wondering because this is exactly what I need: How do I integrate now a city which uses "mein abfallkalender"? Couldn't figure that out so far as I also cannot find it in the [Supported Service Providers](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/README.md#supported-service-providers).

mampfes:
@t-pohl mein-abfallkalender.de is supported via the generic ICS source - see description above.

Example configurations for ICS are not listed in the Supported Service Providers list so far. I already have this in mind, but didn't find a good solution so far.

t-pohl:
Got it to work using [this steps here](https://github.com/mampfes/hacs_waste_collection_schedule/issues/727). Would be cool to make this more prominent in the README. Otherwise really great update, thank you very much!

PR #762, opened 2023-03-15
fix: add workaround for wrong kwu deployment

paschdan:
adds a workaround for the current issue on the kwu calender, where they have deployed a version pointing to local test system


mampfes:
Thanks for your contribution.

PR #768, opened 2023-03-16
Added service samiljo_se_wastetype_searcher.py

Lance0095:
samiljo_se_wastetype_searcher.py is a servicescript used to map new waste types not yet in NAME_MAP.


mampfes:
Thanks, Lance0095

Issue #763, opened 2023-03-15
Request: can someone please add Felsberg


HolliDolli:
I did my best to understand the code, but I can't figure out to add a datasource. 
Can someone add Felsberg https://www.zva-sek.de/online-dienste/abfallkalender-2023/abfallkalender-2023.html for me?
I will take care for the upcoming years (changing existing code is much more easy :-) ), but I'm not able to transfer from another location.

Thanks.

Holger

Edit: And I forgot to mention: There is an additional problem: They added the date to the summary-text.
Testcases might be:
Felsberg/Felsberg and
Guxhagen/Guxhagen

mampfes:
fixed by #771

PR #771, opened 2023-03-17
added zva_sek_de #763

5ila5:
#763


mampfes:
Thanks for your contribution, 5ila5.

Issue #767, opened 2023-03-16
last Saturday every month 


scns:
Is it possible to make an static rule for every last Saturday every month?  or need i to make static dates?

mampfes:
I think a rule for last XXX day is not possible with the current `static` source. You have to go with static dates --> still use the `static` source and add them to the `dates` section.

scns:
Thx for your answer. i will do it that way for now!

mampfes:
The static source has now support for weekdays.

(Please wait for the next release).

PR #776, opened 2023-03-19
Enhance static source to be able to set reoccurring dates based on Weekdays

5ila5:
requested by #767.
It's not that difficult to set those kind of reoccurring events using `rrule`. I'm not sure if this is the best way to set these in the yaml file so feel free to suggest changes


mampfes:
Thanks for your contribution.

mampfes:
I slightly refactored the yaml config and removed alternative ways of configuration to make the code a little bit easier.

PR #785, opened 2023-03-20
Fix missing dates in meinawb_de 

Andre0512:
I found some missing collection dates around holidays and a wrong key in the example.


mampfes:
Thanks for your contribution.

PR #787, opened 2023-03-20
Adding uppsalavatten se

Lance0095:
Added new source script and markdown for uppsalavatten_se. Updated README.md+info.md using update_docu_links.py. Output from test_sources.py appended below.

$ python3 test_sources.py -s uppsalavatten_se -l -i
Testing source uppsalavatten_se ...
  found 2 entries for Test1
    2023-04-05: Restavfall [mdi:trash-can]
    2023-03-22: Matavfall [mdi:food-apple]
  found 4 entries for Test2
    2023-03-31: Restavfall [mdi:trash-can]
    2023-10-01: Slam []
    2023-04-01: Slam []
    2023-10-01: Slam []
  found 2 entries for Test3
    2023-03-31: Restavfall [mdi:trash-can]
    2023-03-31: Matavfall [mdi:food-apple]


Lance0095:
It would be possible to get more collection dates if one would parse a pdf instead of a json. But that would add more complexity and add another package to the requirements.

mampfes:
Thanks for your contribution.

I don't want to add PDF parsing, because this is even more fragile than web scraping.

Issue #791, opened 2023-03-21
Item "slam" from Sysav breaks the component


epejans:
Getting waste collection schedule from Sysav where this includes "slam", breaks the component.
Sysavs API responds with a week rather than a date for this situation, i.e. "v42 Okt 2023"
Code only handles dates.

(Should be sufficient to check if date, proceed, if string don't do anything (or try to find Monday in the week...?))

mampfes:
Could you provide a test adress which evokes the error?

epejans:
HI,
Sure!
Ekhagsvägen 204, Svedala

Den tis 21 mars 2023 kl 20:19 skrev Steffen Zimmermann <
***@***.***>:

epejans:
[image: image.png]

Den ons 22 mars 2023 kl 08:08 skrev Peter Jansson ***@***.***>:

mampfes:
fixed by #793

epejans:
Awesome, thanks!
/Peter

Den tors 23 mars 2023 20:28Steffen Zimmermann ***@***.***>
skrev:

PR #793, opened 2023-03-22
fixed sysav_se where waste type SLAM would raise an error

5ila5:
because the api returns a week instead of a date
fix for #791 
returns the correct week if normal ISO format is expected and not US format


mampfes:
Thanks for your contribution.

PR #798, opened 2023-03-23
Update AbfallIO.py

macrojames:
added VIVO Landkreis Miesbach to AbfallIO wizard


mampfes:
Thanks for your contribution.

macrojames:
I think they switched to another product.
Their website doesn't use the AbfallPlus Api anymore but the app looks like it still does.
The Website has a custom domain with a "WasteCollectionServlet" packaged in "com.athos.kd.warngau.CheckAbfuhrTermineParameterModel"
The old API does still respond but does not deliver dates in 2023. :(

Issue #792, opened 2023-03-21
Source Request: East Northamptonshire 


andrewchown:
East Northamptonshire webpage is here: https://www.east-northamptonshire.gov.uk/info/200084/recycling_rubbish_and_waste/1936/bin_collection_day

If it could be added that would be great!

mampfes:
fixed by #801

PR #801, opened 2023-03-25
added east_northamptonshire_gov_uk

5ila5:
#792

"translated" the original javascript. for some reason they use 2022-6-20 as reference


mampfes:
Thanks for your contribution.

Issue #269, opened 2022-07-01
Is it possible to use this for the Landkreis Vechta? 


HumongousFungus85:
The AWV (Abfallwirtschaft Vechta) uses Abfallplus somehow, however it uses it's own App.
I can not use the original Abfallplus app and it prompts me to use the AWV app instead if I try.
Is it still possible to use this integration with it? Since I can not get the required IDs.
[AWV Website](https://www.abfallwirtschaft-vechta.de/)

Thanks in advance

markluebbehuesen:
Hey,

I am from Oldenburg. I have generated a ICS file and it works as expected.

Here you can create an ICS file, which you can import again.
https://www.abfallwirtschaft-vechta.de/index.php/termine/abfuhrkalender 

Mark

HumongousFungus85:
Hi,

Yes I got it to work with the ics file as well.

However, I would prefer to set it up through the Abfallplus API, so it will constantly update itself in case there are some changes, or when the year changes.

mampfes:
fixed by #819

PR #819, opened 2023-03-29
added abfallwirtschaft_vechta_de

5ila5:
wanted by #269

Does not  currently return any entries  because they did not publish any pickup dates after March. But tested it with the dates in the past.

current message on the web page:
```
Wichtige Info zu den neuen Abfuhrterminen ab 01.04.2023

Die Abfuhrtermine und neuen Abfuhrbezirke, die ab dem 1. April 2023 gültig sind, sind aktuell noch nicht im AWV-Onlinekalender und der AWV Abfall-App abrufbar. Sie werden rechtzeitig bis zum 1. April 2023 veröffentlicht
```


mampfes:
Thanks for your contribution.

Issue #670, opened 2023-02-08
Request: Stockton On Tees, UK


covamalia:
https://www.stockton.gov.uk/bin-collection-days

I have also dropped an email to their digital mailbox, [digital@stockton.gov.uk](mailto:digital@stockton.gov.uk) to see if there is anything open to the public that can be used, so will see if that gets any traction and update if I get anywhere that way.

5ila5:
fixed by #830

PR #830, opened 2023-04-01
adding stockton_gov_uk

5ila5:
wanted by #670


custom_components/waste_collection_schedule/waste_collection_schedule/source/stockton_gov_uk.py
```diff
@@ -0,0 +1,104 @@
...
+from bs4 import BeautifulSoup
+from waste_collection_schedule import Collection
+
+# Source code based on gateshead_gov_uk
+
+TITLE = "Stockton-on-Tees Borough Council"
+DESCRIPTION = "Source for Stockton-on-Tees Borough Council."
+URL = "https://www.stockton.gov.uk/"
+TEST_CASES = {
+    "100110203615": {"uprn": "100110203615"},
```
mampfes, line 16: As we had this issue multiple times now: Could you add a test case with an integer (instead of string)?

Issue #829, opened 2023-04-01
AbfallNavi missing Fraktion


flotomation:
I noticed something odd:

With AbfallnaviDE there is an issue with my waste type "Papier", the integration isn't fetching it not correctly.
I use service_id: oberhausen 
The calendar is filled with all the types/fraktions except Papier. 
When I use the API manually or with the site WBO Wirtschaftsbetriebe Oberhausen, I get the correct dates, even with the Papier Type. 

WBO:
<img width="729" alt="image" src="https://user-images.githubusercontent.com/129587589/229286124-129d0b32-0c80-4030-b6eb-e62636ba4e38.png">

HA Calendar:
<img width="926" alt="image" src="https://user-images.githubusercontent.com/129587589/229286206-0d72c7f1-c471-4acd-98a8-8453761331cb.png">

Could you please look into it?

5ila5:
there seems to be an error with the `hausnummer` parameter i get a lot more entries if `hausnummer` is an Integer, seems like providing hausnummer as Iteger e.g:
```yaml
waste_collection_schedule:
  sources:
    - name: abfallnavi_de
      args:
        service: oberhausen
        ort: Oberhausen
        strasse: Blumenthalstraße
        hausnummer: 22 #instead of "22"
```
seems to be a work around. can you try if it helps?
but I will check why this happens

flotomation:
I guess hausnummer was the issue, I had an integer configured. With a string and "hausnummer" in my config, I now get the behavior I was expecting and my Papier is up and running again. Thank you

PR #832, opened 2023-04-01
allow integer haunummer arguments in abfallnavi_de

5ila5:
A general question: how should something like this be handled? Should every argument only allow a single datatype? Especially for house numbers you might accidentaly use an integer instead of a string.
Here in abfallnavi_de if `hausnummer` is an int, it is handles as if `hausnummer` was  `None` (because it does not find the house number in the returned json), but this does not raise an error but returns valid but incomplete or wrong values (see #829)


mampfes:
I would also do it exactly as you proposed it in the PR. The user should not care about integers or strings, we should handle this internally and give the service-provider api whatever they want.

mampfes:
And thanks for your contribution.

PR #834, opened 2023-04-02
Add full calendar and service-based icons to HVCgroep source

se-bastiaan:
A few days ago I installed this integration, and I noticed that only the upcoming dates from by waste calendar are available.
Since the website I use for this information has more than just the upcoming dates available I started on a journey to get this in my Home Assistant calendar.

This resulted in the changes I am proposing in this PR.

The following things were adjusted in this source:
 - I have added a few more test cases.
 - Each of the services that use this system has slightly different icons, so I have added those to each service individually. That way the icon mapping also works for other systems than the original HVCgroep that this integration was written for.
 - I have added code to handle multiple addresses to be returned from the API. BagIds are unique and some addresses may have more info than just the postal code and house number (one such address is in the example). There is also a house letter and a suffix available in the API, which makes it possible for people to select the right bagId for their address by providing more information.
 - The API has an endpoint available that lists all waste collection dates for the current calendar year, including those that are available as upcoming dates. By using this endpoint the source can provide the framework with more information and will completely fill the calendar with all waste collection dates available for the current year.

While changing things for this source I stumbled across a few small things as well:
 - Stadwerk072 and RMN do no longer use the Opzet afvalkalender software, so those are removed
 - Spaarnelanden is not Spaarne Landen, so that is corrected

If I weren't lazy I would have split this in multiple PRs, since it is quite a list of changes. Sorry 😅


mampfes:
Great, thanks for your contribution!

custom_components/waste_collection_schedule/waste_collection_schedule/source/hvcgroep_nl.py
```diff
@@ -169,14 +306,25 @@ def fetch(self):
             raise Exception("no data found for this address")
 
         bag_id = data[0]["bagid"]
+        if len(data) > 1 and self.house_letter and self.postfix:
+            _LOGGER.info(f"Checking {self.house_letter} {self.postfix}")
+            for address in data:
+                if (
+                    address["huisletter"] == self.house_letter
+                    and address["toevoeging"] == self.postfix
```
wouterhardeman, line 314: Hi, I happened to configure the plugin yesterday with this service and came across this pull request.

Where does postfix come from? I see 3 names:
- addition (in docs)
- suffix on line 289 and 294
- postfix here on line >309

It is also possible to have no house letter (huisletter) but to have a postfix/addition/suffix (toevoeging).

The data[0]["bagid"] should not make this a breaking change but this seems incorrect.

se-bastiaan: good catch. I couldn't decide on the right name so changed it around a few times and probably didn't copy the last version of my code into my local git repo after some testing 😅

PR #837, opened 2023-04-03
Added chiltern.gov.uk source

kanthamohan:
Added chiltern.gov.uk source that covers parts of Buckinghamshire, UK


custom_components/waste_collection_schedule/waste_collection_schedule/source/chiltern_gov_uk.py
```diff
@@ -0,0 +1,74 @@
...
+from waste_collection_schedule import Collection
+
+TITLE = "Buckinghamshire Waste Collection - Former Chiltern, South Bucks or Wycombe areas"
+DESCRIPTION = "Source for chiltern.gov.uk services for parts of Buckinghamshire"
+URL = "https://chiltern.gov.uk"
+
+TEST_CASES = {
+    "Test_001": {"uprn": "200000811701"},
+    "Test_002": {"uprn": "100080550517"},
+    "Test_003": {"uprn": "100081091932"},
```
5ila5, line 16: could you please add an additional test case (or convert one of them) to an Integer, as this was a problem multiple times in the past (you Example also uses an Integer value)

kanthamohan: Added new commit with an integer test case which passes with the test script

5ila5:
Rebased your fork on master.

Thanks for your contribution.

Issue #839, opened 2023-04-04
EcoHarmonogram problem with importing when are 2 matching harmonograms


sebastian-bugajny:
EcoHarmonogram returns 2  harmonograms with the same name like:

❯ python3.9 EcoHarmonogramPL.py Częstochowa "" Sokola 12b | grep Gnaszyn
 Gnaszyn-Kawodrza - Zab. jedn. (N-ry parz.)
 Gnaszyn-Kawodrza - Zab. jedn. (N-ry parz.)

Based on Ecoharmonogram App it shows 2 harmonograms:
1. 01.01.2023 - 31.03.2023
2. 01.04.2023 - 31.10.2023

It looks like when I have integration configured like that:

waste_collection_schedule:
  sources:
    - name: ecoharmonogram_pl
      args:
        town: Częstochowa
        street: Sokola
        house_number: 12b
        additional_sides_matcher: Gnaszyn-Kawodrza - Zab. jedn. (N-ry parz.)

I don't see any pending waste schedules. I suppose it imports only the first matching schedule and the second one is skipping. If there a way to fetch both of them.

sebastian-bugajny:
@5ila5 I've tested your changes and now it works perfectly. Thank you.

mampfes:
fixed by #840

PR #840, opened 2023-04-04
fixing ecoharmonogram_pl to get events from multiple schedule periods

5ila5:
fix for #839
The old code iterated over the schedule periods but returned after the first (or returned only the last if additional_sides_matcher is not set). This does not seem to be the wanted behaviour. This increases the found Events (test cases) From:
```
  found 158 entries for Simple test case
  found 234 entries for Sides multi test case
  found 18 entries for Sides test case
  found 397 entries for Sides multi test case with district
  found 156 entries for Simple test with community
```
to
```
  found 158 entries for Simple test case
  found 782 entries for Sides multi test case
  found 78 entries for Sides test case
  found 397 entries for Sides multi test case with district
  found 156 entries for Simple test with community
  ```
Its hard for me to validate if all of them are actually wanted events. @pawelhulek created and improved this source, so maybe he wants to take a look at this before merging.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
@@ -44,37 +44,42 @@ class Source:
 
         entries = []
         for sp in schedule_periods:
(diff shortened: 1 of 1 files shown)


5ila5:
As there is no further response and this seems to fix an issue, I will merge this now.

PR #844, opened 2023-04-05
Fix naming and documentation of new args for HVCgroep

se-bastiaan:
@wouterhardeman found some mistakes in my [last PR](https://github.com/mampfes/hacs_waste_collection_schedule/pull/834). So here are some fixes :)


mampfes:
Thanks for fixing.

PR #846, opened 2023-04-06
Ics docu

mampfes:
This PR adds support for ICS source example configurations. This addresses the following problem:

There are a lot of service providers which are fully supported by the generic ICS source. Especially if the service provider offers a "ical subscription" or at least provide a static link which doesn't change over time, the generic ICS source is a match. However, a lot of users don't know that their service provider is fully supported, because they don't show up in the README.md. Sometimes users also complain if there is no dedicated source, because the feel like a second choice (which is not true, because a lot of dedicated source internally use the same service as the ICS source.

With this PR, support for ICS source based service providers can be easily added. Just create a .yaml file with title, url, and at least one test-case and run update_docu_links.py. This will create a corresponding markdown file and add the service provider to README.md and info.md.
In addition, test_sources.py has been enhanced to test the test-case in the .yaml file.
`test_sources.py -I` tests all .yaml file.
`test_sources.py -y husel` tests the .yaml file husel.yaml.


mampfes:
@5ila5 Any comments on this?

doc/source/ics.md
5ila5, line 1: Could it be useful to also add and automatically update a "tested with" section in the ics.md file, to better communicate, that lots of providers are supported by this source and to give more examples?

5ila5 (commented):
I like the idea and implementation. I thought about how we could better display supported ICS providers myself.

doc/source/ics.md
mampfes: good idea. will do

Issue #816, opened 2023-03-29
Request: aha Zweckverband Abfallwirtschaft Region Hannover


Myjestix:
Hi,
are there any plans to add "aha Zweckverband Abfallwirtschaft Region Hannover" to the integration?
https://www.aha-region.de/abholtermine/abfuhrkalender

They offer an ICS download but I would rather not have to update the ICS files manually.

Thanks!

Zeptran:
Hi,

I've discovered a manual process to trick the AHA website into exposing the iCal file for download. Here's an example YAML config for getting the data (with my address changed):

```
waste_collection_schedule:
  sources:
    - name: ics
      args:
        url: https://www.aha-region.de/abholtermine/abfuhrkalender
        method: POST
        params:
          strasse: "00001@Aachener Str. / Südstadt@Südstadt"
          hausnr: 2
          ladeort: "00001-0002 "
          ical_android: 10
      customize:
        - type: Abfuhr Restabfall # from calendar event name
          alias: Restabfall # matches type in sensor configuration above 
```

The tricky part is getting the proper params, which requires making a request on the AHA website while the dev window is open on a browser. In the Network tab, make a request and scroll all the way up again. The params should be under Payload. When entering the ladeort, make sure to include a space after, otherwise it won't work.

<img width="954" alt="image" src="https://user-images.githubusercontent.com/21012208/230723068-ba985bd2-bc16-45fb-b647-ff1dad3073f2.png">

It took a while to figure out how to get the different trash types, but it depends on the ical_android number:
10 = Restabfall
33 = Leichtverpackung (Gelber Sack / Gelbe Tonne)
24 = Papier
70 = Bioabfall

Just paste the YAML snippet with the different ical_android number under sources, and it should just work.

For the sensor config part, I used the following:

```
sensor:
  #das hier ist für den Abfuhrkalender
  - platform: waste_collection_schedule
    name: next_collection  
    source_index: [0, 1, 2, 3]
  # next yard waste collection
  - platform: waste_collection_schedule
    name: next_Restabfall_collection
    types:
      - Restabfall
    source_index: 0
  - platform: waste_collection_schedule
    name: next_Biomüll_collection
    types:
      - Biomüll
    source_index: 3
  - platform: waste_collection_schedule
    name: next_Papier_collection
    types:
      - Papier
    source_index: 2
  - platform: waste_collection_schedule
    name: next_Gelber-Sack_collection
    types:
      - Gelber-Sack  
[...]

PR #863, opened 2023-04-10
added aha_region_de ICS source

5ila5:
It's quite complicated to integrate and you will have multiple calendars. 
Would it be best to write a source? 
would probably not be that complicated.

based on #816


mampfes:
OK, then I would go for a dedicated source.

5ila5:
closing in favour of a real source #867

PR #869, opened 2023-04-11
Update fife_gov_uk.md

apbarratt:
Thank you so much for your work adding Fife to the tool.
I've added a more advanced example that features Fife's bin type names and a quicky example of sensors for them too. 

I noticed that you include a type for bottles in your code, though I'm not aware of a bottle collection.  Happy to add to the example though if this is the case :)

I've also added the yaml I used to create a widget on my dashboard.  I just thought it a nice example and wanted to share what I was using but do say if you feel this isn't appropriate here.

### DIFF
--- a/doc/source/fife_gov_uk.md
+++ b/doc/source/fife_gov_uk.md
@@ -33,3 +33,80 @@ waste_collection_schedule:
 
 An easy way to discover your Unique Property Reference Number (UPRN) is by going to <https://www.findmyaddress.co.uk/> and entering in your address details.
 Otherwise you can inspect the web requests the Fife Council website makes when entering in your postcode and then selecting your address.
+
+### A more complete example with a dashboard card
+
+Here's a more complete example of the sources that adds pictures and shorter names to the types that come back from Fife's API:
+
+```yaml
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your work.

- I probably forgot to remove the Glass key from my template.
- I'm not sure if all of this should be part of the documentation of a source. I couldn't find any other source with this kind of documentation

dt215git:
Might be a good example to add to the faq .md showing how use use aliases/images and the link between config and sensors.

mampfes:
I agree, the widget example should not be part of the source documentation because only a few people will find it. I think for now the FAQ is indeed the best place. Maybe we should think of an extra page (or at least chapter) which gathers widget examples (ideally including screen shots).

apbarratt:
That sounds reasonable to me.  I'll remove the widget and sensors part for now and I'll remove the images bitty too.
I might have just missed it, but I didn't see a mention anywhere that sensor types rely on the alias name if it has been set, so I may look to add that somewhere when I'm finding a good place to plonk the widget examples too :)

mampfes:
Thanks for your contribution.

Issue #872, opened 2023-04-12
Source Request for Kiel Germany


jkasten80:
There are ICS files build like this https://abki.de/abki-services/abki-leerungen-ical?data=2023-streetname-housenumber-stand-12-04-2023

5ila5:
Do you know if the URL always contains the current date or is it only updated on some days and contains for example on a Sunday still the date of Friday?

5ila5:
fixed by #874

jkasten80:
Did i do something wrong? 
![image](https://user-images.githubusercontent.com/85489899/231961694-fb6a8884-5f65-465c-be5a-21a15c15c774.png)

jkasten80:
I got it, it didnt download the source file. I did it manual and now it works.

5ila5:
this source is not yet part of the current release (1.38.0) it will be part of 1.39.0 or is available with the master (bzw. git) version

PR #873, opened 2023-04-12
ics: add support for additioan formatted urls other than year

5ila5:
ics source: Adds support for formatted URLs for all wildcards/formats available from datetime strftime
https://docs.python.org/3/library/datetime.html#strftime-and-strptime-format-codes

to support providers like mentioned #872

-- edited --
added ICS source description for abki.de (#872)


mampfes:
I see one problem with this approach:

There are 2 year fields in the URL, and both have a different meaning: The first one is the desired year, and the 2nd one is the current year (= now). This will make a difference in December: In December, the ICS source will (try to) fetch data for the current year and also for the next year. But with this approach, it will also set to current year field in the url to the upcoming year - which will probably fail.

Therefore, I think it is easier to create a dedicated source for AWB Kiel.

mampfes:
I prefere to close this PR if we currently don't need the extra fields. Just to keep the ICS simple.

Issue #324, opened 2022-08-06
Austria - Saubermacher (Daheim)


leochras:
Would it be possible to include this API?
I don't really know how to do it, because each type of trash has it's own id

https://daheim.saubermacher.at/api/calendar/60629/

dt215git:
There look to be 7 different id's in that response, so presumably your have to sort your waste into 7 different categories?

Doesn't the Saubermacher app gives a more helpful description of the waste category collection for any of the given dates rather than just saying '166357' is due for collection. If so, you may be able to work out what types of waste each id's corresponds to. Then it'd be possible to use a dictionary to make the HACS integration display something more readible. For example:

```
WasteTypes = {
    "166351": "Paper?",
    "166352": "Organic?",
    "166353": "Plastic?",
    "166354": "Household?",
    "166355": "Metal?",
    "166356": "Batteries?",
    "166357": "Wood?",
}
```

leochras:
Ok thanks, I'll look into that.
There's also a graphql api, but I don't know if thats better or worse..
https://prd-daheimapp-api.herokuapp.com/graphql

leochras:
Unfortunately, there's a new API which isn't public anymore and now requires authentication. Is it possible to provide authentication somehow?

Also, there's an upcoming competitor to "Daheim" in our region called "Cities" which also provides an API, which also requires authentication and also spoofing the user-agent.

https://api.citiesapps.com/garbageplans

5ila5:
I wrote a Source for CITIES (#880)

leochras:
Great work! I'll try it out and close this if it works!

leochras:
It seems to work, thank you very much!

PR #880, opened 2023-04-15
Adding support for CITEIS app

5ila5:
Adding Support for cities supported by the app [CITIES](https://citiesapps.com/) (Mostly Austrian cities) 
suggested in #324

As this are major additions I do not want to merge this request by my own, without additional eyes on the code.


mampfes:
I need more time, but will have a look at it asap.

custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
```diff
@@ -0,0 +1,875 @@
...
+    },
+    {
+        "title": "Edelsbach bei Feldbach",
+        "url": "http://www.edelsbach.at",
+        "country": "at",
+    },
+    {
+        "title": "Jagerberg",
+        "url": "http://www.jagerberg.info",
+        "country": "de",
```
mampfes, line 115: Should be `at`, or?

custom_components/waste_collection_schedule/waste_collection_schedule/source/citiesapps_com.py
```diff
@@ -0,0 +1,67 @@
...
+AUTH_TOKEN = "Basic NjQzNmI2NWFiZjBmYzkwMDEyOTdkNjU0OmQ1Y2IxMzYxNzJiODBkMmI0NTQyN2MyZDUyYTZhMTNmMDhlOGViZDE0M2FkNTVlNjI3MzZmZTY3ZjNkNw=="
+
+
+class Source:
+    def __init__(self, city: str, calendar: str):
+        self._city: str = city
+        self._calendar: str = calendar
+
+    def fetch(self):
+        ciites_apps = CitiesApps()
```
mampfes, line 54: typo :-)  ciites -> cities

custom_components/waste_collection_schedule/waste_collection_schedule/source/citiesapps_com.py
```diff
@@ -0,0 +1,67 @@
...
+    "Leichtfraktion": "mdi:recycle",
+    "Leichtverpackung": "mdi:recycle",
+    "Gelber": "mdi:recycle",
+    "Sonder-": "mdi:dump-truck",
+    "Abfallwirtschaftszentrum": "mdi:house",
+    "Strauchschnitt": "mdi:tree",
+}
+
+
+AUTH_TOKEN = "Basic NjQzNmI2NWFiZjBmYzkwMDEyOTdkNjU0OmQ1Y2IxMzYxNzJiODBkMmI0NTQyN2MyZDUyYTZhMTNmMDhlOGViZDE0M2FkNTVlNjI3MzZmZTY3ZjNkNw=="
```
mampfes, line 45: Isn't this the same token as in the service folder? If yes, I would import it here to use the definition from service.

doc/source/citiesapps_com.md
```diff
@@ -0,0 +1,233 @@
+# App CITIES
+
+Support for schedules provided by [App CITIES](https://citiesapps.com), serving App CITIES, Austria.
+
+## Supported cities
+
+|City|Website|
+|-|-|<!--Begin of service section-->
```
mampfes, line 8: I assume that the github markdown renderer will not like the comment within the table. Didn't you fix a similar problem some weeks ago?

mampfes (commented):
Thanks! This looks like huge effort. But great that we have much more cities supported in Austria now.

I'm not a big fan of the wizard's anymore, because a lot of people have problems with it. But in this case I think it is ok, because the wizard is only the fallback.

custom_components/waste_collection_schedule/waste_collection_schedule/source/citiesapps_com.py
5ila5: You're right this token is not needed here at all anymore

custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
```diff
@@ -0,0 +1,875 @@
...
+    },
+    {
+        "title": "Kaindorf",
+        "url": "https://www.kaindorf.at",
+        "country": "at",
+    },
+    {
+        "title": "Ottendorf an der Rittschein",
+        "url": "https://www.ottendorf-rittschein.steiermark.at",
+        "country": "de",
```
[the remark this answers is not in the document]
5ila5: looks like some cities set a wrong country_abbreviation variable. I will allways set "at". currently only austrian cities are supported anyway

PR #883, opened 2023-04-16
Add Chiemgau Recycling for Landkreis Rosenheim Source

cebor:
Hello,

this is my first attempt to read out the collection data for the paper garbage can from the Chiemgau/Rosenheim region.

What is not yet clear to me is how this project deals with other dependencies. Should I just put them in the requirments.txt?

BR
Felix


cebor:
And how can i let run a Test for this specific source?

5ila5:
> And how can i let run a Test for this specific source?

test_sources.py -s SOURCE

Detailed info:
[Dedicated source contribution guide](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md)

5ila5:
We generally avoided PDF parsing as it is pretty hard to get reliable data.
Your approach also needs regular updating the pdf file path and will not work for longer than one year without changing the url. 
It looks like you can also use the "Die Blaue Tonne" App. We could have a look how the App gets its data and if we could use the same API to get the data without PDF parsing.

cebor:
Hello @5ila5,

thanks for your feedback. I already tried to contact Chiemgau Recycling, but got no answer.

For now I implemented my own public api which is parsing and caching the dates for "Landkreis Rosenheim" from their site. 

The implementation has been adapted and now pulls the data from this API. I hope it is better now. 

At the moment only the "County of Rosenheim" is supported, "Rosenheim City" and "County of Traunstein" will follow.

mampfes:
Agree, an "unofficial" API in between is an additional point of failure. If there is no other solution a comment in the Readme/markdown of the source would be ok for me.

5ila5:
Thanks for your contribution

Issue #891, opened 2023-04-19
Abki recycling waste wrong data


jkasten80:
The source is working fine, but for the recycling waste it takes the data from the test street Achterwehrer Strasse 1 A and not from my street. 

Could be also Aarhusstrasse from the source.

I tested it again, if I put my street in the source where Aarhusstrasse is, than everything is fine.

5ila5:
Sorry, did not see this as this only affected recycling waste 
fixed by #893

jkasten80:
Awesome! Now everything looks fine. Thx for your fast fix!

PR #893, opened 2023-04-19
fixing abki_de recycling

5ila5:
fix for #891


mampfes:
There is another occurance of the test data in the source:
https://github.com/mampfes/hacs_waste_collection_schedule/blob/996b43dc51d18a766589b77b96c027c0ef04917a/custom_components/waste_collection_schedule/waste_collection_schedule/source/abki_de.py#L93

PR #915, opened 2023-04-23
change grafikai.svara.lt api url to use https instead of http

justasmalinauskas:
Hi,

As grafikai.svara.lt started to use https, it broke schedule collection. This PR changes url to https


5ila5:
Thanks for your contribution.

Issue #902, opened 2023-04-20
need city, street, house number support for infeo.at


PeaceTrain:
@dm82m 
I would really need the feature to use city, street, house number for infeo.at
customer would be IKB

PeaceTrain:
they seem to change the IDs for at least the calendar ID and the CityID :-(
https://services.infeo.at/aloa/api/ikb/wastecalendar/export/?calendarId=<CALENDARID>&cityId=<CITYID>&streetId=<STREETID>&housenumber=<HOUSEID>&fractionIds=173&fractionIds=174&fractionIds=175&fractionIds=176&outputType=ical

dm82m:
Will have a look next week

dm82m:
@PeaceTrain PR is open, if you want to test it before it is released you can do so:

1) replace the file on your home assistant custom_components/waste_collection_schedule/waste_collection_schedule/source/infeo_at.py with the content of that one here: https://raw.githubusercontent.com/mampfes/hacs_waste_collection_schedule/c39e2a73dce01b5abd19b150709b3dd6aa64a760/custom_components/waste_collection_schedule/waste_collection_schedule/source/infeo_at.py

2) add this to your configuration.yaml
```
waste_collection_schedule:
  sources:
    - name: infeo_at
      args:
        customer: ikb
        city: your_city
        street: your_street
        housenumber: your_housenumber
```

3) restart your server and report back if you got the data

HINT: by default we are loading all available fraction data from the service. You can filter that later on in the waste collection schedule itself. see https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/faq.md and the question 'How can I hide a waste type I don't want to see?'.

and one sidenote: yes, infeo.at is frequently changing ids. that is why I added it to hacs_waste_collection_schedule. I take care of this in the code. for you as the user there is nothing to do after adding your configuration.

PeaceTrain:
thank you very much! works like a charm

PR #916, opened 2023-04-23
#902

dm82m:
implemented support for city, street and housenumber for infeo.at and added Innsbrucker Kommunalbetriebe as new customer


5ila5:
If you are already on this you could add Salzburg as customer:
test:   "salzburg": {"customer": "salzburg", "city": "Salzburg", "street": "Adolf-Schemel-Straße", "housenumber": "13"} works

dm82m:
added it to the PR @5ila5.

Fun fact @5ila5 and @mampfes:
- Bogenschütz and Salzburg is using this endpoint: https://services.infeo.at/**awm**/api/salzburg/wastecalendar/
- Ikb is using this endpoint: https://services.infeo.at/**aloa**/api/ikb/wastecalendar/ but also works with avm

5ila5:
Thanks for your contribution.

PR #920, opened 2023-04-25
Update lebach_de.md

zippone:
"Abfuhrkalender_" was removed from the filename... searched almost an hour to fix the integration after a reboot


mampfes:
Please update also the corresponding source file in the folder `doc/ics/yaml`. Otherwise you changes will be overwritten the next time `update_docu_links.py` is executed.

Issue #932, opened 2023-04-30
Issue: zaw-online.de is not working


STmsDbg:
Hi,

the source German "zaw-online.de" is not working any more.
I think they changed the website/api.

I hope you can fix it.

coreuter:
It looks like they have switched to a service called "mymuell.de".

In the meantime, the calendar can be made functional again by using an .ics file. Navigate to https://www.zaw-online.de/abfallkalender/,  generate an .ics file via the city and street name and include it in your configuration (see https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/ics.md).

```
waste_collection_schedule:
  sources:
    - name: ics
      args:
        file: custom_data/calendar/2023_abfallkalender.ics
```
*Note*: Replace `custom_data/calendar/` with the path where you placed the .ics file.

dt215git:
zaw-online.de schedules seem to be coming from urls like this: `zaw.jumomind.com/mmapp/api.php?r=trash&city_id=89&area_id=13`
There is already a source for [jumomind](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/jumomind_de.md); Can you get your schedule by switching to that source?

coreuter:
I can confirm that using the source `jumomind_de` with the `city_id` and the `area_id` returned by the provided Python script works for me!

STmsDbg:
Yes, I can confirm that it works with the source jumomind_de. Thanks

5ila5:
ZAW Source removed.
Alternative officially [documented (jumomind_de)](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/jumomind_de.md)
by #936

PR #936, opened 2023-05-01
Updateing_jumomind_de

5ila5:
jumomind_de providers are now listed in info.md and README.md
removed ZAW as they switched to jumomind_de #932

Not sure if we should realy add all 216 MyMuell cities to README.md I allready filtered out duplicates and removed weird endings but some might allready be part of an other source (Darmstadt is allready implemented in ead_darmstadt_de )


mampfes:
Thanks! If it is obvious for the user that the cities use MyMuell, I would not add all cities. Then a single entry for MyMuell would be sufficient in my mind.

But if users might have trouble to find MyMuell, it is ok for me to add all cities.

5ila5:
I think it is often not obvious, that a city uses mymuell. It often isn't even mentioned on the bin collection page or mentioned but as müllapp and you need to klick the link to see the name (`mymuell`). I noticed that there are all cities listed, even if they are part of a Landkreis I'm currently removing them and replace them with just the name of the Landkreis.

PR #77, opened 2021-09-13
Change ICS split_at to use regex, and make casing uniform.

cpw:
The ICS split feature is a great idea, but for maximum functionality, it should use a regular expression based split string instead of a plain delimiter.

My garbage has summary lines as follows:
```SUMMARY:Garbage\, green bin\, yard trimmings\, and blue bin
SUMMARY:Yard trimmings\, green bin\, and black bin
SUMMARY:Garbage\, green bin\, yard trimmings\, and blue bin
SUMMARY:Black bin\, yard trimmings\, and green bin
SUMMARY:Garbage\, blue bin\, green bin\, and yard trimmings
SUMMARY:Black bin\, green bin\, and yard trimmings
SUMMARY:Garbage\, blue bin\, yard trimmings\, and green bin
SUMMARY:Black bin\, green bin\, and yard trimmings
SUMMARY:Garbage\, green bin\, yard trimmings\, and blue bin
SUMMARY:Green bin\, yard trimmings\, and black bin
SUMMARY:Garbage\, blue bin\, green bin\, and yard trimmings
SUMMARY:Green bin\, yard trimmings\, and black bin
SUMMARY:Garbage\, yard trimmings\, green bin\, and blue bin
SUMMARY:Black bin\, yard trimmings\, and green bin
SUMMARY:Garbage\, blue bin\, green bin\, and yard trimmings
SUMMARY:Green bin\, yard trimmings\, and black bin
SUMMARY:Green bin
```

So it really needs a bit of processing to handle it and make uniform categories. Using a regular expression for splitting:
```
        split_at: "\\, [and ]*"
```
allows me to make uniform strings out of the summary lines. using `.capitalize` allows me to turn everything into a uniform case. (lower or upper would probably work fine as well).

I end up with garbage in 5 categories:
* Garbage
* Blue bin
* Green bin
* Black bin
* Yard trimmings

This seems to be the minimal change to allow this to work. Note that this ICS is actually from recollect, so by doing this, you'd probably allow ALL recollect ICS calendars to work pretty nicely.

PR #954, opened 2023-05-06
Update ReCollect docs and tests with new split_at regex

kaechele:
This fixes the regex to allow for collections with exactly two waste types. The current regex does not split these correctly (no comma before the "and").
It seems that Ottawa always has at least 3 waste types per event, so the issue does not occur there.

I've added Simcoe County as a test case that uses two waste type events as the norm.

This also addresses a comment in #77 where a user was looking for the correct regex to use in this case.


mampfes:
Thanks for your contribution.

PR #977, opened 2023-05-14
Add cumberland council

pwae:
A copy of inner west - as it's using the same data provider.
Updated with relevant local examples.


custom_components/waste_collection_schedule/waste_collection_schedule/source/cumberland_nsw_gov_au.py
```diff
@@ -0,0 +1,122 @@
...
+TEST_CASES = {
+    "Berala Community Centre": {
+        "suburb": "Berala",
+        "street_name": "Woodburn Road",
+        "street_number": "98 to 104",
+    },
+    "McDonald's Auburn": {
+        "suburb": "Auburn",
+        "street_name": "Parramatta Road",
+        "street_number": "116",
```
5ila5, line 19: could you please provide a Test case with integer street_number (integer values should also be accepted) (and fix your code if it fails with integer value)

pwae: Confirmed it works fine with an integer value, and added test case in:

Thanks!
```
./test_sources.py -s cumberland_nsw_gov_au                                                                                add-cumberland-council ✱
Testing source cumberland_nsw_gov_au ...
  found 104 entries for Berala Community Centre
  found 104 entries for McDonald's Auburn
  found 104 entries for Chickenlicious Guildford
```

pwae:
Items resolved & branch bought up to date

5ila5:
Thanks for your contribution

PR #982, opened 2023-05-18
nuernberger_land_de: add filter to include all events

jodli:
As `mahe112` noted in [this](https://github.com/mampfes/hacs_waste_collection_schedule/pull/549#issuecomment-1552076989) comment, I prepared this PR to fix the issue of missing events in the source: `nuernberger_land_de`.


5ila5:
Thanks for your contribution

PR #989, opened 2023-05-23
Add Barnsley MBC

LZStealth:
Adding Barnsley MBC

This is collected from the BMBC website using postcode and house number.
Only provides waste information 3 weeks in advanced.


custom_components/waste_collection_schedule/waste_collection_schedule/source/barnsley_gov_uk.py
```diff
@@ -0,0 +1,87 @@
...
+import re
+
+from datetime import datetime
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Barnsley Council"
+DESCRIPTION = "Source for barnsley.gov.uk services for Barnsley MBC, UK."
+URL = "https://wwwapplications.barnsley.gov.uk/wastemvc/ViewCollection/SelectAddress"
+TEST_CASES = {
+    "Test_001": {"postcode": "S703QN", "address": "18"}
```
5ila5, line 12: `address` argument must be `houseNo`.
Please provide an additional test case containing an integer `houseNo` as source, who do not accept int values for numeric values caused some trouble in the past.

LZStealth:
Cancelled Request.
Found an issue where low level house numbers would not be selected correctly and would create an additional selection dialogue.
The above changes will also be made and resubmitted.

PR #995, opened 2023-05-24
Method does not work in Firefox/Safari in south_norfolk_and_broadland…

marsdeat:
…_gov_uk.md

The javascript returns an error on the first line for both Firefox and Safari-based browsers, but works as intended on Chromium. Added a clarification as I cannot fix the js myself.


5ila5:
Thanks for your contribution

PR #1001, opened 2023-05-25
Clarify SNBC documentation

marsdeat:
add new arguments to the template up front, and clarify the payload method's shortcomings. Clarify what SNBC is.


5ila5:
Thanks for your contribution

PR #1002, opened 2023-05-25
Update South Norfolk and Broadland in info.md

marsdeat:
Clarify South Norfolk and Broadland - without getting too deep into local politics, they're still technically independent, so people will search for either individually.


5ila5:
I think the old one is fine. The title should not be too long. 
Your changes would be overwritten by the update_docu_links.py script which will take the `TITLE` argument from the source file.

https://github.com/mampfes/hacs_waste_collection_schedule/blob/9aa5f2c2f27ed148cb19db046e4d7973ae52ffd5/custom_components/waste_collection_schedule/waste_collection_schedule/source/south_norfolk_and_broadland_gov_uk.py#L10-L14

PR #1006, opened 2023-05-26
Update south_norfolk_and_broadland_gov_uk.py

marsdeat:
Separate SNC and BDC entries for info lists, as they're still separate authorities that will be searched separately.


5ila5:
Please run the update_docu_links.py script.
You get 3 Entries now get 3 entries `South Norfolk and Broadland Council`, `South Norfolk Council` and `Broadland District Council`
The best way would be to change the TITLE argument to one of them and add the other one with the `extra_info` 

Please do not open a new pull request every time you change something. just push to your existing one.

PR #1008, opened 2023-05-28
Fix variable names (typo) and update search to support Garden Waste Collections in eastherts_gov_uk

WhimsySpoon:
**BREAKING CHANGE (for this source)**

East Herts operate a paid-for Garden Waste collection service. Only properties that have paid for this service will be offered the next collection dates on the lookup. I pay for this service, yet wasn't receiving any dates for the next collection.

In my debugging, I noticed that the council's website has been updated with the following guidance on the search:

> Please enter your property name/number and postcode ONLY. Please do NOT enter your street.

The component was previously providing the street name which was resulting in it ignoring the house number and returning all 35 properties in the post code and selecting the first returned property by default. That property does not subscribe to the garden waste service, hence my problem.

I have removed the option to supply a street name (quite why the council hasn't removed this from the search is beyond me) and also noticed a typo in the house name/number variable, so the data was never being injected in the web page.

This PR now adds the ability to return Garden Waste collections (for those properties subscribed) and fixes the typos so the actual address lookup will take effect. I have also updated the docs accordingly.

Unfortunately this does introduce a breaking change as the configuration for HA's `sensor.yaml` has changed from:

`house_name_numer:` to `house_name_number:`

Could this be included in the release notes, should the change be approved please?

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/eastherts_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/eastherts_gov_uk.py
@@ -1,4 +1,5 @@
 from datetime import datetime
+import logging
 
 import requests
 from bs4 import BeautifulSoup
@@ -25,19 +26,30 @@ ICON_MAP = {
 
 API_URL = "https://uhte-wrp.whitespacews.com/"
 
+_LOGGER = logging.getLogger(__name__)
 
 class Source:
     def __init__(
         self,
+        address_name_numer=None,
         address_name_number=None,
         address_street=None,
         street_town=None,
         address_postcode=None,
     ):
-        self._address_name_number = address_name_number
+        self._address_name_number = address_name_number if address_name_number is not None else address_name_numer
         self._address_street = address_street
         self._street_town = street_town
         self._address_postcode = address_postcode
+
+        if address_name_numer is not None:
+            _LOGGER.warning("address_name_numer is deprecated. Use address_name_number instead.")
+
+        if address_street is not None:
+            _LOGGER.warning("address_street is deprecated. Only address_name_number and address_postcode are required")
+
+        if street_town is not None:
+            _LOGGER.warning("street_town is deprecated. Only address_name_number and address_postcode are required")
 
     def fetch(self):
         session = requests.Session()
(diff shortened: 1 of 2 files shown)


5ila5:
To avoid breaking changes you could use something like this, since all parameters are optional anyway:

```python
    def __init__(
        self,
        address_name_numer=None,
        address_name_number=None,
        address_street=None,
        street_town=None,
        address_postcode=None,
    ):
        self._address_name_number = address_name_number if address_name_number is not None else address_name_numer
```

WhimsySpoon:
Thanks for your swift and helpful feedback, @5ila5. I've made the changes you've requested, and added some logging regarding the deprecation. This is my first attempt at Python, so apologies if if I've missed the mark entirely.

I'd be happy to pull the warnings and revert to the changes you've suggested if they are unsuitable.

WhimsySpoon:
I'm going to convert this to draft for the time being. I've just noticed that two of the three sensors went `unknown` a few hours ago.

I don't think it's my changes as I've re-downloaded from HACS and the problem persists, but I want to be sure. I've just tried it in the test runner (using the current release) and I can see that it's reporting the same number of entries from the council website. Is there anything further I can do on the debug side? It seems they're just not making it into the sensor in HA.

WhimsySpoon:
I'm certain this change isn't the cause of the weird behaviour (especially as I've confirmed it's happening with the current release. Looking at the HTML that's being captured, I'm seeing `</body>` and `</html>` tags half way down the page. After these tags, the rest of the page appears. It's almost as if the page is being parsed before it's fully rendered as I don't see this when consuming it in a browser.

@5ila5  Is this a known issue that's affected other sources in the past?

I'm going to re-open this PR as the changes are valid and will continue digging on the other issue.

5ila5:
Thanks for your contribution.

Your changes look fine. And the source returns values (for the test cases):
```
➜ custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s eastherts_gov_uk -lit
Testing source eastherts_gov_uk ...
  found 4 entries for Example
    2023-06-03 : Refuse Collection Service [mdi:trash-can]
    2023-06-09 : Recycling Collection Service [mdi:recycle]
    2023-06-16 : Refuse Collection Service [mdi:trash-can]
    2023-06-23 : Recycling Collection Service [mdi:recycle]
  found 4 entries for Example No Postcode Space
    2023-06-03 : Refuse Collection Service [mdi:trash-can]
    2023-06-09 : Recycling Collection Service [mdi:recycle]
    2023-06-16 : Refuse Collection Service [mdi:trash-can]
    2023-06-23 : Recycling Collection Service [mdi:recycle] 
```

If the sensors do not display any data, it's probably a missconfigured sensor (did the name of collections change and do not match the sensor `types` anymore?)

PR #1026, opened 2023-06-07
Correct council name tmbc_gov_uk.md

nick-hills:
Updated the page title which reflected another authority


mampfes:
Thanks for your contribution.

PR #1057, opened 2023-06-29
Support addresses that do not have a comma in Recycle Coach:

haimgel:
# Support addresses that do not have a comma in Recycle Coach

Example: London, Ontario, Canada uses Recycle Coach, and the existing code works with this minor modification. E.g. this is sufficient to make it work for London, Ontario:

```yaml
    waste_collection_schedule:
      sources:
        - name: "recyclecoach_com"
          args:
            street: 1065 Sunningdale Rd E
            city: London
            state: "ON"
```


5ila5:
Thanks for your contribution

your if statement is not needed we can just do:
```python
 streetpart = self._format_key(zone_res['address']).split(",")[0]
```
the second part wan't used anyway.


I added your config as test case for potential future changes.

I added London, Ontario to EXTRA_INFO to show on the README.md and INFO.md file

PR #1080, opened 2023-07-11
Change URL of bielefeld_de to new URL

janpfischer:
Bielefeld changed it's url form 
```

https://anwendungen.bielefeld.de/WasteManagementBielefeld/WasteManagementServlet
```

to 
```
https://anwendungen.bielefeld.de/WasteManagementBielefeldTest/WasteManagementServlet
```

Yes, unbelievable but true. They added "Test" to their production URL. Also the URL on their [Main site](https://service.bielefeld.de/detail/-/vr-bis-detail/dienstleistung/3329451/show) refers to that link. 
The above mentioned URL  
```
https://anwendungen.bielefeld.de/WasteManagementBielefeldTest/WasteManagementServlet
```

is not callable in a Browser directly. In the Browser it requires some parameters. 
```
https://anwendungen.bielefeld.de/WasteManagementBielefeldTest/WasteManagementServlet?SubmitAction=wasteDisposalServices
```
But the Integration is working again after that change. I've tested it. 

It may be possible that they will change that back in the near future. I will monitor it and perhaps change that back.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/bielefeld_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/bielefeld_de.py
@@ -11,13 +11,13 @@ DESCRIPTION = "Source for Stadt Bielefeld."
 URL = "https://bielefeld.de"
 TEST_CASES = {
     "Umweltbetrieb": {
-        "street": " Eckendorfer Straße",
+        "street": "Eckendorfer Straße",
         "house_number": 57,
     },
 }
-SERVLET = (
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for the notice I'd rather add a fail save if they decide to switch back so that we (or they) don't break all configurations again (without an update).

cleaned up some mess in the code as well

Thanks for your contribution this fix will be part of Version 1.41.0 or is available as git / master version

janpfischer:
Great, thank you, @5ila5! Also great idea with the fallback!

PR #1086, opened 2023-07-15
Add renosyd.dk source.

lexitus:
Adds support for renosyd.dk, plus country code for Denmark.

Seems to work reliably, but I haven't been able to track down a property signed up for green waste, so no test example for that at the moment.


5ila5:
Thanks for your contribution

PR #1100, opened 2023-07-18
Add Stratford DC as a source.

12pt:
Adds Stratford-on-Avon DC as a source.

The website is [here](https://www.stratford.gov.uk/waste-recycling/when-we-collect.cfm/part/calendar).

Stratford DC uses what they call a "123+" system:
- Food waste every **1** week
- Recycling every **2** weeks
- General Waste once every **3** weeks
- Green/Garden waste every **other** week

Confirmations:
- I've updated README.md and info.md
- I've added stratford_gov_uk.py as a source
- I've run the test program and it downloads the schedules correctly.

Screenshot from 2023-07-18 13-46-31


12pt:
I've identified a bug with time parsing, please hold

PR #1107, opened 2023-07-21
Change get parameter - verify webrequests

maverickme:


5ila5:
Thanks for your contribution it seems like they fixed their ssl

PR #1123, opened 2023-07-29
country code error

dt215git:
Minor edits to Chiemgau entry to mitigate the following error when updating readme/info docs.

```bash
Orphaned sources without country =========================
filename:/doc/source/chiemgau_recycling_lk_rosenheim.md, title:Chiemgau Recycling - Landkreis Rosenheim, url:https://chiemgau-recycling.de/, country:rosenheim
```


5ila5:
You're right missed that

PR #1147, opened 2023-08-08
Adding cardiff_gov_uk

sgsabbage:
This adds Cardiff Council as a source.

Unfortunately they don't provide an easy API to use to get this information, so the source has to pretend to be a browser to gain access to the data.

This is built off Tom Brien's (MIT licensed) [existing work with the Cardiff Waste API](https://github.com/TomBrien/cardiffwastepy), many thanks to him for figuring out how it works so that I didn't have to!

For now the source requires a UPRN, but I have defined the `__init__` method with a keyword argument for the UPRN so that the source can be updated in future with postcode/house number if desired without breaking existing setups.

Source and MD files added, info and README updated. Two tests created, one for houses with glass collections, one for those without.


5ila5:
Thanks for your contribution

PR #1158, opened 2023-08-15
Added Calgary, AB, Canada as a source

adechant:
Added Calgary as a source. Updated and alphabetized the README and INFO sections for Canada.


5ila5:
The md files are automatically updated using the `update_docu_links.py` script with the url parameter.

Thanks for your contribution

d-allin:
@adechant Unless I'm missing something, the quadrant seems to be hardcoded as "SE"
`schedule_download = requests.get(
            SCHEDULE_LOOKUP_URL,
            params={"address": self._street_address, "quadrant": "SE"},
        )`
So addresses outside that quadrant do not pull data unless the user goes in and change that value in calgary_ca.py. At least, this was an issue I ran into and fixed by changing "SE" to my quadrant.

PR #1164, opened 2023-08-22
Add support for subcouncil databases in innerwest.nsw.gov.au

fenugeek:
My AU NSW local council (**Inner West Council**) was an amalgamation of three councils several years ago (_Marrickville_, _Leichhardt_ and _Ashfield_), but the waste collection databases are still - STILL! - split by individual old-council areas.

The previous code only searched collection details for the old Marrickville area. This PR adds a search by suburb across the other two databases, to find details for everyone in Innerwest.

Searching is preferred to forcing users to specify out what old council source they're in since there isn't an official way to find out what council you used to be in anymore.


5ila5:
Thanks for your contribution

PR #1167, opened 2023-08-22
Source Fix - Exeter Gov UK

AtomBrake:
Fixing the REGEX value to remove the ordinals


5ila5:
Thanks for your contribution

PR #1184, opened 2023-08-26
Sensor event index

5ila5:
adding the ability to show the Nth collection date for a sensor wanted by #1099

I haven't spent that much time in the homeassistant custom components code (except for sources) so I might have missed stuff that also need to be changed but after a bit of testing everything seems to work.

I'm also not sure if  `event_index` is a good name for this parameter but I couldn't think of a better name.
Feel free to change the name and the documentation if you like


mampfes (approved):
Looks good. I also don't have a better name for the configuration option.

Could you please add a description in the "Attributes for sources" table in installation.md.

mampfes (approved):

Issue #1214, opened 2023-09-01
Request for Landkreis Osnabrück


seppeel:
Hey,

can you please add Landkreis (not City) Osnabrück to the supported list?

The source is https://www.awigo.de/

Filling out the form on this page generates a link to an ics file like
https://www.awigo.de/fileadmin/kalender/AWIGO_Kalender_Rest_Papier_Gelb_Bio_Schadstoffmobil_577242001.ics

This link is only valid for a few days and then has to be generated again. I hope it's possible to realize with this addon.
Thanks for your work.

Tobias

PR #1216, opened 2023-09-01
add ICS source awigo_de

5ila5:
requested by #1214


5ila5:
did not see

> This link is only valid for a few days and then has to be generated again. I hope it's possible to realize with this addon.
> Thanks for your work.

PR #1240, opened 2023-09-06
Fix typos Ravensburg

twiddern:
Thanks for adding the new data source, a typo will be fixed with this request.

Lan*d*kreis


5ila5:
Thanks for your contribution

PR #1251, opened 2023-09-08
Adding french city mamirolle source

AghilasMessara:
Adding French city Mamirolle

```bash
Testing source mamirolle_info ...
  found 2 entries for TestSource
    2023-08-21 : Poubelle grise [mdi:trash-can]
    2023-08-14 : Poubelle jaune [mdi:recycle]
```


5ila5:
I removed your additional dependency (would need to installed on homeassistant and probably also be added to the manifest file). Added a check weather the date is actually after (or at) today (previous implementation would probably show next collection at the beginning this year January instead of next year January in December)

PR #1260, opened 2023-09-12
Haveafald new triggers also with "240 L" description

LasseLegarth:
The description on renosyd is as following:
1 x Restaffald 240L 
1 x Papir/Pap 240L 
1 x Emballage 240L 
1 x Haveaffald 240 L 
1 x Storskrald 

The space between 240 and L in Haveaffald didn't trigger before, fixed it with \s?


5ila5:
Thanks for your contribution

PR #1274, opened 2023-09-21
Updated Maroondah VIC AU to use OpenCities API

DanAE111:
The local council have updated their website so it is now using OpenCities like many other local councils in this jurisdiction. I've updated the source to reflect that and this is working well with the test_source.py script and in my instance of Home Assistant.


5ila5:
Thanks for your contribution

PR #1280, opened 2023-09-23
New Source: RM of Morris, Manitoba Canada - Recollect

JamesGiesbrecht:
Adding RM of Morris, Manitoba, Canada recollect ICS to project.

Documentation was updated via `./update_docu_links.py` .

Tests against the new source ran successfully:

```bash
$ ./custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -y recollect
Testing ICS recollect
  found 46 entries for Cambridge, MA, USA
  found 46 entries for Ottawa, ON, Canada
  found 21 entries for Georgetown, TX, USA
  found 25 entries for Sherwood Park, AB, Canada
  found 29 entries for Morris, MB, Canada
```


5ila5:
Thanks for your contribution

added your example to the test cases of the yaml (would get overwritten by the update_docu_links script)

PR #1302, opened 2023-10-05
Add GAUM Mistelbach (AT)

heindlalexander:
it's more or less a copy of "GVA Baden" as the same source of data is used.


heindlalexander:
I see there is already #1295 requesting a solution for all "umweltverbaende_at".
so it may be closed...

5ila5:
Thanks for your contribution but we will move to one source for all umweltverbaende_at municipalities added by # #1295  (like already mentioned by @heindlalexander)

PR #1311, opened 2023-10-09
Update southglos_gov_uk.py

### LINKED ISSUE #1305: South Glos UK doesn't return all Recycling / Food Bin dates
The South Gloucester UK source is only returning the Food Bin and Recycling bins every other week, when they should be every week.

Calendar in HA shows like this for me:
image

The South Glos webpage shows:
image

And the service endpoint that you are querying replies with:
image

I could put a PR in to treat the Rx dates as Recycling and Food as well?

AlexCPU:
Recycling is every week.  I've checked the endpoint for a number of address across South Glos (including the provided test cases, and this change would fix the issue on all of them.

It looks like the underlying endpoint is using:

- Cx = Recycling + Food
- Rx = Bin, Recycling + Food
- Gx = Garden

Fixes #1305


5ila5:
Thanks for your contribution

PR #1321, opened 2023-10-14
New source: City of Greater Geelong rubbish collection

dtbell91:


5ila5:
Does this work reliable? Testing this I get error 403. Probably some cloudflare stuff I see when visiting the site. But this may be due to me being located in Germany not Australia

dtbell91:
I've only had it running for a day but it updated happily overnight and again when I called the fetch service.

403 doesn't match the responses I was getting in development when it was broken (I'd get a 500 error if the payload values were wrong), so I'd agree it's probably just geoblocking.

5ila5:
Thanks for your contribution

Issue #912, opened 2023-04-21
Whittlesea Council dates in HA appear 1 day early


mikz-t:
Hi, using the whittlesea council source, it appears that all collection dates appear 1 day earlier than the actual collection date

HA

![image](https://user-images.githubusercontent.com/88444793/233633154-7fce9298-cea5-4be9-acce-ab8ec8c2c788.png)

Council Site (Source)

![image](https://user-images.githubusercontent.com/88444793/233633338-71ba20ac-ffff-4264-a95e-8b1a5b1a819b.png)

confirmed with my actual address along with test case address, dates are 1 day earlier for all waste collection types

5ila5:
That's interesting. I get for the same address, the correct dates reported by the integration.
![grafik](https://user-images.githubusercontent.com/38183212/233793537-8c2ae374-992d-4721-9e8b-e414b476883a.png)
It could be due to some kind of timezone missmatch (Tested with timezone Germany/Berlin UTC/GMT +2)

mikz-t:
@5ila5 any chance you could change your timezone to GMT +10 (Melbourne) to test? i've checked my HA instance and timezone is set correctly. I'm assuming the council's website is displaying dates based on the browsers timezone.

also sorry for the noob question, but could you guide me on how i could run the python script manually, interested to see if the polling to the council's site is returning the dates in UTC or AEST (GMT +10)

5ila5:
The source file is: [custom_components/waste_collection_schedule/waste_collection_schedule/source/whittlesea_vic_gov_au.py](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/source/whittlesea_vic_gov_au.py)

test source ([documentation](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md#test-the-new-source-file)):
```bash
custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s whittlesea_vic_gov_au -lt
```
Will run the script and return all processed entries.

I tried to change my IP address to an Australian one (using TOR) and to spoof my timezone but I still got the same output for the webpage and the source.

Are you sure, that your dates are the correct ones? Was this a one time display error?

mikz-t:
ok, I've run the script from the CLI, and all dates still appear 1 day earlier for the test addresses, plus I even modified one of the test addressees to my actual on and its out by a day as well.

[...]

PR #1328, opened 2023-10-16
Update whittlesea_vic_gov_au.py - DST Fix

mikz-t:
Referring to https://github.com/mampfes/hacs_waste_collection_schedule/issues/912

previous fix was to apply 10 hours, however due to DST time zone is now GMT+11 and issue has reappeared.

fix is to offset by 1 day as opposed to hours. I did test 11 hours but this issue would reappear in April (when DST ends)


5ila5:
thanks for your contribution

PR #1330, opened 2023-10-16
Reorder innerwest_nsw_gov_au APIs to work around duplicate data

thecowan:
Reorder Inner West Council possible APIs to attempt old Leichhardt Council before attempting old Marrickville Council.

A single suburb is duplicated between the 3 lists: "Leichhardt" appears in both Leichhardt and Marrickville API endpoints. Only a single street in Leichhardt is listed by the Marrickville endpoint, and that with only a single property; this is likely bad data.

This is not a lasting fix, as it doesn't protect against future duplicates, but there are thousands of properties in "true" Leichhardt, and only one in "false" Leichhardt, so at least this reordering fixes for most people.


5ila5:
Thanks for your contribution

PR #1345, opened 2023-10-21
fix for calgary requiring separate quadrant declaration

d-allin:
Removal of quadrant from the request made in pull https://github.com/mampfes/hacs_waste_collection_schedule/pull/1209 results in errors and no data being populated. Possibly this is an upstream issue somewhere else in the code rather than the city's json, but restoring the quadrant in the street_address arg resolves the issue and that's what I went with.

These changes assume that the quadrant will always be the last "word" (separated by spaces) in the street_address string. This is always the case with Calgary addresses, to my knowledge. There are no changes required to be made by users in their configuration's street_address arg.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/calgary_ca.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/calgary_ca.py
@@ -8,7 +8,7 @@ TITLE = "Calgary (AB)"
 DESCRIPTION = "Source for Calgary waste collection"
 URL = "https://www.calgary.ca"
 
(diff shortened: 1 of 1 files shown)


5ila5:
could you provide some test case(s) where providing no quadrant results in some Error. This prevents the removal or false implementation of potential future updates or fixes

d-allin:
While I was having errors that were seemingly resolved by the changes in my pull request, I've since been unable to re-produce them after reverting the changes in my installation. I did not take care to log the errors at the time.
Classic misdiagnosis. Looks like this wasn't needed.

PR #1356, opened 2023-10-24
Update wealden_gov_uk.py

ryanbdclark:
Changed type name for garden collection as previous type name with spaces has seemingly stopped working


5ila5:
What do you mean with stopped working? All this change does is change the name of garden waste used in homeassistant. This change does not alter the way this source parses collections at all.

I understand that the name is probably not ideal and changing it to just Garden might be appreciated but I do not get what you mean by your message.

5ila5:
@ryanbdclark Is it just the name change or do you experienced any other problems (as the name change does not impact the fetching process at all) (You can use the customize options to change the name in your configuration.yaml)

PR #1360, opened 2023-10-26
pgh.st now requires https

gazpachoking:
pgh.st was down for a while, and just came back up. Seems that it only responds to requests on https now, not http.


5ila5:
Thanks for your contribution

PR #1362, opened 2023-10-26
Fixing problem where organic isn't shown  properly

### LINKED ISSUE #1361: Weekly organic collection not shown
As picked up in #1282 (and addressed in #1284) there was a change in the data used by the Inner West Council, as a move to organic recycling not waste.

The problem is that the refactor of #1284 didn't take into consideration that the weekly collection (now `organic` not `rubbish`) isn't listed out on each date but only once in the dataset, meaning that the loop will only put it in once and then it stops returning it, causing it to not be available.

aaronpowell:
Fixes #1361

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/innerwest_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/innerwest_nsw_gov_au.py
@@ -38,6 +38,7 @@ ICON_MAP = {
     "waste": "mdi:trash-can",
     "recycle": "mdi:recycle",
     "organic": "mdi:leaf",
+    "special": "mdi:star",
 }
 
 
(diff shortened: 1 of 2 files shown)


5ila5:
Did some digging myself and found: 

- first two test cases seem to be no longer work and addresses are not listed on their website (change street number (810 -> 813, 599 -> 597)
- this approach of seems to only work on address where garbage and recycling collections are not at the same day (like 597 Darling Street ROZELLE) and ignores organic collections on addresses where garbage and recycling are on the same day (like 29 Lackey Street SUMMER HILL)

Maybe you can check if there is a better approach to this. I noticed that we get the collection dates differently than the web page. maybe we should use the API the website uses to populate the calendar.

aaronpowell:
Thanks @5ila5 I'll look into those problems.

Didn't realise how much of a variance there is in our local area on the waste pickups...

bbr111:
@aaronpowell

is this still needed?
if yes, please consider 5ila5 s comment
if no, please close.

aaronpowell:
Nope - been running the latest for a while now and totally forgot this PR was opened.

PR #1365, opened 2023-10-26
Fix for Bristol nextCollectionDate timestamp

johnbutton1200:
Fix for nextCollectionDate timestamp from API containing `T01:00:00` and breaking the integration


5ila5:
Thanks four your contribution

M21B8:
This fix doesn’t work, there’s a typo in it.

the error relates to T01:00:00

the change has T10:00:00

johnbutton1200:
This fixed the issue for me but perhaps different hour values are returned
for different areas. Maybe a better fix would be for this to not be reliant
on specific hour values being returned to parse the date?

M21B8:
It probably should, but I don’t get how it fixed it for you, do you not see the typo? The fix doesn’t line up with the error message.

johnbutton1200:
Yes, sorry you are entirely correct. Please submit an updated fix if you
can. I'm away with no access to a computer for a while.

M21B8:
https://github.com/mampfes/hacs_waste_collection_schedule/pull/1390

PR #1366, opened 2023-10-26
Fix mamirolle info args

### LINKED ISSUE #1351: Can't configure source without arguments
Hello Team,
I'm trying to configure into HomeAssistant the source that I added Mamirolle info, but I have argument issues.

`args` is marked `required`, so if none is passed, the configuration is invalid.

```
Invalid config for [waste_collection_schedule]: required key not provided @ data['waste_collection_schedule']['sources'][0]['args']. Got None. (See /config/configuration.yaml, line 27). 
```

If a dummy argument is passed. The configuration is valid but the source setup fails.
```
Error during setup of component waste_collection_schedule

  File "/usr/src/homeassistant/homeassistant/setup.py", line 288, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 109, in async_setup
    api.add_source_shell(
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 202, in add_source_shell
    SourceShell.create(
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 196, in create
    source = source_module.Source(**source_args)  # type: ignore
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: Source() takes no arguments
```
I understand that we want the configuration to fail early but the real error will still be seen only when the source is actually instantiated. Because of that I think the arguments shouldn't be required.

What do you think about this?

AghilasMessara:
Add an ignored argument to keep the arguments as required in HA source configuration
Fixes #1351

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/mamirolle_info.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/mamirolle_info.py
@@ -9,7 +9,12 @@ DESCRIPTION = "Source script for mamirolle.info"
 COUNTRY = "fr"
 URL = "http://mamirolle.info/"
 
-TEST_CASES = {"TestSource": {}}
+TEST_CASES = {
+    "TestSource": {},
+    "IgnoredArgument": {
+        "_": ""
+    }
+}
 
 ICON_MAP = {
     "Poubelle grise": "mdi:trash-can",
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

PR #1369, opened 2023-10-29
add issue forms

5ila5:
adding these issue forms could help to get better Bug reports and source requests.

@dt215git and @mampfes please provide feedback if this is a good idea/implementation.

issue forms documentation: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms

>  Note: Issue forms are currently in beta and subject to change.

try these templates at https://github.com/5ila5/template_test (feel free to add create issues)

This disables the possibiltiy to create "normal" issues so please provide feedback about this if we should use this or if we should change something

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -1,0 +1,82 @@
+name: Bug Report
+description: Report a bug with a source, sensor or the integration in general
+title: "[Bug]: "
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this Bug Report!
+  - type: dropdown
+    id: error
+    attributes:
+      label: "I Have A Problem With:"
+      options:
+        - A specific source
+        - Sensor configuration
+        - The integration in general
+      multiple: true
+    validations:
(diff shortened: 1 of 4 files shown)


mampfes:
Yes, I think this is a very good idea. I wasn't aware of this github feature.

I like especially the checklist for common errors and the last checkmark in the bug-report template.

Thanks!

dt215git:
Looks good to me and I find them helpful when reporting issues in other repos.

For the Bug issues, would it be worth adding a checklist item that says something like "Have you checked your HA instance and confirmed the source is present"?

A lot of the issues we get are because the git/master version runs ahead of HACS, and prompting them to update the component before submitting the issue may reduce those from happening.

5ila5:
I added a link to the discussion tab and a feature request form 
> I like especially the checklist for common errors and the last checkmark in the bug-report template.

liked it as well so I "_stole_" it from the [Qtile](https://github.com/qtile/qtile) repository

> A lot of the issues we get are because the git/master version runs ahead of HACS, and prompting them to update the component before submitting the issue may reduce those from happening.

added a checkbox for this (I think this includes the second point)

PR #1374, opened 2023-11-03
Update documentation of c_trace_de.md

kbabioch:
This fixes an inconsistency in the documentation, as `gemeinde` wasn't documented due to a copy & paste error.

Also it extends the documentation of `ortsteil`.


5ila5:
Thanks for your Contribution

PR #1375, opened 2023-11-03
Fix "Date not available" error for stockton_gov_uk

deanlongstaff:
Fixed an issue where the date returned from stockton_gov_uk would be "Date not available" during winter months for garden waste. This would cause an error when converting it to a date string.


5ila5:
Thanks for your Contribution

PR #1378, opened 2023-11-04
Fix requirements.txt

joshuayuen99:
Fix syntax of requirements.txt to allow for `pip install -r requirements.txt` to properly work again.


5ila5:
Thanks,
Sorry for the error

PR #1382, opened 2023-11-05
New Source: Blacktown City Council

rayoz12:
Adds a source provider for Blacktown City Council (NSW) from Australia

- City: Blacktown Council LGA
- Country: Australia
- Provider: [Blacktown City Council](https://www.blacktown.nsw.gov.au/Services/Waste-services-and-collection/Waste-collection-days)

Changes:

- [X] Python Source File
- [X] Source Markdown
- [X] README
- [X] Info.md

# Other notes
This source is extremely similar to Campbelltown Council, with the API's being exactly the same but the domain (the council) being different. I just copied and pasted that .py file and docs with minimal modification.

Also running `update_docu_links.py` seems to change line endings in the files on my windows machine (maybe I'm in the wrong here for that lol), so I've updated the docs manually with what the output should have been.

# Test Run
```
(env) PS C:\Users\Ryan\Desktop\MyCoding\python\hacs_waste_collection_schedule\custom_components\waste_collection_schedule\waste_collection_schedule\test> python .\test_sources.py -s blacktown_nsw_gov_au -l -i
Testing source blacktown_nsw_gov_au ...
  found 2 entries for Plumpton Marketplace
    2023-11-06 : General Waste [trash-can]
    2023-11-06 : Recycling [mdi:recycle]
  found 2 entries for Rooty Hill Tennis & Squash Centre
    2023-11-06 : General Waste [trash-can]
    2023-11-06 : Recycling [mdi:recycle]
  found 2 entries for Workers Blacktown
    2023-11-08 : General Waste [trash-can]
    2023-11-15 : Recycling [mdi:recycle]
  found 2 entries for Hythe St
    2023-11-06 : General Waste [trash-can]
    2023-11-06 : Recycling [mdi:recycle]
```


5ila5:
Thanks for your contribution.

This source my cause some problems as the requested site is protected by incapsular which can make it hard to get the data with a web request. For me this source does not work but I have a German IP so this may be the cause of the "Blockage"

I changed your behaviour on locationId == 0 to raise an exception instead of returning no elements. This way you get a log entry instead of just no elements.

rayoz12:
Thanks, I've observed the issue with incapsular when a user agent header isn't passed, which is why I've added it to the request in the function.

PR #1389, opened 2023-11-06
Update redbridge_gov_uk.py

chrisns:
redbridge defines garden collection as 'spring' when its not a calendar day and code doesn't catch that currently


5ila5:
Thanks for your contribution

PR #1390, opened 2023-11-06
Update bristol_gov_uk.py

M21B8:
The fix appears to have a typo


5ila5:
Thanks for your contribution.

I changed this fix so that all timestamps (starts with T) will be removed

PR #1397, opened 2023-11-09
New source: BIR

odin-h:
Source for https://bir.no/

### DIFF
--- a/README.md
+++ b/README.md
@@ -878,6 +878,7 @@ Waste collection schedules in the following formats and countries are supported.
 <details>
 <summary>Norway</summary>
 
+- [BIR (Bergensområdets Interkommunale Renovasjonsselskap)](/doc/source/bir_no.md) / bir.no
 - [IRiS](/doc/source/iris_salten_no.md) / iris-salten.no
 - [Min Renovasjon](/doc/source/minrenovasjon_no.md) / norkart.no
 - [Movar IKS](/doc/source/movar_no.md) / movar.no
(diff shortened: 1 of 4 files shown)


5ila5:
Thanks for your contribution.

Did some reformatting ran the update_docu_links.py script, added test cases, made house letter optional, and added some comments to the md file.

odin-h:
I see you've used spaces between the house number and letter in the test case keys. Not sure if that would cause the test to fail since the BIR does not return any results if you separate them when making a regular search on the website. I also noticed I've forgotten to change the key name for the test case I added when I made the PR. It should not be "Villa Paradiso". That's a remnant from me referencing [oslokommune_no.py](https://github.com/ccsalvesen/hacs_waste_collection_schedule/commit/a5639c405e3df33584288e6b7bc594bf302926f8) when I implemented this source.

5ila5:
The test case key is just the name of the test listed above the results when calling the test. Only the arguments relevant and there are no spaces (noticed naming of test cases is slightly wrong but not a major problem as the real address is easily recognisable by the actual test case) 

Test are all running fine:

```bash
➜ ./custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s bir_no     
Testing source bir_no ...
  found 20 entries for Villa Paradiso
  found 30 entries for Mardalsrenen 12 B
  found 22 entries for Alf Bondes Veg 13 B
```

PR #1406, opened 2023-11-11
Solution to Issue 1401

HolliDolli:
Data from Init is destroyed. Subsequent calls failed.


5ila5:
Looks good. The line you removed does not make sense there.

Thanks for your contribution

Issue #1430, opened 2023-11-18
[Bug]: Republic Service incorrect date when delay due to holidays


carefulcomputer:
### I Have A Problem With:

A specific source

### What's Your Problem

I just installed this integration from HACS. I configured republic services but the date on calendar doesn't seem correct.
Here is what HAAS calendar shows. Check the pickup after 24th Nov. Republic service shows it to be on 25th but integration shows it on 26th.
<img width="1240" alt="image" src="https://github.com/mampfes/hacs_waste_collection_schedule/assets/3331518/61c30fd2-a1a4-4bd6-a98a-5acbc9e0fdbc">

But here is republic service calendar
<img width="1042" alt="image" src="https://github.com/mampfes/hacs_waste_collection_schedule/assets/3331518/b94763f4-dc7c-4571-a6aa-f3775ff63179">

Also a lesser important issue: method field did not work for me, it threw an error.  i see that it was resolved but seems it has not made it to 'release' to new version yet.

### Source (if relevant)

.republic_services

### Relevant Configuration

```YAML
My config-


waste_collection_schedule:
  sources:
    - name: republicservices_com
      args:
        street_address: "33865 Cassio Cir, Fremont, CA 94555" 

(random address from same pickup zone)
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

carefulcomputer:
awesome to see that there is a fix ready to be merged. Any idea what is the timeline for merge ?

PR #1431, opened 2023-11-19
fix republicservices holliday collection

5ila5:
fix for #1430

Not sure why it was implemented the other way in the first place but the old holiday check implementation seems to behave wrongly. 

@dt215git did I miss something? I think there must be a reason you checked for a holliday in the past 5 days

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/republicservices_com.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/republicservices_com.py
@@ -28,6 +28,9 @@ TEST_CASES = {
     "No Method arg": {
         "street_address": "8957 Park Meadows Dr, Elk Grove, CA 95624",
     },
+    "33865 Cassio Cir, Fremont, CA 94555": {
+        "street_address": "33865 Cassio Cir, Fremont, CA 94555",
+    },
 }
 DELAYS = {
     " one ": 1,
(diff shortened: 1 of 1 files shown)


dt215git:
@5ila5
If I remember correctly, looking into the past 5 days was intentional.
The query initially returns a collection date that doesn't take into consideration delays due to public holidays.
It then looks for any public holiday dates and tries to work out if any of the original collection dates would be impacted.
If the original collection was due on a Thursday, you need to check whether any of  the preceeding days in that week were public holidays. If there were, then the actual collection would move by 1 day per public holiday.
I think the 5 day limit was because I couldn't find any examples on the website where collections were shifted into different weeks, so 5 days would cover the number of weekdays.
To be honest, it's not obvious what rules Republic Service apply (it could be different for different States), so there may be better ways of trying to deal with this.

5ila5:
What about keeping the old implementation but check if a collection date is already moved and then do not apply the second holiday delay? 
Is there a reason you used the holiday incorporated check? it looks like you Cycle thorough all holidays and then set incorporated to True (or not) so the check will never return false. or do i miss something?

dt215git:
When looking at it previously, I came across a case where the original collection date was delayed by a public holiday, but the shifted collection date then fell on a second public holiday in that same week. That new date was then also delayed for a second time so the actual collection date was 2 days after the original date.

I was keeping track of the public holidays that had already impacted schedules so that when original/revised dates were checked against the list of public holidays, they wouldn't be affected by one that had already been accounted for.

carefulcomputer:
Do we know when this will be merged ?

5ila5:
I do not think this is a good fix (as this probably adds new issues). There needs to be some deeper investigation what their website java script actually does and it does not seems to be that easy (as it is some "compiled" Typescript and it does not look like they provide the original source)

PR #1447, opened 2023-11-24
Update swindon_gov_uk.py

joshwillcock:
SBC have introduced a new split week system. 
The old form returns collections up to this weekend. Come Monday the data to scrape will be on the new form address. (but works as of today).

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/swindon_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/swindon_gov_uk.py
@@ -28,7 +28,7 @@ class Source:
     def fetch(self):
         params = {"uprnSubmit": "Yes", "addressList": self._uprn}
         r = requests.post(
-            "https://www.swindon.gov.uk/info/20122/rubbish_and_recycling_collection_days",
+            "https://www.swindon.gov.uk/info/20193/look_up_your_collections_from_27_november_2023",
             params=params,
         )
         r.raise_for_status()


joshwillcock:
Thankyou again @mampfes for the constant work this plugin takes.

5ila5:
This fix does not look like it will last very long. What about getting the link(s) from the <https://www.swindon.gov.uk/info/20122/rubbish_and_recycling_collection_days> page?

joshwillcock:
> This fix does not look like it will last very long. What about getting the link(s) from the https://www.swindon.gov.uk/info/20122/rubbish_and_recycling_collection_days page?

Currently it doesn't work with that URL; that was the URL I originally provided when adding the source but it is no longer working. 

Even if another step was left to check which of the two links on that page were suitable for that period of time, there is no guarantee that the code would not break when SBC remove the option (SBC IT teams charge for 'development hours' so there is no guarantee they'll update it in any specific timeframe).

From my understanding of SBC and the current method, it can be left broken until SBC make an additional change, or we can fix it immediately and then fix it in the future if it's needed.

I would propose having it work today for an uncertain amount of time is a better experience than leaving it broken for now until we assume SBC won't amend it further based on the path on the URL?

5ila5:
I thought about something like this https://gist.github.com/5ila5/404680b69e205241f4386388ff59a763 Updating a source every few month is kind of janky and not very much compatible with our release schedule (~ 1th per month)

joshwillcock:
Your gist looks like a good solution providing they keep the additional button when they go back down to one form. (Until recently the form itself was located on: https://www.swindon.gov.uk/info/20122/rubbish_and_recycling_collection_days).

Your solution still resolves the issue for now and means if they keep the additional a href to the new form it will work indefinitely, it's only if they go back to the old method it might have issues?

5ila5:
The old version should still work line 36 - 39:

```python
        try:
            return self.parse_data(soup)
        except Exception as e:
            exc = e
```

joshwillcock:
Closing as SBC have today decided to go back to the old method. - @5ila5 your new code might future proof the offering but isn't required for immediate access.

Issue #1424, opened 2023-11-16
[Bug]: Cities Apps: HTTPError: 403 Client Error: Forbidden for url: https://api.citiesapps.com/auth/guest 


thefox69:
### I Have A Problem With:

A specific source

### What's Your Problem

Since one day  an error with cities app happens.
Was running fine for some month.
Maybe Cities app changed again the authentication

### Source (if relevant)

Cities Apps Stockerau

### Logs

```Shell
fetch failed for source App CITIES: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", 
line 134, in fetch entries = self._source.fetch() ^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/citiesapps_com.py", 
line 53, in fetch cities_apps = CitiesApps() ^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py", 
line 1094, in __init__ r.raise_for_status() File "/usr/local/lib/python3.11/site-packages/requests/models.py", 
line 1021, in raise_for_status raise HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://api.citiesapps.com/auth/guest
```

### Relevant Configuration

```YAML
#waste_collection_schedule:
#    sources:
#    - name: citiesapps_com
#      args:
#        city: Stockerau
#        calendar: Abfallkalender - Abfuhrzone D
waste_collection_schedule:
    sources:
    - name: citiesapps_com
      args:
        city: Rudersdorf
        calendar: Rudersdorf 3
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

[...]

PR #1448, opened 2023-11-24
fix citiesapps

5ila5:
fix for #1424

citiesapps now requeires authentication. They moved a lot to a v2 API. Waste collections are currently still controlled by the v1 API. Which means this implementation needs an overhaule when they move this to the V2 API as well.


AlpenFlizzer (approved):
I installed your branch and used new email and password keys in configuration.yaml to set citiesapp authentication information. After HA restart, Waste Collection Schedules were immediately displayed again.

PR #1451, opened 2023-11-25
Fixed duplicate collections for southampton_gov_uk.py

paulschmeida:
Southampton City Council source was using a regex to parse the whole page, which created duplicate entries, due to the page having both list view and calendar view tabs showing the same events. I have prefiltered page text with a simple regex that only matches one of the tabs and removes all duplicates.


5ila5:
Thanks for your contribution

Issue #1409, opened 2023-11-11
[Bug]: sources: abfallnavi_de service: nds get only Data for 2024 not 2023


totow93:
### I Have A Problem With:

A specific source

### What's Your Problem

I haven't received any data for this year for ~1 weeks, only for 2024.

I think after the data for 2024 has been released, the integration will only fetch data from 2024 and not longer from 2023.

Data for 2023 is still available: https://abfallnavi.de/nds/
![image](https://github.com/mampfes/hacs_waste_collection_schedule/assets/100964266/3dc83b5b-36ab-45ec-b51e-2b9527fdbb72)

### Source (if relevant)

abfallnavi_de

### Logs

```Shell
didn't find somsthing waste_collection_schedule relatet
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: abfallnavi_de
      args:
        service: nds
        ort: !secret ort
        strasse: !secret strasse
      customize:
        - type: "Gelbe Wertstoffbehälter -2 wö.-"
          alias: Gelbetonne
        - type: "Restabfall -2 wö.-"
          alias: Restabfall 
        - type: "Restabfall -4 wö.-"
          show: false
        - type: "Bioabfall -2 wö.-"
          alias: Bioabfall 
        - type: "Altpapier -4 wö.-"
          alias: Altpapier 
      calendar_title: "Abfallkalender"
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
I cannot reproduce this issue 

this config returns dates for 2023 and 2024

```yaml
# waste_collection_schedule source configuration
waste_collection_schedule:
  sources:
    - name: abfallnavi_de
      args:
        service: nds
        ort: Norderstedt
        strasse: Adenauerplatz
```

[...]

PR #1459, opened 2023-12-01
regioentsorgugn_de get entries for all years + reformatting

5ila5:
fix for @altacoon problem mentioned in #1409


altacoon:
Super fast fix! Thank you!

PR #1489, opened 2023-12-06
implement Blackpool Council source

incognitojam:
https://www.blackpool.gov.uk/Residents/Waste-and-recycling/Bin-collections/Bin-collections.aspx


5ila5:
thanks for your contribution

PR #1502, opened 2023-12-09
Update offenbach_de.py

deese:
Fixed a typing issue when a new year comes.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/offenbach_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/offenbach_de.py
@@ -32,7 +32,7 @@ class Source:
 
         entries = self.fetch_year(now.year)
         if now.month == 12:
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/offenbach_de.py
```diff
@@ -32,7 +32,7 @@ def fetch(self):
 
         entries = self.fetch_year(now.year)
         if now.month == 12:
-            entries.append(self.fetch_year(now.year + 1))
+            entries = self.fetch_year(now.year + 1)
```
5ila5, line 35: Needs to be += or collections in deceber will not be shown in deceber.

deese: Totally right :D Thanks!

5ila5:
Thanks for your contribution

PR #1507, opened 2023-12-10
Fix year wraparound bug + month misspelt in Renosyd.dk

lexitus:
Now next year's schedule appears on Renosyd's website the year wraparound bug was revealed. Also I can't spell 🙂


5ila5:
Thanks for your contribution

PR #1509, opened 2023-12-11
fix geelongaustralia

dtbell91:
Fixes geelongaustralia_com_au source, now dynamically grabs the `VIEWSTATE` and `EVENTVALIDATION` values used by the backend.


5ila5:
Thanks for your contribution. Did some minor reformatting (like replace `type(...) != ...` with `isinstance(..., ...)`)
Unfortunately I cannot test your source as I get a Cloudflare captcha (probably because I have a german IP)

PR #1515, opened 2023-12-12
correct dependency on `beautifulsoup4` instead of `bs4`

JamieMagee:
[`bs4`][1] is a placeholder for [`beautifulsoup4`][2] to prevent typo-squatting. From the description of the `bs4` package on PyPI:

> This is a dummy package managed by the developer of Beautiful Soup to prevent name squatting. The official name of PyPI’s Beautiful Soup Python package is [beautifulsoup4](https://pypi.python.org/pypi/beautifulsoup4). This package ensures that if you type pip install bs4 by mistake you will end up with Beautiful Soup.

While the placeholder does prevent a malicious actor from hijacking the package, it's not a good idea to rely on this.

This **does not** affect the imports like:

```python
from bs4 import BeautifulSoup
```

[1]: https://pypi.org/project/bs4/
[2]: https://pypi.org/project/beautifulsoup4/

### DIFF
--- a/custom_components/waste_collection_schedule/manifest.json
+++ b/custom_components/waste_collection_schedule/manifest.json
@@ -6,6 +6,6 @@
   "documentation": "https://github.com/mampfes/hacs_waste_collection_schedule#readme",
   "integration_type": "hub",
   "iot_class": "cloud_polling",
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

PR #1524, opened 2023-12-15
Update zva_wmk_de.py

Demel75:
Once again the URL was changed. For next year is no year required in url.


5ila5:
Thanks for your contribution. I did some changes as this is a more robust approach to the problem

Issue #1537, opened 2023-12-19
[Bug]: Exception: Region ... not found (Landkreis Waldshut, AbfallPlus)


halli77:
### I Have A Problem With:

A specific source

### What's Your Problem

I use this integration for the first time and can't get it working. To create the config i used the wizard. Calendar is created, but empty. Logs shows some exceptions. I also tested with other communities or other streets in my district "Waldshut", with same effect. And I tried de.k4systems.unterallgaeu instead de.abfallwecker as app_id, because "Waldshut" is listet there as well. Same result.
Since I'm running out of ideas how to get it running, it might be a bug?
On my service providers site https://www.abfall-landkreis-waldshut.de/ there is an info, that a new city/street directory is implemented, maybe that could have an impact here?
<img width="1239" alt="Bildschirmfoto 2023-12-19 um 10 51 15" src="https://github.com/mampfes/hacs_waste_collection_schedule/assets/12441798/926848f0-e5d3-4ee6-8b51-bedd1d5015bc">

### Source (if relevant)

app_abfallplus_de

### Logs

```Shell
Logger: waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:136
Integration: waste_collection_schedule (documentation)
First occurred: 10:20:24 (2 occurrences)
Last logged: 10:20:48
fetch failed for source Apps by Abfall+: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 134, in fetch entries = self._source.fetch() ^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/app_abfallplus_de.py", line 124, in fetch for d in self._app.generate_calendar(): ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/service/AppAbfallplusDe.py", line 750, in generate_calendar self.select_kommune() File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/service/AppAbfallplusDe.py", line 490, in select_kommune raise Exception(f"Region {self._region_search} not found.") Exception: Region Lauchringen not found.
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: app_abfallplus_de
      args:
        app_id: de.abfallwecker
        city: Lauchringen
[...]

PR #1539, opened 2023-12-19
Update app_abfallplus_de.md

halli77:
Updated doc for Abfall+ and added an example for Kreis Waldshut (Issue #1537), thank to the help of @5ila5


doc/source/app_abfallplus_de.md
```diff
@@ -141,7 +154,7 @@ The app_id can be found from the url of the play store entry: https://play.googl
 | de.k4systems.leipziglk | Landkreis Leipzig |
 | de.k4systems.abfallappbk | Bad Kissingen |
 | de.cmcitymedia.hokwaste | Hohenlohekreis |
-| de.abfallwecker | Rottweil, Tuttlingen, Waldshut, Prignitz, Nordsachsen |
+| de.abfallwecker | Rottweil, Tuttlingen, Kreis Waldshut, Prignitz, Nordsachsen |
```
5ila5, line 157: Will get overwritten by the update_doculinks.py script. Takes information from SUPPORTED_SERVICES in [custom_components/waste_collection_schedule/waste_collection_schedule/service/AppAbfallplusDe.py](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/service/AppAbfallplusDe.py) which is a generated output from the `generate_supported_services` function (same file).
Need to have a deeper look why it reports this istead of Kreis Waldshut

PR #1585, opened 2023-12-28
fix last few days in december having no dates

snow2k9:
Next collection dates were in next year so HA said "Unknown"


5ila5:
Thanks for your contribution
But I really did not like your approach with the while loop. So I made some changes: use dedicated function which will be call a second time in December (catching potential errors if data is not yet there). I reformatted the whole source as well (pre-commit hooks wiht .pre-commit-config.yaml settings)

snow2k9:
Hey,
no problem, was my first time contributing with Python. Thank you, works like a charm and looks way cleaner.

PR #1598, opened 2023-12-30
Added support for worms_de

omphteliba:
Added support for worms_de via generic  ICS source. Test went well.


5ila5:
Could you add a comment to replace the year in the link with "{%Y}" this will fetch the calendar for the current year (and tries next year in December) and do so in the examples.

could you please run the update_docu_links.py script and commit the modified files as well?

omphteliba:
@5ila5 thanks for the feedback. I hope the PR is better now! Let me know if I need to change something else.

5ila5:
Thanks for your contribution

I added the line : "- Replace the year in the url with `{%Y}` to keep the link valid for following years."
and changed the link to the webpage with one which will work.

omphteliba:
Thanks a lot for correcting my mistakes!

PR #1608, opened 2023-12-31
Update uttlesford_gov_uk.py

dearlove:
Fix logic to handle dates that are in the next year (website doesn't specify year).
Tested using test_sources.py and now returns the correct year:
python test_sources.py -s uttlesford_gov_uk
Testing source uttlesford_gov_uk ...
Tuesday 02 January 2024
Tuesday 09 January 2024
Tuesday 16 January 2024
Tuesday 23 January 2024
Tuesday 30 January 2024
Tuesday 06 February 2024
  found 12 entries for Brook Cottage, CM6 1LW
Thursday 04 January 2024
Thursday 11 January 2024
Thursday 18 January 2024
Thursday 25 January 2024
Thursday 01 February 2024
Thursday 08 February 2024
  found 12 entries for Springfields, CM6 1BP

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/uttlesford_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/uttlesford_gov_uk.py
@@ -1,4 +1,5 @@
 import datetime
+import calendar
 
 from bs4 import BeautifulSoup
 import requests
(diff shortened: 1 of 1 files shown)


5ila5:
I don't understand what's the different between your approach and the old one. The old approach increases the year by one but keeps the date the same. and your approach does the same but with adding the correct amount of days with way more complicated and harder to understand checks.

Can you please explain what bug does this fix (as the returned dates do not change using the old or your method (when replacing `datestr = trimsuffix(datestr) + " " + today.strftime("%Y")` with `datestr = trimsuffix(datestr) + " 2023"`)

dearlove:
Hi,
I wrote the original code and use this in my own home assistant installation. 
Once the scheduled bin collections crossed into 2024 it was returning the date in 2023 i.e. not incrementing the year properly. 
I have probably missed something as I can't see why the original method didn't work and after trying various ways to troubleshoot it couldn't find a way to fix it. When I tested using a completely different method to increment the year it worked straight away and returned all dates as 2024 so I committed it as a fix. 
I haven't logged a bug report but could do if required. 
It may be difficult to test the bug now that we are into 2024.
Thanks

dearlove:
I found the issue with the original code.
I wasn't assigning it back to the variable.  Embarrassing really.
I've reverted the more complicated fix and put in a much simpler fix.
Hopefully this is more acceptable.

5ila5:
Thanks for your contribution.

Just reformatted the source (based on .pre-commit-config.yaml) to match the formatting of the other code in this repo.

PR #1609, opened 2023-12-31
refactored to fetch monthly schedules and christmas tree collection

tr4nt0r:
refactored the code to fetch schedules as monthly ics files for 12 months in advance. This will download also schedules for the next year

removed  myquote method, standard url encoding works just fine

added download of christmas tree collection schedules. limited this feature to december and january only.


5ila5:
Thanks for your contribution 

I reformatted the code base on .pre-commit.yaml

PR #1613, opened 2024-01-01
Tewkesbury Borough Council API URL change + remove postcode option

tribey2:
Tewkesbury Borough Council recently changed their API URL, and now accepts UPRN only.

Therefore, I have updated the API URL and removed the UPRN option, and validated that this is working on my install.


5ila5:
Thanks for your contribution.
As the old method still works and I want to avoid breaking functioning configuration I added it back as a fallback and added a logger message to switch to a UPRN

PR #1622, opened 2024-01-02
Basildon source updated to new provider

Eyeanman:
Basildon Council have changed their API to mybasildon.powerappsportals.com, Bristol Council also uses a "powerappsportals.com" API so was used as the basis of this code, unfortunately the API responses are different so had to be amended.

Updated:
basildon_gov_uk.py
basildon_gov_uk.md

**BREAKING CHANGES**
API now requires a UPRN (previously postcode/address)


5ila5:
Thanks for your contribution.
I want to avoid breaking old configurations therefore I added the option to provide postcode and address as well.
Did some reformatting based on .pre-commit-config.yaml as well

Eyeanman:
Awesome, thank you!

PR #1626, opened 2024-01-03
Fix movar_no (Updated Kommunenummers for 2024 onward)

seedzero:
From January 2024 there are new Kommunenummers which breaks the script. This pull request changes to the new Kommunenummers.


5ila5:
you could remove this id completely to avoid this error from reoccurring. Is this a bad idea? could there be address duplicates returning the wrong id?

seedzero:
> you could remove this id completely to avoid this error from reoccurring. Is this a bad idea? could there be address duplicates returning the wrong id?

I just tested this and it is possible to remove the Kommunenummer, it still returns the same results. I'm not from Norway myself, I don't know if there is a possibility of address duplicates, but I imagine there is the possiblity of adding an adress from outside of the area and having it fail.

PR #1644, opened 2024-01-07
fix Geelong Australia boolean logic error from reformat

dtbell91:
When this was reformatted in #1509 it flipped whether it was an error based on the type or not. This should fix it up and match the preferred formatting.


5ila5:
Thanks for the correction

Issue #1584, opened 2023-12-28
[Bug]: EDG Dortmund Changed Format and Links


mhellmeier:
### I Have A Problem With:

A specific source

### What's Your Problem

The waste schedule for "EDG Entsorgung Dortmund" doesn't work anymore.

The examples and links described in `doc/ics/edg_de.md` don't work anymore. The company changed its links and `.ics` format.

### Source (if relevant)

edg_de

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

Xento:
The now seem to include the size of the bin and the interval in their ics file: Restabfall 80 l 02-wöchentl.

If you don't want to have individual types for the interval and size than you can use this config.
```
waste_collection_schedule:
  sources:
    - name: ics
      args:
        url: https://kundenportal.edg.de/WasteManagementDortmund/WasteManagementServiceServlet?ApplicationName=Calendar&SubmitAction=sync&StandortID=8278125001&AboID=89280&Fra=P;R;B;W
        regex: "^(\\w*) \\d* .*"
```

Than you have this four types:
- Restabfall
- Bioabfall
- Wertstoffe
- Altpapier

Maybe this can be added to the docs

mhellmeier:
Thanks @Xento, just tested your solution, and it works. Nice idea with the Regex to filter the other stuff away.

I can update the documentation so that other people are also informed.

PR #1645, opened 2024-01-07
fix: add regex to edg_de documentation example

mhellmeier:
This pull request updates the documentation of the EDG in Germany, as discussed in #1584 since they changed the format of the ics file.

Thanks to @Xento for the idea!


5ila5:
Thanks for your contribution I run the  ./update_docu_links.py script as your format did not match the one generated by the script

PR #1651, opened 2024-01-08
Adding My Utilitiy for Winnipeg MB Canada

coreyjansen:
Adding source for [My Utilitiy)](https://myutility.winnipeg.ca) in Winnipeg, MB

* [x]  Added custom_components/waste_collection_schedule/waste_collection_schedule/source/myutility_winnipeg_ca.py
* [x]  Added doc/source/myutility_winnipeg_ca.md
* [x]  Updated README.md via update_docu_links.py
* [x]  Updated info.md via update_docu_links.py
* [x]  Tested with test_sources.py

> Testing source myutility_winnipeg_ca ...
  found 98 entries for TestWinnipeg


5ila5:
Thanks for your contribution.

I made some reformatting/refactoring:
- genearal refromattin base on .pre-commit-config.yaml
- if not: continue instead of if .. to lower indentation depth
- address param provided with the params keyword instead of formatting it manually (only " ") 
- removed if before replace as replace only replaces if present by itself
- removed return [] after raise as it was unreachable

PR #1653, opened 2024-01-09
Add Service Montreal.ca

julienboriasse:
Currently, Montreal does not offer an iCal/Webcal subscription method.
The GeoJSON file provides sector-specific details.
The waste collection schedule is then interpreted from English natural language. Not every sector follows the same structure.
This method is not highly reliable but serves as an acceptable workaround until a better solution is provided by the city.


requirements.txt
```diff
@@ -10,3 +10,4 @@ requests>=2.31.0
 urllib3>=2.0.7
 jinja2>=3.1.2
 lxml>=4.9.4
+shapely
```
julienboriasse, line 13: Used to map address location to sector.

PR #1654, opened 2024-01-09
Add Montreal service

julienboriasse:
Currently, Montreal does not offer an iCal/Webcal subscription method.
The GeoJSON file provides sector-specific details.
The waste collection schedule is then interpreted from English natural language. Not every sector follows the same structure.
This method is not highly reliable but serves as an acceptable workaround until a better solution is provided by the city.


5ila5:
Thanks for your contribution. I did some major refactoring to remove your large function with very deep indentation. feel free to have a look. I checked the test output and it did not change.

I added the new dependency to the manifest.json as well to communicate the dependency to home-assistant

Issue #1655, opened 2024-01-09
[Bug]: source 'City of Karlsruhe' stopped working


alu-ka:
### I Have A Problem With:

A specific source

### What's Your Problem

Release 1.44.0:
Due to changes on the website the source '**City of Karlsruhe**' (name: karlsruhe_de) stopped working.
I start troubleshooting and add my findings here.

### Source (if relevant)

karlsruhe_de

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

alu-ka:
This code (file: custom_components\waste_collection_schedule\waste_collection_schedule\source\karlsruhe_de.py) works for me:
```
import datetime
import re

import requests
from bs4 import BeautifulSoup
from waste_collection_schedule import Collection  # type: ignore[attr-defined]

TITLE = "City of Karlsruhe"
DESCRIPTION = "Source for City of Karlsruhe."
URL = "https://www.karlsruhe.de/"
TEST_CASES = {
    "Östliche Rheinbrückenstraße 1": {
        "street": "Östliche Rheinbrückenstraße",
        "hnr": 1
    },
    "Habichtweg 4": {
        "street": "Habichtweg", 
        "hnr": 4
    },
    "Machstraße 5": {
        "street": "Machstraße", 
        "hnr": 5
    },
    "Bernsteinstraße 10 ladeort 1": {
        "street": "Bernsteinstraße",
        "hnr": 10,
        "ladeort": 1
    },
    "Bernsteinstraße 10 ladeort 2": {
        "street": "Bernsteinstraße",
        "hnr": 10,
        "ladeort": 2
     }
}


ICON_MAP = {
    "Restmüll": "mdi:trash-can",
    "Bioabfall": "mdi:leaf",
    "Papier": "mdi:package-variant",
[...]

PR #1657, opened 2024-01-10
Fix [Bug]: source 'City of Karlsruhe' stopped working #1655

alu-ka:
I have tested the changes on my local HA installation. Please review the code (as I'm not an experienced python developer). Thanks!

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/karlsruhe_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/karlsruhe_de.py
@@ -1,9 +1,8 @@
-import datetime
-import re
+from datetime import datetime
 
 import requests
-from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+from waste_collection_schedule.service.ICS import ICS
 
 TITLE = "City of Karlsruhe"
 DESCRIPTION = "Source for City of Karlsruhe."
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/karlsruhe_de.py
```diff
@@ -33,11 +39,11 @@
     "Bioabfall": "mdi:leaf",
     "Papier": "mdi:package-variant",
     "Wertstoff": "mdi:recycle",
-    "Sperrmüllabholung": "mdi:wardrobe",
+    "Sperrmüllabholung": "mdi:wardrobe"
 }
 
 
-API_URL = "https://web6.karlsruhe.de/service/abfall/akal/akal.php"
+API_URL = "https://web6.karlsruhe.de/service/abfall/akal_2023/akal.php"
```
MrRagga-, line 46:
```suggestion
API_URL = "https://web6.karlsruhe.de/service/abfall/akal/akal_2024.php"
```
Shouldn't we use the 2024 version, since they might be deprecate the 2023 soon, too.
The 2024 is also mentioned here https://tsk.karlsruhe.de/abfuhrkalender, the at the bottom of the page.

MrRagga-: Not really professional, that they do it like this, but it seems that they started "versioning" it every year.

alu-ka: You are right. In the meantime, they have changed the URL again. The other URL still works, but we should use the new one.
And yes, it's not really professional that they started to create a "version" every year. Let's see what happens next.

5ila5:
Thanks for your contribution. I prefer to use the proivded ical file as this reduces complexity of this source. I made the URL to use the current year.

PR #1660, opened 2024-01-11
Update awb_emsland_de.py

berndroewe:
Correction of description and URL for AWB Emsland.


5ila5:
Thanks for your contribution

PR #1671, opened 2024-01-14
feat: add `details_format`: `hidden`

swoga:
I would like to suggest an additional option for `details_format` to hide all information about upcoming collections in the attributes.


5ila5:
Thanks for your contribution

PR #1680, opened 2024-01-16
Fix buergerportal_de.py

shelker:
adding /api to buergerportal_de bedburg api url


5ila5:
Thanks for your contribution. Missed that one

PR #1607, opened 2023-12-31
Fix lrasha_de

steffenrapp:
The link to the calendar has changed as well as all of the IDs starting with the new year 2024.

I have tested the changes on my local installation.

PR #1683, opened 2024-01-16
CmCityMedia Update: Disable Regions

Newspicel:
I have added a disabled tag for regions that no longer use this cmcitymedia api. 
For example: #1607 

I checked this by looking to see if the regions were returning any data.


5ila5:
Thanks for your contribution

PR #1691, opened 2024-01-18
Update AWHas Documentation

auenkind:
Old Portal of Awhas has been discontinued, there is now a new Application that provides ical Files


5ila5:
you need to update the doc/ics/yaml/awhas_de.yaml file and run the update_docu_links.py script to update the md file or your changes will get overwritten

5ila5:
Thanks for your contribution. I added a comment to use the `{%Y}` as year placeholder.

PR #1702, opened 2024-01-20
Fix Cheshire East Garden Waste naming

greghesp:
Some naming changes within the Cheshire East feed broke the Garden Waste collection data.  
Added new logic to handle changes.


5ila5:
The old method seems fine for the old method could you provide a test case where the old method fails. This may prevent breaking things in potential future changes

5ila5:
Thanks for your contribtuion

Issue #1642, opened 2024-01-06
[Bug]: Exeter City Council date time bug


schklom:
### I Have A Problem With:

A specific source

### What's Your Problem

The integration does not work. It seems that `re.compile(REGEX_ORDINALS).sub("", d.text)` returns an empty string for some reason.

https://github.com/mampfes/hacs_waste_collection_schedule/blob/d57d1549f16371674a4c76c4224ced7f34870c00/custom_components/waste_collection_schedule/waste_collection_schedule/source/exeter_gov_uk.py#L47-L49

### Source (if relevant)

exeter_gov_uk

### Logs

```Shell
Logger: waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:136
Integration: waste_collection_schedule (documentation)
First occurred: 14:27:55 (2 occurrences)
Last logged: 14:37:49
fetch failed for source Exeter City Council: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 134, in fetch entries = self._source.fetch() ^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/exeter_gov_uk.py", line 47, in fetch date=datetime.strptime( ^^^^^^^^^^^^^^^^^^ File "/usr/lib/python3.11/_strptime.py", line 568, in _strptime_datetime tt, fraction, gmtoff_fraction = _strptime(data_string, format) ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ File "/usr/lib/python3.11/_strptime.py", line 349, in _strptime raise ValueError("time data %r does not match format %r" % ValueError: time data '' does not match format '%A, %d %B %Y'
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: exeter_gov_uk
      args:
        uprn: "10023120282"

sensor:
  - platform: waste_collection_schedule
    name: My_garbage
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

[...]

PR #1710, opened 2024-01-22
Source Exeter_gov_uk: Add check for missing collection dates

ReneNulschDE:
This is to fix issue: #1642 
A check for missing dates was added.

In the users example, the date for Garden waste collection was missing.


5ila5:
Thanks for your contribution

Issue #1682, opened 2024-01-16
[MANNHEIM not available via Abfall+ anymore]: 


RobertoCravallo:
### I Have A Problem With:

The integration in general

### What's Your Problem

Data for MANNHEIM is not available via Abfall+ anymore! They have their own APP now. 

Please remove Mannheim form the list. 

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

ReneNulschDE:
Mannheim is now using the same approach like "Offenbach am Main" (source: offenbach_de). Url for Mannheim:  https://www.insert-it.de/BmsAbfallkalenderMannheim

BlaM:
Has this been solved? The ticket is still open, but there is an updated example config in the documentation. When I try to use the example code from the documentation I run into `source not found: insert_it_de` errors in my logs, though.

weaseel:
Any news when the next release is going to happen. Almost missed putting out the bins because the new Mannheim app is garbage (pun intended)

derDeno:
Actually you can use it for now with a workaround until the next release releases. 

For that go to HACS and inside the waste collection integration there are three dots on the top right. There you can select to redownload the repo and select which version to download. Just select master instead of a specific version. This will download the latest state.

Sometimes you need to delete the current integration and redownload it to work.

weaseel:
Ah thanks. I checked „beta versions“ and was looking for „master“. But I never realized master is on the bottom of the list

PR #1714, opened 2024-01-23
Change Mannheim from AppAbfallplus to ics

ReneNulschDE:
This is to fix issue: #1682 

- Remove Mannheim from AppAbfallplus
- Add new source mannheim_de (copy of offenbach_de, without wizard)
- Add new doc source mannheim_de


5ila5:
As this does not implement anything you couldn't do with the generic ICS source, I would prefer providing documentation for the ICS source for Mannheim ([guide](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md))

derDeno:
Maybe it would make sense to implement insert-it as a new platform.
As there are multiple cities using this platform and they all work the same way, it would make sense to implement a wizard.

The wizard could use the app id to obtain the API base url and using two other API calls it's possible to generate the street ID and the location ID.

derDeno:
@5ila5 I understand what you mean. Personally, I'd prefer a simple wizard that keeps track of all the insert-it apps and eliminates even the few clicks needed for the ics link. But yes, I can see why it would also make sense to keep things simpler

5ila5:
closing this for #1717
@ReneNulschDE Thanks for your efforts anyway

PR #1719, opened 2024-01-25
Fix: Core - Add check for unknown source

ReneNulschDE:
If an unknown source is in the configuration, multiple exceptions are raised. Reason is that None is added to the source_shell list. This adds a check to prevent this.

Example exception:
```
  File "/custom_components/waste_collection_schedule/calendar.py", line 35, in async_setup_platform
    dedicated_calendar_types = shell.get_dedicated_calendar_types()
                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get_dedicated_calendar_types'
```

### DIFF
--- a/custom_components/waste_collection_schedule/__init__.py
+++ b/custom_components/waste_collection_schedule/__init__.py
@@ -198,14 +198,15 @@ class WasteCollectionApi:
         source_args,
         calendar_title,
     ):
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

Issue #1716, opened 2024-01-24
[Bug]: fetch failed for source Recycle!


Gunth:
### I Have A Problem With:

A specific source

### What's Your Problem

Hi,
Since this morning I've an error with the Recycle! source, all entities has the state unknown.
Please look a the logs bellow for more details.

Current installed version 1.15.0

G.

### Source (if relevant)

Recycle!

### Logs

```Shell
2024-01-24 11:54:26.937 ERROR (SyncWorker_6) [waste_collection_schedule.source_shell] fetch failed for source Recycle!:
Traceback (most recent call last):
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 134, in fetch
    entries = self._source.fetch()
              ^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/recycleapp_be.py", line 54, in fetch
    r.raise_for_status()
  File "/usr/local/lib/python3.11/site-packages/requests/models.py", line 1021, in raise_for_status
    raise HTTPError(http_error_msg, response=self)
requests.exceptions.HTTPError: 401 Client Error: Unauthorized for url: https://api.fostplus.be/recycle-public/app/v1/access-token
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: recycleapp_be
      calendar_title: recycle
      args:
        postcode: ----
        street: ----
        house_number: ----
        add_events: ----
      customize:
        - type: PMD
          alias: PMC
        - type: Déchets résiduels
          alias: Résiduels
        - type: Déchets organiques
          alias: Organiques
        - type: Glass
          alias: Verre
        - type: Paper-cardboard
          alias: Papiers-Cartons
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
[...]

PR #1722, opened 2024-01-25
Update x-secret in headers recycleapp_be

janholbrouck:
Recycleapp_be stopped working and it seems the issue was an updated x-secret header. In this PR I fix the x-secret header.

Fix for https://github.com/mampfes/hacs_waste_collection_schedule/issues/1716


5ila5:
Thanks for your contribution.
I really don't like this approach of hard-coding this token. But I don't see a good alternative as they hard-code it in their JavaScript and its URL does not look like it is a static URL either.

jlebutte:
It doesn't fix the issue

In addition to the updated secret value, it seems that the URL changed too. When I try via the website, I can see this call to the API : api.fostplus.be/**recyclecms**/app/v1/access-token
In the source file, the URL is api.fostplus.be/**recycle-public**/app/v1/access-token

PR #1739, opened 2024-01-29
Fix Mamirolle info first day of monthIgnore letters in special dates like '1er'

AghilasMessara:
Fix date parsing for cases like first day of month '1er'


5ila5:
Thanks for your contribution

PR #1760, opened 2024-02-04
Add dependency on calendars to manifest.json

codyc1515:
This integration has a calendar platform, therefore it should wait for that platform to be ready before initialising.


mampfes:
@codyc1515 I checked the built-in integrations in HA - no one is setting `calendar` as dependency. So, are you really sure that we need this?

codyc1515:
[Seems I was wrong](https://discord.com/channels/330944238910963714/554842238073700352/1204631236765556817).

PR #1765, opened 2024-02-05
Move Bonn Orange from AbfallApp to own source

ReneNulschDE:
Adding a new source Bonn Orange.
Fixes: #1762 

Based on Pforzheim and other "Athos"-based-solutions with modifications.

Removed BonnOrange from AbfallApp

Test cases: OK


wobeco:
That sounds like a change - but I'm unsure: what steps do I need to take now to experience the solution here?

ReneNulschDE:
> That sounds like a change - but I'm unsure: what steps do I need to take now to experience the solution here?

Just wait until this request gets accepted by the lead maintainers and a new version is released. After that a new source will be available for you.

wobeco:
Vielen Dank für die schnelle Reaktion :-)

5ila5:
Adding a dedicated source does not seem to be necessary as the app is till promoted on their website and still return valid data. I don't think the source itself is the problem of #1762

```
Testing source app_abfallplus_de ...
  found 128 entries for de.k4systems.bonnorange Auf dem Hügel
```

ReneNulschDE:
OK. I'm sure I have seen an info that the app will change - but I can't find it anymore. I'll close it - but we the needed changes...

Issue #1304, opened 2023-10-05
Entity could not be found


svenschedlbauer:
Hi, I don't know where I made the mistake, but I don't find the entity.
I have added following lines with the correct address to the configuration.yaml, but the name "zakb_de" can not be found.

```yaml
waste_collection_schedule:
    sources:
    - name: zakb_de
      args:
        ort: ORT
        strasse: STRAßE
        hnr: HAUSNUMMER
        hnr_zusatz: HAUSNUMMERZUSATZ
```

> edit by @5ila5: fixed yaml fomatation

5ila5:
> "zakb_de" can not be found.

what do you mean with that exactly? can you please provide the error log or further explain what you mean by that. 
the yaml looks fine and the source seems to work on my side

5ila5:
yes if you did not set a different `calendar_title` you should see your calendar entity.

Have you restarted homeassistant after modifying your configuration (quick reload does not work for this integration)?

Are there any logs indicating an error?

svenschedlbauer:
Yes, I have restarted HA 2 or 3 times in the meanwhile. 
- I have installed waste collection schedule, then restart
- I have added the code to the configuration.yaml, then restart

5ila5:
Are there any logs for waste_collection_schedule

svenschedlbauer:
Im Log finde ich folgendes:

Logger: homeassistant.setup
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source/zakb_de.py:54
Integration: waste_collection_schedule ([documentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
First occurred: 16:19:34 (1 occurrences)
Last logged: 16:19:34

Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 288, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 109, in async_setup
    api.add_source_shell(
  File "/config/custom_components/waste_collection_schedule/__init__.py", line 202, in add_source_shell
    SourceShell.create(
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 196, in create
    source = source_module.Source(**source_args)  # type: ignore
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #1774, opened 2024-02-06
Update zakb_de.py post request parameters

jgmGit:
There seems to be a change in the post request parameters for zakb. Please check the attached image

image

I have updated it and also added some tests for the failing addresses. 
The old tests still seems to be working.


jgmGit:
Fix this issues btw
https://github.com/mampfes/hacs_waste_collection_schedule/issues/1304

PR #1802, opened 2024-02-18
Fixes Abfall.IO for AWB Landkreis Göppingen (#1783)

### LINKED ISSUE #1783: [Bug]: abfall_io source for f_id_kommune 899 (AWB Göppingen) needs a time period in the request to get a valid ICS response
### I Have A Problem With:

A specific source

### What's Your Problem

My waste scheduler source for f_id_kommune: 899 stopped working a few days ago. After some tests to verify my configuration, I realised that the request from generated from the awb-gp.de adds a timeperiod=20240101-20241231 parameter.
Without this, the ICS has no values
For the moment I added the generated URL as ICS source and it works now. But the Source abfall_io seams to be broken without this timeperion parameter (which cant be set in the configuration)

### Source (if relevant)

_No response_

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
- name: abfall_io
      args:
        key: !secret abfall_io_key
        f_id_kommune: 899
        f_id_strasse: XXX
        f_abfallarten:
          - 20
          - 17
          - 59
          - 18
          - 19
          - 60
      calendar_title: "Müllkalender"
      customize:
        - type: Hausmüll
          icon: mdi:trash-can
        - type: Gelber Sack
          icon: mdi:recycle
        - type: Bioabfall
          icon: mdi:biohazard
        - type: Grüngutsammlung
          icon: mdi:leaf
```

### Checklist Source Error

### Checklist Sensor Error

### Required

fabssor:
Gracefully fixes #1783 by not destroying old configurations.
For AWB Landkreis Göppingen the API seems to be changed. 
Now the wizzard uses the GraphQL API to return the correct configuration. Still uses reponse package for communication with the database to don't bring in any new dependency.
Allows the source abfall_io to have to different configurations.
Does not break old configurations and allows if other communes switch to the new GraphQL API to be integrated easily.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/AbfallIO.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/AbfallIO.py
@@ -45,7 +45,7 @@ SERVICE_MAP = [
     {
         "title": "AWB Landkreis Göppingen",
         "url": "https://www.awb-gp.de/",
-        "service_id": "365d791b58c7e39b20bb8f167bd33981",
+        "service_id": "f35bd08b1d18d9c81fcdee75dbcce5d3",
     },
     {
         "title": "Göttinger Entsorgungsbetriebe",
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/abfall_io.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/abfall_io.py
@@ -71,6 +71,11 @@ TEST_CASES = {
         "f_id_strasse": 3475,
         "f_id_strasse_hnr": 185575,
     },
+    "Göppingen": {
+        "key": "f35bd08b1d18d9c81fcdee75dbcce5d3",
+        "idhousenumber": 7074,
+        "wastetypes": [20,17,59,18,19,60]
+    },
 }
 _LOGGER = logging.getLogger(__name__)
 
@@ -99,11 +104,13 @@ class Source:
     def __init__(
         self,
         key,
-        f_id_kommune,
-        f_id_strasse,
+        f_id_kommune = None,
+        f_id_strasse = None,
         f_id_bezirk=None,
         f_id_strasse_hnr=None,
         f_abfallarten=[],
+        idhousenumber = None, # New API
+        wastetypes = None, # New API
     ):
         self._key = key
         self._kommune = f_id_kommune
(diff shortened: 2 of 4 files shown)


fabssor:
While digging around I also found Böblingen which can use the GraphQL API. 
I added it as service "AWB Landkreis Böblingen 2".
I also added it as Testcase with the new and the old config. We can see both return the same Data.

hphde:
Thanks for the Böblingen fix @fabssor

hphde:
Possible waste Types Böblingen from the query:
```json
{
	"wasteTypes": [
		"50",
		"53",
		"299",
		"31",
		"326",
		"325",
		"333",
		"327",
		"334",
		"335",
		"328"
	]
}
```

fabssor:
@hphde the possibles wastes should be correctly returned by the wizzard, aren't they?

hphde:
Are you referring to https://www.awb-bb.de/start/entsorgen/Abfuhrtermine.html? This is where I got the data from

fabssor:
@hphde No there is a wizard python script in this Repo: custom_components/waste_collection_schedule/waste_collection_schedule/wizard/abfall_io.py
This should also return you the YAML configuration and should give you all the available waste types

fabssor:
The new Version in this PR should Show the waste Types for "Böblingen 2"

5ila5:
This approach adds quite a lot of complexity. The approach with the wizard is not ideal anyway, as a lot of people struggle running them correctly, and we generally want to remove/reduce the dependence on those wizards. The ICS file you can get on the new API base looks like it will stay valid and does not require any post parameters to be looked for. So I think a better approach is to use the ICS source and add documentation for supported providers using an [ICS source](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md). And adding a note in the abfall.io markdown and maybe add not working providers to an OUTDATED variable to add a logger.warning or logger.error advising to change to the ICS version.

@fabssor what do you think?

5ila5:
@hphde that's the point of the ICS source you just copy the URL from the website in the ICS configuration and everything works

I opened a first version PR (#1958), please feel free to comment

fabssor:
@5ila5 I liked the concept of the wizzards but its not my decision to take. I tried the ICS way and for sure it works out flawlessly. I will close this PR.

PR #1810, opened 2024-02-19
add source multiple to combine multiple sources to one Entity

5ila5:
This addresses issues like #1801

@mampfes what do you think about this approach?


custom_components/waste_collection_schedule/waste_collection_schedule/source/multiple.py
```diff
@@ -0,0 +1,139 @@
...
+                "until": "2022-12-31",
+            },
+        ],
+    },
+}
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_soruce(source: str, args: dict | list[dict]) -> list:
```
mampfes, line 80: typo

mampfes (approved):

PR #1812, opened 2024-02-20
Make use of new Renosyd (DK) API and update `devcontainer`

carlreid:
While the old(?) API might still be kept around, Renosyd now advises to check schedule on https://mit.renosyd.dk. As such, I checked the API call made on there and sure enough it's to a different service which I thought I would try contribute an update here.

Would be great if @lexitus could check this over in relation to the `renosyd_dk` changes, and of course feel free to contribute directly to the fork to make any adjustments 👍

image

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,33 +1,38 @@
 // See https://aka.ms/vscode-remote/devcontainer.json for format details.
 {
 	"name": "Waste Collection Schedule development",
-	"image": "ghcr.io/ludeeus/devcontainer/integration:stable",
-	"context": "..",
+	"image": "mcr.microsoft.com/devcontainers/base:ubuntu",
 	"appPort": [
 		"9123:8123"
 	],
 	"containerEnv": {
 		"TZ": "Europe/Berlin" // Set your local timezone here
 	},
-	"postCreateCommand": "container install",
-	"extensions": [
-		"ms-python.python",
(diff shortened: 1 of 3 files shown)


lexitus:
Thanks @carlreid , I had just today seen that the new schedule page was up and was thinking about looking into it! 

Changes to renosyd_dk look good, much shorter code than before. I tested for myself, and the results check out fine. 

- I guess we can remove some of the obsolete entries in ICON_MAP, but maybe it's better to wait until the transition period to the new bin types is done.
- Storskrald seems to have been moved to its own endpoint, but since you have to order it anyway I'm not sure I see the value in adding it back.

Can't comment on the devcontainer changes.

carlreid:
Thanks for taking the time to check it over @lexitus. I've removed the storskrald entry, and agree we can see which other entries emerge as the transition happens.

5ila5:
So we wait merging this?

carlreid:
I think the additional tweaking of any new types/entries in `ICON_MAP` can happen later once we observe the new types appear on the API. The changes in this PR already to include some of the types (_Glass, Metal, and Hard Plastic_).

I think merging this sooner would be most beneficial to get the new collection dates from the newest source.

lexitus:
Yes, agree it would be good to get this merged as the old API is not being updated past 1st March.

5ila5:
Thanks for your contribution.

I did some reformatting and allowed integer house_number arguments to avoid confusion while configuration

PR #1817, opened 2024-02-21
renoweb dk

r-poulsen:
$ test_sources.py -s renoweb_dk
Testing source renoweb_dk ...
  found 26 entries for test_01
  found 205 entries for test_02
  found 345 entries for test_03

An issue when using address_id is fixed and error handling and documentation is improved.


5ila5:
Thanks for your contribution

PR #1839, opened 2024-02-27
heilbronn_de fix duplicate bug

5ila5:
fixes bug where collections are returned multiple times instead of once


Eisbaer2:
Thank you @5ila5. Everythings fine now 👍

PR #1863, opened 2024-03-03
Added support for the London Borough of Newham

richkershaw:
Support for Newham's bin collections site, which uses UPRNs.


5ila5:
Thanks for your contribution.

- Did some minor refactoring
- Allowed integer UPRNs using zfill(12) (removed need to include starting `0`s
- minor changes in md file

PR #1864, opened 2024-03-04
Improves SRV återvinning AB implementation

clind922:
- Adds city as a configurable argument.
- Updated the documentation with more up to date configuration information and added examples.


5ila5:
Thanks for your contribution.

PR #1875, opened 2024-03-08
Add source for Hornsby Shire Council waste collection schedule.

sh00t2kill:
HSC publish this as a PDF format, but there is a rotation schedule based on what "zone" you are in.

I built an intermediary API for this to connect to, which makes it easier to maintain, and the source connects to this API endpoint.


custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
```diff
@@ -0,0 +1,66 @@
...
+        bin_this_week = data['bins']['this_week']
+        entries.append(Collection(date=green_collection_date, t="Green Waste", icon=ICON_MAP.get("Green")))
+        entries.append(Collection(date=yellow_collection_date, t="Recycling", icon=ICON_MAP.get("Yellow")))
+        entries.append(Collection(date=yellow_collection_date, t="General Waste", icon=ICON_MAP.get("Refuse")))
+        entries.append(Collection(date=green_collection_date, t="General Waste", icon=ICON_MAP.get("Refuse")))
+        if self.show_nights:
+            entries.append(Collection(date=green_collection_date-datetime.timedelta(days=1), t="Green Waste Tonight", icon=ICON_MAP.get("Green")))
+            entries.append(Collection(date=yellow_collection_date-datetime.timedelta(days=1), t="Recycling Tonight", icon=ICON_MAP.get("Yellow")))
+            entries.append(Collection(date=yellow_collection_date-datetime.timedelta(days=1), t="General Waste Tonight", icon=ICON_MAP.get("Refuse")))
+            entries.append(Collection(date=green_collection_date-datetime.timedelta(days=1), t="General Waste Tonight", icon=ICON_MAP.get("Refuse")))
```
5ila5, line 64: I really do not like this option

5ila5 (commented):
Please add a markdown file for documentation purposes (please add a note that the information is not fetched from the official council website) and run the update_docu_links.py script. [Contribution guide](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md)

custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
sh00t2kill: Me either. Will remove it.

I didn't see that guide so will add what's needed.

sh00t2kill:
Thanks @5ila5. Please let me know if anything else needs fixing or changing.

5ila5:
Thanks for your contribution.
- Did some reformatting (based on .pre-commit-config.yaml)
- made one test case integer to test integer and string values

PR #1900, opened 2024-03-20
Add source for Port Stephens Council

sh00t2kill:


5ila5 (changes requested):
Thanks for your contribution. I would like to suggest some changes:

- please remove commented out print statements
- please add a second test case with an int street_number to test both versions (users should not care whether they are using integers or strings. Lead to problems in the past)
- I'd prefer the params argument instead of passing parameters as part of the URL which make it pretty hard to read (e.g. `API_URLS["Collection"] + str(property_id) + f".json?start={last_year}-12-31T13:00:00.000Z&end={this_year}-12-30T13:00:00.000Z"`

doc/source/portstephens_nsw_gov_au.md
```diff
@@ -0,0 +1,42 @@
+# Ku-ring-gai Council
```
5ila5, line 1: Wrong Headline?

sh00t2kill:
I was writing this on public transport; Looking at it with fresh eyes today I can tell i quickly merged and created the PR a few minutes from my stop :)

Should all be fixed up now.

5ila5:
Thanks for your contribution.
I did some refactoring:
- removed the remaining arguments not passed with the params argument
- removed duplicate code by adding a get_id function

and some reformatting (based on .pre-commit-config.yaml)

PR #1911, opened 2024-03-25
Update braintree_gov_uk.py

amildenhall:
Proposal of two fixes for source "braintree_gov_uk.py"

1. Only the first fetch after a HA reboot would actually succeed. Tracked this down to self.form_data keeping data from previous fetches which caused subsequent fetches to fail. My proposed fix is a re-initialize of self.form_data upon every fetch. (See commit f5a92b99f9f1756115b0fa9eeefe9f56f550b6a9)
2. Braintree council have changed the name of their "Green Bin" to "Garden Bin". Have updated this in the ICON_MAP. (See commit 3c6f70416a51071e95626686eb52a6f450d88ae7)


5ila5:
Thanks for your contribution

PR #1917, opened 2024-03-28
fix rctcbc_gov_uk: parsing new hidden calendar fields

### LINKED ISSUE #1916: [Bug]: rctcbc_gov_uk returning extra incorrect dates
### I Have A Problem With:

A specific source

### What's Your Problem

I've been using this integration for a while without issue.
I've been away for a few weeks, and come home to find 3 entries for each entry for each week, when there should only be one.

See attached screenshot.
The Wednesday entries are the only valid ones. Not sure why they're being duplicated though. :-/

Screenshot 2024-03-28 at 03 44 55

### Source (if relevant)

rctcbc_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: rctcbc_gov_uk
      args:
        uprn: "200003766278"
      calendar_title: "Bin Collection"
```

### Checklist Source Error

### Checklist Sensor Error

### Required

5ila5:
They added a hidden `printableCalendar` which contains all available months of data. The old script found all events in the visible and the printableCalendar but associated them all with the month of the visible calendar which lead to a lot of wrong additional entries. This fix tries to parse the printableCalendar first which saves useless web requests. If it fails it falls back to the old approach.

fixes #1916

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/rctcbc_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/rctcbc_gov_uk.py
@@ -1,8 +1,8 @@
 from datetime import datetime
 
 import requests
-from bs4 import BeautifulSoup
-from waste_collection_schedule import Collection
+from bs4 import BeautifulSoup, Tag
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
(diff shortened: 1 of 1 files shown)


neildotwilliams:
Works well for me. Thanks @5ila5

PR #1925, opened 2024-04-01
Added a source for Eigenbetrieb Abfallwirtschaft Landkreis Spree-Neiße

homeassist24:
Added a source for "Eigenbetrieb Abfallwirtschaft Landkreis Spree-Neiße". Created a .py script and the .md file for this source. Updated the README and info.md files. Added 4 tests with the following output ...

Testing source eigenbetrieb_abfallwirtschaft_de ...
  found 56 entries for Forst (Lausitz), Rosenweg
  found 57 entries for Peitz, Am See
  found 56 entries for Guben, Altsprucke
  found 58 entries for Spremberg, Gartenstrasse


custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_lkspn_de.py
```diff
@@ -0,0 +1,96 @@
...
+
+# URL, DESCRIPTION  and TITLE of the website
+TITLE = "Eigenbetrieb Abfallwirtschaft Landkreis Spree-Neiße"
+DESCRIPTION = "Source for Eigenbetrieb Abfallwirtschaft Landkreis Spree-Neiße."
+URL = "https://www.eigenbetrieb-abfallwirtschaft.de"
+COUNTRY = "de"
+TEST_CASES = {
+    "Forst (Lausitz), Rosenweg": {"city": "4", "street": "344"},
+    "Peitz, Am See": {"city": "8", "street": "1077"},
+    "Guben, Altsprucke": {"city": "5", "street": "410"},
```
5ila5, line 15: Please make one test case all Ints as the yaml input of integers without `"` will lead to an integer argument, which should still work as users should not care what type their argument has. (lead to problems in the past (e.g. #1935))

5ila5 (changes requested):
Why did you name this source `aw_lkspn_de`? The URL is `eigenbetrieb-abfallwirtschaft.de` so our general naming convention would be `eigenbetrieb_abfallwirtschaft_de`

custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_lkspn_de.py
homeassist24: Added a test with integer input. Test run successful ...

homeassist24:
> Why did you name this source `aw_lkspn_de`? The URL is `eigenbetrieb-abfallwirtschaft.de` so our general naming convention would be `eigenbetrieb_abfallwirtschaft_de`

You're right. Renamed as suggested.

5ila5 (approved):

5ila5:
Thanks for your contribution.
Just refactored your ifs to remove the unnecessary indentation depth.

PR #1943, opened 2024-04-05
Adding staffordbc gov uk

bicciess:
Added a source for Stafford Borough Council. Unfortunately, I'm only able to scrape the next collection dates for the normal refuse and recycling. 
In Stafford both Blue (Mixed Recycling) and Brown (Garden) bins are collected on the same day but in the code I only present this as "Blue Bin".
I'm a bit new to github so I'm hoping I've done this all correctly.


5ila5:
Thanks for your contribution

PR #1956, opened 2024-04-09
Adding hawkesbury nsw gov au

zzznz27:
# Hawkesbury City Council Source 

## Overview 
Added source for the Hawkesbury city council in NSW Australia

This source uses the API available to pull the different collection dates for each type of garbage and expands the schedule for 52 weeks. 

This API does not get updated often and as a result the weekly schedule has to be inferred from the provided dates

Documentation has also been updated.

### test_sources.py output
```
hacs_waste_collection_schedule\custom_components\waste_collection_schedule\waste_collection_schedule\test> python test_sources.py -s hawkesbury_nsw_gov_au
Testing source hawkesbury_nsw_gov_au ...
  found 104 entries for south windsor, 539a George Street
  found 104 entries for Windsor, catherine street 7
  found 104 entries for Kurrajong, Bells Line Of Road 1052 
```


5ila5 (changes requested):
Organic waste needs to be optional (last test case fails) are 52 weeks enough (how often do they update the week1 attribute?) maybe calculate an appropriate amount `weeks = ((dt.datetime.now() +dt.timedelta(days=365) ) -  basedate).days // 7`


they provide a {bin_prefix}_schedule property for the frequency, so I think you should use it

You have a lot of duplicate code I'd prefer to put it in an additional method.


Was just testing around (feel free not to use it):

```python
    def get_data(self, bin_prefix:str, fields) -> list[Collection]:
        entries:list[Collection] = []

        frequency = int(fields.get(f'{bin_prefix}_schedule'))
        if frequency == 0:
            return entries
        
        base_string = fields.get(f'{bin_prefix}_week1', dt.datetime.min.strftime('%Y-%m-%d'))
        basedate = dt.datetime.strptime(base_string, "%Y-%m-%d")

        days = ((dt.datetime.now() +dt.timedelta(days=365) ) -  basedate).days

        date_list =[basedate + dt.timedelta(days=x) for x in range(0, days, frequency)]
        name = bin_prefix.replace("bin", "").title()
        for dateStr in date_list:
            entries.append(Collection(dateStr.date(), name, ICON_MAP.get(name.upper())))
        return entries

    def fetch(self):
        # .......
        for key in entry["fields"].keys():
            if key.endswith("_schedule"):
                entries += self.get_data(key.replace("_schedule", ""), entry["fields"])
        return entries


```

zzznz27:
I've made changes as recommended. 
Ive kept the bin types static to keep it easy to read and i don't foresee any other bin types being added anytime soon

5ila5 (approved):

PR #1957, opened 2024-04-09
add "include_ordered_pickup_entries" setting to renoweb_dk

syre:
Adds a setting to `renoweb_dk` for also including entries that are not collected automatically but should be ordered for pickup (`afhentningsbestillingmateriel = True`).

For "fh" municipality that is for example:
- Farligt affald (hazardous waste)
- Tekstilaffald (textile waste)
- Storskrald (large objects waste)
- Haveaffald (garden waste)

Tests:
```bash
./test_sources.py -s renoweb_dk --weekday
Testing source renoweb_dk ...
  found 26 entries for test_01
  found 195 entries for test_02
  found 291 entries for test_03
  found 56 entries for test_04
```


5ila5:
Thanks for your contribution

Issue #1962, opened 2024-04-10
[Bug]: Auckland City Council stopped working around update to HA 2024.4.2 or 1.47.0


xbmcnut:
### I Have A Problem With:

A specific source

### What's Your Problem

All sensors for `aucklandcouncil_govt_nz` now show unavailable. Seems to have stopped working around either my recent update to 1.47.0 or HA 2024.4.2. Configuration has not changed at all. Tried downgrading to 1.45.0 and 1.46.0, still the same. Reinstalled 1.47.0 and still no go.

![image](https://github.com/mampfes/hacs_waste_collection_schedule/assets/1492137/44bc1f6d-4a70-46ea-be25-7db66dcae72b)

### Source (if relevant)

aucklandcouncil_govt_nz

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: aucklandcouncil_govt_nz
      args:
        area_number: redacted
        # from https://www.aucklandcouncil.govt.nz/rubbish-recycling/rubbish-recycling-collections/Pages/rubbish-recycling-collection-days.aspx
      customize:
        - type: rubbish
          alias: Rubbish
          icon: mdi:trash-can
        - type: recycle
          alias: Recycling
          icon: mdi:recycle
      calendar_title: Auckland Council Rubbish & Recycling Collections
  day_switch_time: "17:00"
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

sahne2000:
Your are not alone. Same issue. I had a quick look yesterday and from what I can see Auckland council website has slightly changed. I could not finish it actual work task took over. Hopefully today or someone else has time to adjust.

BambamNZ:
[...]

PR #1968, opened 2024-04-12
[bug fix] Fix up Auckland Council - appears to be a website change

nicwise:
Looks like the council have changed their site.

So lets move this over to use Beautiful Soup because who wants to keep state in booleans.

Tests run clean with

```
python3 test_sources.py --source aucklandcouncil_govt_nz -l
```

fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/1962


5ila5:
Thanks for your contribution.

- Did some reformatting (base d on .pre-commit-config.yaml)
- made two test cases int as both should be possible and therefore tested

nicwise:
nice - thank you!

kellyatkinson:
I haven't been able to get this source working since this change was made last week. A few of us have reported this: https://github.com/mampfes/hacs_waste_collection_schedule/issues/1985

Issue #1940, opened 2024-04-04
[Feature]: Add Garden Waste option to waver.


mjdyson:
### I propose a feature for:

Sources

### Describe your wanted feature

Garden Waste collection day is not currently provided by the website API, however it can be inferred.

The Waverley website states that garden waste is collected on the same week as as the domestic waste
https://www.waverley.gov.uk/Services/Bins-and-recycling/Rubbish-and-recycling-collections/Check-bin-collection-day

See excerpt:

![image](https://github.com/mampfes/hacs_waste_collection_schedule/assets/4598068/fe82c899-a617-4809-895d-24667eb7e4c5)

I have a working prototype code that may be useful for other users - i'm happy to PR or discuss alternatives:
* in the configuration, the user provides the day of the week that their garden waste is collected as an int (optional)
* when building the collection from the website scraper, the code inserts a garden waste entry every time it encounters a "domestic" item, using an offset (in days)

```
for u1 in u1s:
            lis = u1.find_all("li", recursive=False)
            entries.append(
                Collection(
                    date=datetime.strptime(
                        lis[1].text.replace("\n", ""), "%d/%m/%Y"
                    ).date(),
                    t=lis[2].text.replace("\n", ""),
                    icon=ICON_MAP.get(
                        lis[2].text.replace("\n", "").replace(" Collection Service", "")
                    ),
                )
            )
            if self._garden_waste:
                if "Domestic" in lis[2].text:
                    date = datetime.strptime(
                                lis[1].text.replace("\n", ""), "%d/%m/%Y"
                            ).date()
                    entries.append(
                        Collection(
                            date - timedelta(datetime.weekday(date) - self._garden_waste),
                            t="Garden Waste Collection Service",
                            icon=ICON_MAP.get("Garden Waste")
                        ),
                    )
```

kalmatthew:
This used to work and I can't quite work out when it failed. Also despite that text if you use their online checker it does display the dates for the brown garden waste bins

kalmatthew:
That is odd, the really obvious thing to ask is if you renewed it for this year? Assuming you did, what day does it get collected (mine is Wednesday with domestic waste being Friday). When I have a chance to get to a proper computer I'm also going to look at the source for the page I get to try and understand why the parse doesn't work for me, when I can see the detials.

mjdyson:
[...]

PR #1987, opened 2024-04-17
Update waverley_gov_uk.md

kalmatthew:
Added clearer instructions about ensuring only a single entry for your search params and ignoring the guidance from waverly to use only Postcode the issue was discovered when attempting to debug for #1940

### DIFF
--- a/doc/source/waverley_gov_uk.md
+++ b/doc/source/waverley_gov_uk.md
@@ -20,7 +20,7 @@ waste_collection_schedule:
 ### Configuration Variables
 You must supply enough address details for the search to find your property as the first match.
 
-Test this out manually first at [Waverley Borough Council](https://www.waverley.gov.uk/Services/Bins-and-recycling/Rubbish-and-recycling-collections/Check-bin-collection-day) if you are not sure which are needed.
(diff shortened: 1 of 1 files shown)


mjdyson (commented):
- Typo on line 23 "webesites". 
- If we're updating the help file, we may want to add advice on garden waste and the "personalised" results that are linked to the status of a property's subscription status. 
- Also on the Waverley website, a search for 2 + Postcode will return houses 2, 12, 20, 21, 22, 23 etc. Given the nature of the data being linked to address, there may be cases when this wont ever be a single result. I dont know if this can be easily adjusted with beautiful soup, or if it's just how the search works on the bin checker.

5ila5:
Thanks for your contribution.

The request to search takes the parameters: address_name_number, address_street, street_town, address_postcode. You're right there could be multiple matches (house number) But I do not have an elegant solution for this problem as matching the values for which fits best is not that easy as addresses can look a bitt different from one another:

14, BUSBRIDGE LANE, GODALMING, SURREY
GLENDREE HOUSE, 10, BUSBRIDGE LANE, GODALMING, SURREY

kalmatthew:
Is it not possible to select the link that matches the name/number given? I was hoping to have a play around one evening next week.

5ila5:
You probably could probably design something, maybe like:

```python
addr_string = "..."
addr_arr = addr_string.split(", ")


name = addr_arr[-5] if len(addr_arr) >= 5 else ""
number = addr_arr[-4]
# ....
#
# ------ some matching ------
```

But something like this does increase the complexity and adds more points of failure.

Feel free to tinker around and if you think you have a good solution, open a PR.

PR #2039, opened 2024-04-30
Localcitis bueren

codegengrossen:


5ila5:
I do not like this approach: 

- It looks like the exact same script should work for multiple municipalities, so we should use this and use `municipality: bueren-an-der-aare` and `id: 849` as parameter, this way we can support all municipalities that provide a waste collection calendar there. 
- waste_type should not be a partner the source should get all entries and the user can remove entries using customize if he wants to.

codegengrossen:
The script can be revised again. As far as I can see, the location and the ID would have to be specified. I will provide a revised version

bbr111:
@codegengrossen 
Please update PR
I think 2 more source requests are waiting for this.

or did you find something else?

markvp:
Thanks for submitting this! A generic  source already exists in the repo that supports any municipality on the localcities.ch platform. Büren an der Aare has now been added to that source (along with Liestal and Therwil from #2970 and #4398) in PR #6069. Closing this in favour of that PR.

PR #2043, opened 2024-05-01
add new remondis source

lal12:
Add a new remondis source. Remondis is a waste disposal company operating in a bunch of cities. It offers an (old) app for showing the next dates for waste collection. I reverse engineered the app and use it's API.

There might be still some issue left. The API needs a device registration, currently I do this on each init, but this might get us in trouble in the future. So maybe there is a way storing the session somewhere via the HomeAssistant API?


lal12:
Will comment them out.

Yeah good point, it doesn't seem to always be obvious that Remondis removes the waste in some cities, they probably act as a sub or in cooperation. But the list I can retrieve from the API also seems to be more than what they actually do, since it gives back no dates. Probably cities they don't service anymore. So the list would contain "false positives".

And btw. it takes 26 requests (~30-60s) to generate that full list of cities.

5ila5:
I just noticed you do web request on every update_docu_links.py call. I do not like this, the script is probably called quite a lot and should therefore not make any web requests. Maybe hard-code them with a function to get a new list. You could try to filter out the ones that do not provide any collections and add them to the EXTRA_INFO variable

markvp:
Hi @lal12, thanks for the work on this! This PR has been open since May 2024, and the reverse-engineered API with device registration adds significant maintenance risk. We'd suggest closing this for now. If you'd like to revisit with a cleaner API approach, please feel free to open a new issue or PR!

markvp:
Closing as the implementation approach isn't viable for inclusion (see prior review feedback) and there's been no further activity. Happy to reopen if you find a dynamic data source (API/ICS/scrapable site) — please open a fresh issue or PR with that approach.

PR #2049, opened 2024-05-03
Add support for Birmingham City Council (UK), birmingham.gov.uk

dannya:
image


custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
```diff
@@ -0,0 +1,95 @@
...
+from bs4 import BeautifulSoup
+from waste_collection_schedule import Collection
+
+TITLE = "Birmingham City Council"
+DESCRIPTION = "Source for birmingham.gov.uk services for Birmingham, UK."
+URL = "https://birmingham.gov.uk"
+TEST_CASES = {
+    "Cherry Tree Croft": {"uprn": "100070321799", "postcode": "B27 6TF"},
+    "Ludgate Loft Apartments": {"uprn": "10033389698", "postcode": "B3 1DW"},
+    "Windermere Road": {"uprn": "100070566109", "postcode": "B13 9JP"},
```
5ila5, line 16: Please make at least one test case test an integer UPRN (users should not worry about whether their input is handled as a string or an integer, and we had problems with this in the past) (also Your example in the md file is read as an integer)

custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
```diff
@@ -0,0 +1,95 @@
...
+        for table_row in collection_soup.find("table", class_="data-table").tbody.find_all("tr"):
+            collection_type = table_row.contents[0].text
+
+            if collection_type in ICON_MAP:
+                collection_next = table_row.contents[1].text
+                collection_date = re.findall("\(.*?\)", collection_next)
+
+                if len(collection_date) == 1:
+                    collection_date_obj = parse(re.sub("[()]", "", collection_date[0])).date()
+
```
5ila5, line 76: I'd prefer if `collection_type not in ICON_MAP: continue` and `len(collection_date) != 1: continue` to remove unnecessary indentation which make the code harder to read

custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
dannya: Updated.

custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
dannya: Updated.

5ila5 (approved):

5ila5:
LGTM, Thanks for your contribution (did some minor reformatting based on .pre-commit-config.yaml)

PR #2068, opened 2024-05-08
Add config flow (UI) configuration option

### LINKED ISSUE #2016: [Feature]: Config Flow (UI) setup
### I propose a feature for:

The integration in general

The preferred method of setting up integrations seems to now be using the config flow (add integration... menu) rather than yaml. As far as I know this can exist alongside yaml setup, and would allow either setup route.

If there is interest in this, I'm happy to work on it.
Perhaps a basic configuration through config flow and then for more advanced setup, keep it to yaml?

dan-r:
Implements config flow configuration whilst retaining support for YAML configuration. Resolves #2016.

There's some automated testing needed, but I'd appreciate some community testing with this fork as so far I've only tested with a few areas local to me.

Customise functionality is not supported as this time, but the basic functionality (incl. ICS) is there.

This feature borrows very heavily from the config_flow branch, but takes a different approach with how sources are handled. Instead of templating the language and constants files, a JSON list of sources is written by update_docu_links.py and read during the config flow. 

https://github.com/mampfes/hacs_waste_collection_schedule/assets/1384852/1fe6ce3f-051d-4151-b5cf-dae5958fbb27


5ila5:
- Seems to work fine I had an error but could not reproduce it again for some reason (logs, see below).
- I think it would be nice to "move" the wizards to the UI. I thought about a check whether the source or module has a specific method providing values to select.
- Type checking: what about an optional dictionary containing types of attributes. It might be nice to have to also be able to set `vol.Exclusive` as we have a lot of sources you can either configure with some kind of ID and with an address (e.g. jumomind_de allows city street house_number or city_id and area_id)
- low priority and maybe not even wanted but showing the ICS YAMLs in the specific country could lead to the ICS mask with the to-dos shown on top of the configuration fields. Maybe allow to specify default values in the yaml (like regex method or specific data) 

General stuff I noticed while looking at the code:

- why is the first step named user and not country?
- comment above async_step_source is wrong



one time error log:
```
  File "/workspaces/core/config/custom_components/waste_collection_schedule/config_flow.py", line 112, in async_step_args
    ] = cv.string if default is None else cv_map[type(default)]
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: TextSelector.__init__() got an unexpected keyword argument 'type'
```

dan-r:
Thanks for your feedback on these changes!



> I think it would be nice to "move" the wizards to the UI. I thought about a check whether the source or module has a specific method providing values to select.

Interesting - I hadn't spotted the sources with wizards. I'll take a look into that.

> Type checking: what about an optional dictionary containing types of attributes.

Agreed, I'll look into that.

> It might be nice to have to also be able to set vol.Exclusive as we have a lot of sources you can either configure with some kind of ID and with an address (e.g. jumomind_de allows city street house_number or city_id and area_id)

Agreed. I couldn't see a good way to intepret this type of input requirement. It would be a fair bit of initial work but maybe its worth introducing a JSON (or other machine readable) 'definition' of each source?

> low priority and maybe not even wanted but showing the ICS YAMLs in the specific country could lead to the ICS mask with the to-dos shown on top of the configuration fields. Maybe allow to specify default values in the yaml (like regex method or specific data)

Seems reasonable - I didn't go any further with ICS because I wanted to keep it as 'generic' as possible. There were a couple of things like request headers that I'd like to add support for in the UI.

> why is the first step named user and not country?

'user' is the first step the UI setup wizard calls. [See this table](https://developers.home-assistant.io/docs/config_entries_config_flow_handler/#defining-steps). You could have a step called country thats called straight away from the user step though...

> comment above async_step_source is wrong

Fixed :)

muchcodesuchwow:
I never used this integration, because the setup process looked compliacted to me. I am a newbie to this integration and gave this PR a try.

The waste collection in my city in germany is provided by "app_abfallplus_de". The following problem does not only arrise for my city, i tried several others with "app_abfallplus_de".

In the last step of the setup a lot of fields are shown. only a few are labled. I dont know, what to enter there. Somewhere my street needs to be entered, but i dont know where. If i press enter, it tells me, that not all mandatory fields are filled, but these fields are not marked. 
grafik

I want to note, that if i use "Generic" (for ICS) in the next step i am asked for a source, but only ICS is the option. This step can be skipped imo, if "Generic" is selected.
grafik

5ila5:
I made some updates https://github.com/dan-r/hacs_waste_collection_schedule/pull/1 @dan-r what do you think? (fixes @muchcodesuchwow issue as well)

dan-r:
@5ila5 thanks for that - merged it in

custom_components/waste_collection_schedule/config_flow.py
```diff
@@ -0,0 +1,384 @@
...
+class WasteCollectionOptionsFlow(OptionsFlow):
+    """Options flow."""
+
+    def __init__(self, entry) -> None:
+        self._entry = entry
+
+    async def async_step_init(self, options):
+        SCHEMA = vol.Schema(
+            {
+                vol.Optional(CONF_SOURCE_CALENDAR_TITLE): cv.string,
```
ReneNulschDE, line 341: The actual/default value is missing here. Test case: Create a calendar with a custom title. Open the options dialog. Result: custom name is missing in the dialog.

5ila5:
I created a new PR where I added some further stuff (https://github.com/dan-r/hacs_waste_collection_schedule/pull/2)

I think we are at a point where we are nearly feature complete compared to the YAML configuration, so we could probably merge this to the master after some people took a look at the code. We would probably need a bit of documentation but this can be done in the mampfes/master repo. 

https://github.com/dan-r/hacs_waste_collection_schedule/pull/2#issue-2355841304:
> I added a bunch of further configuration options, Not sure if everything I did is a good idea, so an extra set of eyes and feedback would be appreciated.
> 
>     * allow YAML configuration along GUI configuration (not sure if this implementation is the best idea)
> 
>     * added Sensor configuration to inital config flow and to the options flow
> 
>     * added customization configuration to inital config flow and the options flow
> 
>     * added additional translations support
> 
>     * added the options for sources to specify parameter translations for different languages (PARAM_TRANSLATIONS parameter)
> 
>     * added German translations
> 
>     * added initial support for the static source
> 
>     * fixed options flow not showing calendar title
> 
> 
> ## Limitations
> 
>     * You cannot define sensors in the YAML configuration that access the GUI sources (or the other way around)
> 
>     * You cannot have sensors, that access multiple sources (source_index of YAML configuration) (possible future workaround: the `multiple` wrapper source.
> 
>     * static source implementation is not ideal and `multiple` sources wrapper is currently not supported
> 
> 
> ## possible future improvement
> 
>     * implement support for wizards could solve problems with `static` and `multiple` source

oerix:
Any update on this please? thanks

dan-r:
Thanks @5ila5 for your work on this, I've merged the changes in. Sorry for not getting round to it sooner! Functionally I think it is ready to go now.

The only part that I think could be a bit confusing to new users is the sensors step (last one). Is it worth adding a 'do you want any sensors' step with potentially one or two pre-made 'examples'? Personally I don't (yet) really use the sensors and just like having the entries added to my calendar. 

This could always be done as future work and is just a thought.

Dan

5ila5:
@mampfes do you want to take a look before merging this?

mampfes (approved):
Very good job. Thanks for all the work.

Whizzykid:
How do I utilise this UI feature? I tired downloading the most recent update and the master download, but I get a message to say the config can only be done in YAML. Am I missing something?

5ila5:
It should just be configurable like any other integration in the HA GUI. It sounds like you do not have the master version (HACS might have done something weird?). It will be part of a release soon.

PR #2075, opened 2024-05-09
Addition of wast collection schedule for Landratsamt Regensburg (AWIDO.de)

GernotAlthammer:
Added `lra-regensburg` for the Landratsamt Regensburg in awido_de


5ila5:
This does not work like this. This section of the md file generated automatically. You need to add it to the SERVICE_MAP variable in [custom_components/waste_collection_schedule/waste_collection_schedule/source/awido_de.py](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/source/awido_de.py) and run the update_docu_links.py script

PR #2097, opened 2024-05-19
Introduce VS Code debug config

tathamoddie:
This allows you to be editing a source script in VS Code, then just press F5 to run it through the test script with the debugger attached.

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,57 @@
+{
+    // Use IntelliSense to learn about possible attributes.
+    // Hover to view descriptions of existing attributes.
+    // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
+    "version": "0.2.0",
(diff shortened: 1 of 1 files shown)


5ila5:
I generally like this, but I would probably add the -t option as debugging should probably print the whole stacktrace for better debugging.

I'd also like a third option debugging ics (yaml) source and just change the `-s` to a `-y`

tathamoddie:
> I generally like this, but I would probably add the -t option as debugging should probably print the whole stacktrace for better debugging.
> 
> I'd also like a third option debugging ics (yaml) source and just change the `-s` to a `-y`

Both updates made. 👍

5ila5:
Nothing changed, did you push the right thing?

tathamoddie:
Oops, sorry.

Fixed.

5ila5:
Thanks for your contribution

PR #2138, opened 2024-06-11
Add Pronatura Bydgoszcz

jestempablo:
Pronatura Bydgoszcz serves a part of Bydgoszcz, Poland


5ila5:
Please provide a md file and run the updatedoculinks.py script like described in the [Source Contribution Guide](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md)

custom_components/waste_collection_schedule/waste_collection_schedule/source/pronatura_bydgoszcz.py
```diff
@@ -0,0 +1,138 @@
...
+        self._street_name = street_name.upper()
+        self._street_number = str(street_number).upper()
+        self._lang = lang if lang in ["pl", "en"] else "pl"  # Fallback to 'pl' if not 'pl' or 'en'
+
+    def fetch(self):
+        try:
+            return self.get_data()
+        except Exception as e:
+            _LOGGER.error(f"fetch failed for source {TITLE}: {e}")
+            return []
```
5ila5, line 59: Please do not catch any errors, they will be caught automatically

custom_components/waste_collection_schedule/waste_collection_schedule/source/pronatura_bydgoszcz.py
```diff
@@ -0,0 +1,138 @@
...
+                        )
+            # Add recykling entries
+            for day in recykling_days:
+                entries.append(
+                    Collection(
+                        datetime.date(year, month, day),
+                        name_map["papier"],
+                        "mdi:recycle",
+                    )
+                )
```
5ila5, line 120: Is there a reason you are mapping paper plastic and glass to recycling instead of showing all 3. Are they always at the same date? For all addresses?

custom_components/waste_collection_schedule/waste_collection_schedule/source/pronatura_bydgoszcz.py
```diff
@@ -0,0 +1,138 @@
+import datetime
+import json
+import logging
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Bydgoszcz Garbage Collection"
+DESCRIPTION = "Source for Bydgoszcz city garbage collection by Pronatura"
+URL = "https://zs5cv4ng75.execute-api.eu-central-1.amazonaws.com/prod"
```
5ila5, line 9: URL is used by the updatedoculinks.py script and will be shown in the README.md use the website URL here instead and rename this variable to something like API_URL

custom_components/waste_collection_schedule/waste_collection_schedule/source/pronatura_bydgoszcz.py
jestempablo: i will revert to separate because im not sure its all addresses, and that merge is not really necessary

jestempablo:
@5ila5 thanks for the comments! i updated PR now to remove unnecessary features (like lang and combined schedules) and updated the docs as per contribution guidelines

5ila5:
Thanks for your contribution. Just did some minor clean up

- rename to match URL (add _pl)
- removed unnecessary method get_data
- NAME_MAP[waste_type] and ICON_MAP[waste_type] with get equivalents to prevent failour on name change
- modified md file to be more in line with other sources"

Issue #1705, opened 2024-01-21
[Bug]: 1.44.0 -> 1.45.0 update hacs not starting anymore


CedricBojoly:
### I Have A Problem With:

The integration in general

### What's Your Problem

if i try to install shapely manually i get an numpy error
So this version is not working for me

### Logs

```Shell
homeassistant  | 2024-01-21 12:47:29.564 ERROR (SyncWorker_5) [homeassistant.util.package] Unable to install package shapely: error: subprocess-exited-with-error
homeassistant  |
homeassistant  |   × pip subprocess to install build dependencies did not run successfully.
homeassistant  |   │ exit code: 1
homeassistant  |   ╰─> [300 lines of output]
homeassistant  |       Collecting Cython
homeassistant  |         Using cached Cython-3.0.8-cp311-cp311-musllinux_1_1_aarch64.whl (3.5 MB)
homeassistant  |       Collecting oldest-supported-numpy
homeassistant  |         Using cached oldest_supported_numpy-2023.12.21-py3-none-any.whl (4.9 kB)
homeassistant  |       Collecting setuptools>=61.0.0
homeassistant  |         Using cached setuptools-69.0.3-py3-none-any.whl (819 kB)
homeassistant  |       Collecting numpy==1.23.2
homeassistant  |         Using cached numpy-1.23.2.tar.gz (10.7 MB)
homeassistant  |         Installing build dependencies: started
homeassistant  |         Installing build dependencies: finished with status 'done'
homeassistant  |         Getting requirements to build wheel: started
homeassistant  |         Getting requirements to build wheel: finished with status 'done'
homeassistant  |         Preparing metadata (pyproject.toml): started
homeassistant  |         Preparing metadata (pyproject.toml): finished with status 'done'
homeassistant  |       Building wheels for collected packages: numpy
homeassistant  |         Building wheel for numpy (pyproject.toml): started
homeassistant  |         Building wheel for numpy (pyproject.toml): finished with status 'error'
homeassistant  |         error: subprocess-exited-with-error
homeassistant  |
homeassistant  |         × Building wheel for numpy (pyproject.toml) did not run successfully.
homeassistant  |         │ exit code: 1
homeassistant  |         ╰─> [272 lines of output]
homeassistant  |             setup.py:71: RuntimeWarning: NumPy 1.23.2 may not yet support Python 3.11.
homeassistant  |               warnings.warn(
homeassistant  |             Running from numpy source directory.
[...]

PR #2139, opened 2024-06-12
Add danish favrskovforsyning_dk source

danielklejnstrup:


danielklejnstrup:
daniel@daniel-desktop:~/wcs/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/test$ ./test_sources.py -s favrskovforsyning_dk -i -l
Testing source favrskovforsyning_dk ...
  found 2 entries for test
    2024-06-24 : Rest-/madaffald [None]
    2024-06-27 : Plast/mad- og drikkekartoner og glas/metal [None]

danielklejnstrup:
Great, thanks - I'll base my future work on your commit.
Am I free to use other python packages?
Affaldonline could be implemented as a solution for several municipalities, but it would require scraping an autogenerated pdf, which would require pdfquery or a similar package.
In return, this seems to be the best solution, since it seems more standardized between municipalities.

5ila5:
We generally avoided parsing PDFs. But if it works reliably for specific sources, we could think about it. Adding additional dependency caused some problems in the past (#1705)

PR #2141, opened 2024-06-14
Add source: frankston_vic_gov_au

b-goold:


custom_components/waste_collection_schedule/waste_collection_schedule/source/frankston_vic_gov_au.py
```diff
@@ -0,0 +1,136 @@
+import requests
+import json
+from shapely.geometry import Point, MultiPolygon, Polygon
```
5ila5, line 3: shapely currently is not a dependency. We tried adding shapely as dependency in the past and some people had problems running this integration at all (#1705)

b-goold: Ah bugger. Thanks, I'll have a go at replicating the behaviour without Shapely.

b-goold: Resolved with https://github.com/mampfes/hacs_waste_collection_schedule/pull/2141/commits/0205c5dfa4aea9aefbd481c73bfd9cc205a630f6

5ila5:
Thanks for your contribution.

Did some reformatting based on (.pre-commit-config.yaml)

PR #2146, opened 2024-06-15
awn_de: Skip cert verification on POST requests as well

scoop:
Similar to #1333, I've seen SSL cert verification issues on the latest HassOS. After disabling verification to the `POST` requests in addition to the `GET` request that was done in the aforementioned issue, the component is successfully downloading ICS data for me again.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awn_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awn_de.py
@@ -89,6 +89,7 @@ class Source:
         r = session.post(
             SERVLET,
             data=args,
+            verify=False,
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution.

I'm not the biggest fan of disabling SSL verification, but if it's not working properly on some machines it's probably fine to disable it as it's not a very critical application

scoop:
> I'm not the biggest fan of disabling SSL verification, but if it's not working properly on some machines it's probably fine to disable it as it's not a very critical application

Agreed, but since it was already disabled on one of the requests, it's barely worsening it by disabling it on the others 😅

Thanks for merging!

PR #2173, opened 2024-06-21
Add AWG Bassum / LK Diepholz

engels0n:
Waste Collection configuration for AWG Bassum / LK Diepholz


5ila5:
Thanks for your contribution.

You actually need to write a YAML file, which then automatically generates the md file. (this allows easy testing with our test script and automatically adds links in the README.md, INFO.md and ICS.md) I just added this.

I Also added a how-to section to explain gin to replace the year with `{%Y}` for multi-year usage (always fetches the current year and tries the next year in December)

engels0n:
Thank you @5ila5 

I didn‘t know that :)

PR #2188, opened 2024-06-24
Gather cumulative entries, no duplicates.

prudy:
This fixes the problem of using only last collection list, instead using all downloaded entries.
It also does not allow duplicated entries, that I noticed for my location.


5ila5:
Thanks for your contribution

PR #2200, opened 2024-06-29
Update gedling.gov.uk to use alternative collection search tool

jamesmacwhite:
Minor edit to the docs. There is now an enhanced collection search tool, that directly provides the correct schedules and links to the appropriate calendar, without having to use the Gedling's search. The collection search is a proxy to the original, but enhances the data returned into a better structure for displaying on a bespoke website.


5ila5:
Thanks for your contribution

PR #2214, opened 2024-07-02
Add nvaa_se source

williamhogman:
This adds the NVAA source serving Norrtalje municipality in Sweden.


README.md
```diff
@@ -985,7 +985,7 @@ Waste collection schedules in the following formats and countries are supported.
 - [Waldshut](/doc/source/app_abfallplus_de.md) / Abfall+ App: unterallgaeu
 - [WBO Wirtschaftsbetriebe Oberhausen](/doc/source/abfallnavi_de.md) / wbo-online.de
 - [Wegberg (MyMuell App)](/doc/source/jumomind_de.md) / mymuell.de
-- [Wermelskirchen](/doc/source/wermelskirchen_de.md) / wermelskirchen.de
+- [Wermelskirchen (Service Down)](/doc/source/wermelskirchen_de.md) / wermelskirchen.de
```
williamhogman, line 988: Unsure why this happens but it is consistently down for me

custom_components/waste_collection_schedule/waste_collection_schedule/source/nvaa_se.py
```diff
@@ -0,0 +1,121 @@
+from datetime import datetime
+
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "Norrtalje Vatten & Avfall"
+DESCRIPTION = "Source for Norrtalje Vatten & Avfall waste collection services, Sweden."
+URL = "https://sjalvservice.nvaa.se/api/v1/integration/getNextGarbageCollection"
```
5ila5, line 8: This parameter will is shown in the README.md and should not be the API URL but the domain of the service provider (probably https://sjalvservice.nvaa.se/?), you can rename this parameter to something like API_URL for use within this script

williamhogman: Now fixed

README.md
5ila5: This is expected we renamed this source, but it seems like the update script was not run on the version you forked. It is already like this on the current master

5ila5:
Thanks for your contribution.

Just did some reformatting (based on .pre-commit-config.yaml)

PR #2217, opened 2024-07-03
Fix stevenage gov uk

### LINKED ISSUE #1997: Stevenage UK has changed service
### I Have A Problem With:

A specific source

### What's Your Problem

API for Stevenage UK has moved to a different service: https://myaccount.stevenage.gov.uk/en/service/my_bin_collection_schedule

This is issue is more of a notification. If I have time, I'll see if I can fix it.

### Source (if relevant)

stevenage_gov_uk

### Logs

_No response_

### Relevant Configuration

_No response_

### Checklist Source Error

### Checklist Sensor Error

### Required

adam-prickett:
Fixes #1997 

Stevenage has moved to Granicus and so this fix breaks existing implementations, instead requiring a UPRN over a postcode and property number.

The code was largely borrowed from highland_gov_uk

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/stevenage_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/stevenage_gov_uk.py
@@ -1,108 +1,104 @@
-import json
-import requests
-import urllib3
-from datetime import datetime
+import datetime
+import time
 
+import requests
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
-# With verify=True the POST fails due to a SSLCertVerificationError.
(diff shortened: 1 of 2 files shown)


5ila5:
Thanks for your contribution.

I added a link to findmyaddress.co.uk in the md file. And did some reformatting and ran the ./update_docu_links.py script

jamesmyatt:
If highland_gov_uk also uses Granicus, is it worth trying to extract a common structure?

adam-prickett:
> Thank you for this.
> 
> If highland_gov_uk also uses Granicus, is it worth trying to extract a common structure?

Potentially, however Granicus appears to be a frontend for different services behind the scenes and so this implementation required an extra network call to obtain a token for the waste management service

PR #2232, opened 2024-07-08
Update to Hawkesbury NSW GOV AU Integration 

zzznz27:
This change fixes the integration to the Hawkesbury NSW Gov API. 

Changes include: 

- Fix to handle inconsistency between dates and datetimes in bin _week<number> field
- Added Postcode and State to the get request parameters and as a argument for the integration
- Added fix to handle missing organic schedule frequency field when organics are not collected at address


5ila5:
Thanks for your contribution.

I did some reformatting (based on .pre-commit-config.yaml) and changed the request line to use the params keyword to be better readable

PR #2235, opened 2024-07-08
Broxbourne - Correcting source date format handling

SimonLeigh:
Unfortunately, when I originally tested this the active month was "May" which only has three letters. It turns out that the source uses 3 letter month abbreviations as it broke as soon as the month rolled to June.

This fixes the parsing to the correct date format.


5ila5:
Thanks for your contribution

PR #2238, opened 2024-07-10
Add support for Kleszczewo and Kostrzyn commune in Poland

sematuszewski:
Add support for Kleszczewo and Kostrzyn commune in Poland


custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
```diff
@@ -0,0 +1,123 @@
...
+                for cell_index, cell in enumerate(row.findall(".//td")):
+                    if cell_index > 0 and isinstance(cell.text, str):
+                        for day in cell.text.split(","):
+                            entries.append(
+                                Collection(
+                                    datetime.date(year, row_index - 1, int(day)),
+                                    NAME_MAP[cell_index],
+                                    ICON_MAP[cell_index],
+                                )
+                            )
```
5ila5, line 121: I'd highly prefer to prevent this deep nesting by inverting your if condition and continue (if row_index == 0): continue

Is Hard-coding the index to names map a good idea? Wouldn't it be better to just read the name from the th's of the first row?

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
```diff
@@ -0,0 +1,123 @@
...
+        r.raise_for_status()
+        number_id = 0
+        numbers = json.loads(r.text)
+        for item in numbers:
+            if item["value"] == self._street_number:
+                number_id = item["id"]
+        if number_id == 0:
+            raise Exception("number not found")
+
+        r = requests.get(f"{api_url}/reports?type=html&id={number_id}")
```
5ila5, line 97: I'd prefer to use the params argument (`r = requests.get(f"{api_url}/reports", params={"type": "html","id": number_id})`

5ila5:
bandit: `/home/silas/Documents/coding/python/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py:114: B314[bandit]: MEDIUM: Using xml.etree.ElementTree.fromstring to parse untrusted XML data is known to be vulnerable to XML attacks. Replace xml.etree.ElementTree.fromstring with its defusedxml equivalent function or make sure defusedxml.defuse_stdlib() is called` It's probably not the biggest problem, as it comes from a somewhat trusted source, but I think it's a better idea to use BeatuifulSoup

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
sematuszewski: I agree with you:)
I fixed it in the latest commit.

sematuszewski:
> bandit: `/home/silas/Documents/coding/python/hacs_waste_collection_schedule/custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py:114: B314[bandit]: MEDIUM: Using xml.etree.ElementTree.fromstring to parse untrusted XML data is known to be vulnerable to XML attacks. Replace xml.etree.ElementTree.fromstring with its defusedxml equivalent function or make sure defusedxml.defuse_stdlib() is called` It's probably not the biggest problem, as it comes from a somewhat trusted source, but I think it's a better idea to use BeatuifulSoup


You're right, I changed the library to defusedxml.ElementTree.

PR #2261, opened 2024-07-14
Add Boden.se source provider

morotsgurka:
This adds support for [Bodens municipality](https://www.boden.se/) in northen Sweden.
The implementation is a simple adjustment on the [ssam source provider](https://github.com/morotsgurka/hacs_waste_collection_schedule/blob/master/doc/source/ssam_se.md) since our municipalities seem to use the same system and formatting.

The test data addresses are the Municipality City Hall and the local secondary/upper secondary school. 

This is the output of the test script:
```
Testing source boden_se ...
  found 3 entries for Bodens Kommun
    2024-07-29 : Deponi Kärl 140.0l
    2024-07-15 : Matavfall, Kärl 140.0l
    2024-07-15 : Brännbart, Kärl 660.0l
  found 7 entries for Gymnasiet
    2024-07-16 : Brännbart, Container 0m³
    2024-07-16 : Matavfall, Kärl 140.0l
    2024-07-16 : Brännbart, Kärl 180.0l
    2024-07-29 : Deponi Kärl 180.0l
    2024-07-16 : Brännbart, Kärl 240.0l
    2024-07-29 : Deponi Kärl 240.0l
    2024-07-16 : Brännbart, Kärl 660.0l
```


custom_components/waste_collection_schedule/waste_collection_schedule/source/boden_se.py
```diff
@@ -0,0 +1,106 @@
...
+                if item["WasteType"] == "Trädgårdsavfall":
+                    icon = "mdi:leaf"
+            found = 0
+            for x in entries:
+                if x.date == next_pickup_date and x.type == waste_type:
+                    found = 1
+            if found == 0:
+                entries.append(
+                    Collection(date=next_pickup_date, t=waste_type, icon=icon)
+                )
```
5ila5, line 105: I'd prefer 
```python
found = any(x.date == next_pickup_date and x.type == waste_type for x in entries)
if not found:
    entries.append(...)
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/boden_se.py
```diff
@@ -0,0 +1,106 @@
...
+        )
+
+        address_data = json.loads(response.text)
+        address = None
+
+        if address_data and len(address_data) > 0:
+            address = address_data["Buildings"][0]
+
+        if not address:
+            return []
```
5ila5, line 38: please raise an Exception as this will be shown to the user. You implementation will just return an empty calendar without any further information
e.g.:
```python
if not address:
    raise Exception("Address not found")
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/boden_se.py
```diff
@@ -0,0 +1,106 @@
...
+            else:
+                waste_type = (
+                    item["WasteType"]
+                    + " "
+                    + item["BinType"]["ContainerType"]
+                    + " "
+                    + str(item["BinType"]["Size"])
+                    + item["BinType"]["Unit"]
+                )
+                icon = "mdi:trash-can"
```
5ila5, line 95: Why are you doing the exact same thing 3 Times? The else case does exacactly the same as the if's above.
If you want to set custom icons use a ICON_MAP dictionary in the top of the file and get the icon with `ICON_MAP.get(item["WasteType"])`

5ila5 (changes requested):

morotsgurka:
Hi @5ila5 
You're right. I've made the changes suggested and also added more exceptions to give a more clear error message on why a specific address would fail. I also added some comments to further clarify and i refactored the section about adding prefixes to the WasteType and the icons.

These changes should probably be done in the [SSAM_SE source](https://github.com/morotsgurka/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/source/ssam_se.py) i took the implementation from? Or maybe make some kind of reuseable provider where you can specify the endpoint? Since the only difference between SSAM_SE and Boden_SE is the endpoint where we get the address and the schedule data.

5ila5:
I'm generally happy with the code. I think it would be best to write one source for all service providers using this api.

I think they either use <https://edpevent.se/mobile/> (**edpmobile**.boden.se) or <https://edpevent.se/future-avfall/> (**edpfuture**.ssam.se). There are probably more service providers/municipalities using this backend but I could not find a good way to find them.

I only found [uppsalavatten.se](https://www.uppsalavatten.se/sjalvservice/hamtningar-och-berakningar/dag-for-sophamtning-och-slamtomning)

morotsgurka:
Yes i agree, but i do not know what the best approach to that would be?

A simple way would be adding a "edpevent.se" source with two urls + the address as parameters. So the user would provide the URL for the GetWastePickupSchedule, the url for SearchAddress and the address. These could then be piped into a single source. Problem with this it's more setup and could be hard for the user? The user would need to find the url themselves. Some already found urls could be added to the markdown file for the source though.

Another way would be to create a reusable module for the framework, so that for each new source that needs to be created you can just import the module and have a super simple config per new provider using edpevent?
Something like:
```python
from waste_collection_schedule import Edpevent

class Source:
    def __init__(self, street_address):
        self._street_address = street_address
     def fetch(self):
         return GetEventsFromEdpevent(SearchAddressURL, GetWastePickupScheduleURL, self.street_address)

```
This would make it easier to add new sources that use edpevent endpoints.

Some endpoints i found:
Skelleftea.se: https://wwwtk2.skelleftea.se/FutureWeb/SimpleWastePickup/GetWastePickupSchedule
Boden.se: https://edpmobile.boden.se/FutureWeb/SimpleWastePickup/GetWastePickupSchedule
SSAM.se: https://edpfuture.ssam.se/FutureWeb/SimpleWastePickup/GetWastePickupSchedule
Uppsalavatten.se: https://futureweb.uppsalavatten.se/Uppsala/FutureWeb/SimpleWastePickup/GetWastePickupSchedule

5ila5:
I'd do just a single source to prevent any additional overhead when adding a new service provider. You can still add them to the README.md by providing a EXTRA_INFO variable in the file. You can even select the default_params which will be autofilled when selecting one of these aliases in the GUI configuration flow.

so I would do something like this:
<details><summary>Example Code (click to expand)</summary>

```python
SERVICE_PROVIDERS = {
    "skelleftea": {
        "title": "Skellefteå",
        "url": "https://skelleftea.se",
        "api_url": "https://wwwtk2.skelleftea.se/FutureWeb/SimpleWastePickup",
    },
    "boden": {
        "title": "Boden",
        "url": "https://boden.se",
        "api_url": "https://edpmobile.boden.se/FutureWeb/SimpleWastePickup",
    },
    "ssam": {
        "title": "SSAM Södra Smalånds Avfall & Miljö",
        "url": "https://ssam.se",
        "api_url": "https://edpfuture.ssam.se/FutureWeb/SimpleWastePickup",
    },
    "uppsalavatten": {
        "title": "Uppsala Vatten",
        "url": "https://uppsalavatten.se",
        "api_url": "https://futureweb.uppsalavatten.se/Uppsala/FutureWeb/SimpleWastePickup",
    },
}

EXTRA_INFO = [{"title": data["title"], "url": data["url"], "default_params": {"service_provider": provider}} for provider, data in SERVICE_PROVIDERS.items()]

# This is based on the ssam_se source, just changed to work with boden.se
class Source:
    def __init__(self, street_address:str, service_provider:str|None=None, url:str|None=None):
        self._street_address = street_address
        self._url = url
        if url is None:
            if service_provider is None:
                raise ValueError("You must provide either a service provider or a url")
            self._url = SERVICE_PROVIDERS.get(service_provider.lower(), {}).get("api_url")
        if self._url is None:
            raise ValueError(f"Unknown service provider: {service_provider}, use one of {[x for x in SERVICE_PROVIDERS.keys()]}")
        if self._url.endswith("/"):
            self._url = self._url[:-1]
```

</details> 


You could then add a simple update function in the update_docu_links.py script, like:

```python
def update_edpevent_se(modules: dict[str, ModuleType]):
    module = modules.get("edpevent_se")
    if not module:
        print("edpevent_se not found")
        return
    services = getattr(module, "SERVICE_PROVIDERS", {})

    str = ""
    for provider, data in sorted(services.items()):
        str += f'- `{provider}`: {data["title"]}\n'

    _patch_file("doc/source/edpevent_se.md", "service", str)
```

and call this function where the others are called as well (l 226 - 230)

If you then put 
```
<!--Begin of service section-->
<!--End of service section-->
```
somewhere in your md file, it will automatically be filled with provider: title list 

Feel free to suggest another approach or customize any code of this answer

morotsgurka:
I have now made a new pull request #2281 with the proposed changes to merge Boden, Skellefteå, SSAM and Uppsala Vatten into a single source with multiple providers. If that looks good we can then close/remove this pull request for adding just boden.se

PR #2278, opened 2024-07-15
Add new attribute "abfall"

funknerGER:
add a new a attribute "abfall" to select specific waste types. Currently all types would be collected. If the attribute is missing, all types will be collect.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/c_trace_de.py
@@ -160,7 +160,7 @@ BASE_URL = "https://{subdomain}.c-trace.de"
 
 class Source:
     def __init__(
-        self, strasse, hausnummer, gemeinde="", ort="", ortsteil="", service=None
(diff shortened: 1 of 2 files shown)


funknerGER:
the new url works with this parameter "abfall=1|2|5|"

In configuration file I added this parameter:

    ...
    hausnummer: 3
    abfall: 1|2|5|
    ...

5ila5:
I'm generally not the biggest fan of limiting the output on source level instead of hiding them with the customize section, but it can make sense if there are a lot of potentially not needed bin types. If you want to add this option, please also add it to the documentation of that source (md file). Please also remove your commented out line completely.

funknerGER:
I´m also not a big fan of limitation on source level, but I have in my waste collection three types of "Restmüll". This three types have identical names in the source level. So this is the only way to collect the right waste type in my case. 

> I'm generally not the biggest fan of limiting the output on source level instead of hiding them with the customize section, but it can make sense if there are a lot of potentially not needed bin types. If you want to add this option, please also add it to the documentation of that source (md file). Please also remove your commented out line completely.

5ila5:
Thanks for your contribution

Issue #2003, opened 2024-04-21
[Bug]: Manchester UK changed their setup? No longer working


majoracehole:
### I Have A Problem With:

A specific source

### What's Your Problem

Its looks like since this week Manchester is no longer working - I think they have changed their system. I first assumed it was my system but now not so sure. Trying to start from scratch I do think their webpage looks a little different from what I recall and unless I am being daft I cannot find my UPRN code anymore. Unfortunately to dig any deeper is beyond me!

### Source (if relevant)

manchester_uk

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

ReneNulschDE:
Analysis of the web flow:

Required attribute: uprn (sample: 77003491)
All requests after 1. need the Authorization-token. A new token is in the response headers and required for for follow-up calls.

1. HTTP-Get to get an auth-token - Request: `https://manchester.form.uk.empro.verintcloudservices.com/api/citizen?archived=Y&preview=false&locale=en`

2. Read Response-Header: `Authorization`

3. Post-Request: `https://manchester.form.uk.empro.verintcloudservices.com/api/custom?action=bin_checker-get_bin_col_info&actionedby=_KDF_custom&loadform=true&access=citizen&locale=en`
    Request-Header: 
   ```
      Accept: application/json, text/javascript, */*; q=0.01
      Content-type: application/json
      Referer: https://manchester.portal.uk.empro.verintcloudservices.com/
      Authorization: "Token from step 2"
   ```
    Request-Body:
   ```
   {"name":"sr_bin_coll_day_checker","data":{"uprn":"77003491","nextCollectionFromDate":"2024-04-21","nextCollectionToDate":"2025-04-20"},"email":"","caseid":"","xref":"","xref1":"","xref2":""}
   ```

[...]

PR #2295, opened 2024-07-17
Fix Manchester_UK - New service endpoint

ReneNulschDE:
fixes: #2003 

- New service endpoint.

Test cases:
```
Testing source manchester_uk ...
  found 130 entries for domestic #1
  found 130 entries for domestic #2
  found 130 entries for domestic #3
  found 208 entries for domestic #4
  found 130 entries for domestic #5
```


5ila5:
LGTM, thanks

Issue #1977, opened 2024-04-13
[Source Request]: Brent, UK


hirenshah:
### Municipality / Region

Brent, UK

### Collection Calendar Webpage

https://recyclingservices.brent.gov.uk/waste

### Example Address

25 Shaftesbury Avenue, HA3 0QU

### Collection Data Format

As ICS/ICAL file, As html on a webpage

### Additional Information

Once at the mentioned webpage, you enter a Post Code (eg HA30QU) then select an address (eg 25 Shaftesbury Avenue).  This takes you to a page like:
https://recyclingservices.brent.gov.uk/waste/2038877

I'm assuming the ID at the end is unique to the Post Code and Address that has been selected, so that ID can be used going forward.

markvp:
This council is already supported via the generic **ICS** integration. See documentation: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/brent_gov_uk.md

Closing as already supported — feel free to reopen if you hit a specific issue with this source.

PR #2304, opened 2024-07-18
Add Brent Council ICS source

johnnieblows:
Adding Brent Council as an ICS source

#1977


5ila5:
Thanks for your contribution

PR #2334, opened 2024-07-25
Fix lbbd.gov.uk

MrGench:
Fixes a problem where if you paid for garden waste to be collected the source would fail. Now works for either situation.


5ila5:
Thanks for the fix, I changed the logic so that e.g. a rename will not stop the source from working

MrGench:
Sounds good, thanks.

PR #2336, opened 2024-07-25
Double quotes caused error with f-string

morotsgurka:
This fixes the error i wrote about in pull request #2335 
The problem in ahk_heidekreis_de caused the update_docu_links.py to fail

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ahk_heidekreis_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ahk_heidekreis_de.py
@@ -79,7 +79,7 @@ class Source:
             (
                 item
                 for item in data
-                if f"{item["houseNr"]}{item["houseNrAdd"]}" == self._house_number
+                if f"{item['houseNr']}{item['houseNrAdd']}" == self._house_number
             ),
             None,
         )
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks, this double quotes only works for python 3.12 so it's probably best to not do this

morotsgurka:
Oh okay, i was using 3.10. Why is python-version file ignored by gitignore? Would make it easier with pyenv to use the "correct" version of python

5ila5:
I don't really think there is a "correct" version of python, I think containing some backwards compatibility is probably a good thing, I honestly did not see that when merging, and I do not really test against older version of python. I think people should probably be able to use older (still maintained) versions of python like 3.10 for simple source development (even though HA dropped support for 3.11 in 2024.4)

morotsgurka:
With "correct" i meant a known working version. Everything above 3.9 seems to work right now but the docs never mention anything about which version of python to use. Maybe it could help someone if they use for example 3.9 and don't understand why the script wont work.
Maybe a simple "requirements" section in the contributing docs with quick start like:
- Install python 3.10 or higher
- Install dependencies with pip install -r requirements.txt
- Test a source: python3 custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s edpevent_se -i -t
- Create a fork
- Create branch
etc?

PR #2355, opened 2024-07-29
Initial support for Slam (Sludge) pickup/week handling

morotsgurka:
I got a question if it was possible to also get the sludge pickup date. I looked into it with the Borås Test 3 address. It does not return a NextWastePickup date but it does return a week number in WastePickupFrequency eg "Vecka 27".
There was already an implementation when i merged SSAM into edpevent_se where the NextWastePickup can be returned as "v32 Aug 2024". The previous implementation checked if it contained the letter "v" and then created a date from the string using a month map etc. I built on top of this to also work with a simple "Vecka 27" return value.

Not sure if this is the optimal solution for this? I guess this is a bit hard to solve for everything since different municipalities might write their NextWastePickup in a different way.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
@@ -331,14 +331,32 @@ class Source:
         for item in data["RhServices"]:
             waste_type = ""
             next_pickup = item["NextWastePickup"]
+
+            # When no explicit pickup date is provided (e.g. for sludge/slam),
+            # fall back to the week number given in the frequency, e.g. "Vecka 27".
+            if not next_pickup and "v" in item["WastePickupFrequency"].lower():
+                next_pickup = item["WastePickupFrequency"]
             try:
-                if "v" in next_pickup:
+                if "v" in next_pickup.lower():
                     date_parts = next_pickup.split()
-                    month = MONTH_MAP[date_parts[1]]
-                    date_joined = "-".join([date_parts[0], str(month), date_parts[2]])
-                    next_pickup_date = datetime.strptime(
-                        date_joined, "v%W-%m-%Y"
-                    ).date()
+                    # "Vecka 27": derive a date from the ISO week number.
+                    if date_parts[0].lower() == "vecka":
+                        current_year = datetime.now().year
+                        next_pickup_date = datetime.strptime(
+                            f"{current_year} {date_parts[1]} 1", "%Y %W %w"
+                        ).date()
+                    # "v32 Aug 2024": derive a date from week, month and year.
+                    elif date_parts[1] in MONTH_MAP:
+                        month = MONTH_MAP[date_parts[1]]
+                        date_joined = "-".join(
+                            [date_parts[0], str(month), date_parts[2]]
+                        )
+                        next_pickup_date = datetime.strptime(
+                            date_joined, "v%W-%m-%Y"
+                        ).date()
+                    else:
+                        _LOGGER.warning("Failed to parse pickup date: %s", next_pickup)
+                        continue
                 elif not next_pickup:
                     continue
                 else:


custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
```diff
@@ -192,19 +192,38 @@ def fetch(self):
         response = requests.get(getUrl, params=params, timeout=30)
 
         data = json.loads(response.text)
-
+        #print(json.dumps(data, indent=2))
```
bratanon, line 195:
```suggestion
        #print(json.dumps(data, indent=2))
```
Looks like a debug left over.

custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
```diff
@@ -192,19 +192,38 @@ def fetch(self):
...
         entries = []
         for item in data["RhServices"]:
             waste_type = ""
             next_pickup = item["NextWastePickup"]
+
+            # When no date is provided, check the frequency for week number
+            if(next_pickup == ""):
+                #print(item["WastePickupFrequency"])
```
bratanon, line 203: Debug left over

custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
```diff
@@ -192,19 +192,38 @@ def fetch(self):
...
                     date_parts = next_pickup.split()
-                    month = MONTH_MAP[date_parts[1]]
-                    date_joined = "-".join([date_parts[0], str(month), date_parts[2]])
-                    next_pickup_date = datetime.strptime(
-                        date_joined, "v%W-%m-%Y"
-                    ).date()
+                    # Check if it contains the swedish word for week
+                    if date_parts[0].lower() == "vecka":
+                        # Assemble date from week number
+                        #print("Week number: " + date_parts[1])
```
bratanon, line 212: Same again. Debug left over.

custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
```diff
@@ -192,19 +192,38 @@ def fetch(self):
...
+                        current_year = datetime.now().year
+                        next_pickup_date = datetime.strptime(f"{current_year} {date_parts[1]} 1", "%Y %W %w").date()
+                        #print("Next pickup date: " + str(next_pickup_date))
```
bratanon, line 215: Same again. Debug left over.

bratanon (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/edpevent_se.py
morotsgurka: Yes i left the debug messages but commented them out since I thought they could be of use if this (somewhat hacky version to get the date from week numbers) doesn't work for some provider/municipality. But if it's preferred to not include them then sure

5ila5:
You could put the debug messages in the debug logs: `_LOGGER.debug("DEBUG MESAGE")`

I think cleaning the fetch function a bit might be a good idea here. Moving out some conventions in additional functions.

bbr111 (approved):
Thanks for adding sludge/week-number support! Heads-up: this branch was based on a 2024 copy of edpevent_se.py, so merging it as-is would have reverted a lot of subsequent work on master (the Icons enum migration, the SourceArgument* exceptions, and ~15 additional service providers + test cases). I've rebased your change onto current master and made a couple of small tweaks:

- Removed the leftover commented-out print() debug lines (per @bratanon's review).
- Swapped the generic `raise Exception` on an unparseable frequency for a `_LOGGER.warning(...)` + `continue`, so an unexpected string is logged rather than aborting the whole fetch (matches the existing month-parse fallback).

I wasn't able to live-test the "Vecka NN" path from here — if you can confirm it still resolves correctly against your SSAM sludge address that'd be great. Otherwise this looks good to me.

markvp (approved):
Confirmed the sludge path @bbr111 flagged: I ran the `SSAM - Slambrunn` case live and the Slam/Slambrunn service resolves to a valid date (2027-02-01) alongside the normal bins, so the "Vecka NN" week-number handling works end to end. The normal `SSAM - Home` case fetches fine too.

Thanks @morotsgurka for the addition and @bbr111 for the rebase onto master. Good to merge.

PR #2370, opened 2024-08-02
Add McKinney, TX to recollect (and its service provider Waste Connections as well)

elizhyu:


5ila5:
Thanks for your contribution,

I just run the ./update_docu_links.py script 

Just so you know you only need to modify the yaml all the other files are updated by this script

PR #2380, opened 2024-08-05
Add arun.gov.uk source

whi-tw:
BS4 parser required as Arun doesn't support ICS.

Have to set up a session initially by sending GET requests to two URLs to:
1. Start the session
2. Activate the session

This is quite brittle, but I guess so are most of these form scrapers!


5ila5:
Thanks for your contribution

Did some minor reformatting (.pre-commit-config.yaml)

PR #2432, opened 2024-08-12
Il Rifiutologo (Gruppo Hera) - Italy

fabio-garavini:
Integration of [Il Rifiutologo](https://www.ilrifiutologo.it/casa_rifiutologo), italian waste management system


custom_components/waste_collection_schedule/waste_collection_schedule/source/ilrifiutologo_it.py
```diff
@@ -0,0 +1,108 @@
...
+            relative_path="/getNumeriCivici.php",
+            params="{\"idComune\": " + str(self._comune) + ", \"idIndirizzo\": " + str(self._indirizzo) + "}"
+        )
+
+        for number in numeri_civici.json():
+            if number.get('numeroCivico') == str(self._civico):
+                self._civico = number.get('id')
+                break
+            if number == numeri_civici.json()[-1]:
+                _LOGGER.error("Civico non trovato")
```
5ila5, line 75: The logger message is not necessary, all errors of source scripts are caught and logged as error, so this will result in 2 times the identical message in the logs

custom_components/waste_collection_schedule/waste_collection_schedule/source/ilrifiutologo_it.py
```diff
@@ -0,0 +1,108 @@
...
+                raise Exception("Civico non trovato")
+
+        r = api_get_request(
+            relative_path="/getCalendarioPap.php",
+            params="{\"idComune\": " + str(self._comune) + ", \"idIndirizzo\": " + str(self._indirizzo) + ", \"idCivico\": " + str(self._civico) + ", \"isBusiness\": \"0\", \"date\": \"" + datetime.datetime.now().strftime("%Y-%m-%dT%H:%M:%S") +"\", \"giorniDaMostrare\": 31}"
```
5ila5, line 80: I really don't like this: why don't you pass a dictionary and just put do:

```python
def api_get_request(relative_path, params=None):
    return requests.post(
        url=API_URL,
        data={
            "url": REQUEST_BASE_URL + relative_path,
            "type": "GET",
            "parameters": json.dumps(params) if params else None
        },
        headers=HEADERS
    )
```

and `import json` at the top

fabio-garavini:
Ok, I'm sorry for the mess that I'm making with this pull request 😅
Now it should be fine

custom_components/waste_collection_schedule/waste_collection_schedule/source/ilrifiutologo_it.py
fabio-garavini: Done

custom_components/waste_collection_schedule/waste_collection_schedule/source/ilrifiutologo_it.py
fabio-garavini: 👍

5ila5:
Thanks for your contribution.

I'm working on a redesign of the whole source parameter translations. Can I ping you If I have something I can put in a PR to look at the Italian sources (only 3 so far)?

fabio-garavini:
> Thanks for your contribution.
> 
> I'm working on a redesign of the whole source parameter translations. Can I ping you If I have something I can put in a PR to look at the Italian sources (only 3 so far)?

Sure!

5ila5:
Noticed some issues with this source and fixed it in #2447

PR #2439, opened 2024-08-13
rename ICS source: mein_abfallkalender_online to mein_abfallkalender_online

5ila5:
As they moved their service to the .online top level domain


MaxWinterstein:
Took me waaaaaay to long to find this, so I will drop the error message here for all the fellow search users:

Abfalltermine fehlen? !!! Update iCAL Service erforderlich !!!

tl;dr: replace the `mein-abfallkalender.de` part of the URL. It is no longer `.de`, it is `.online`.

```diff
- mein-abfallkalender.de
+ mein-abfallkalender.online
```

PR #2451, opened 2024-08-13
Add generic ICS for GDA Amstetten (gda.gv.at)

tsparber:
GDA Amstetten provides a ICS for each address in the district, selectable via a website.

The official homepage moved from https://amstetten.umweltverbaende.at to https://gda.gv.at.

I kept the EXTRA_INFO value in `umweltverbaende_at.py` as GDA Amstetten is still part of NÖ Umweltverbände but maintains its own homepage as listed in https://www.umweltverbaende.at/verbaende.


5ila5:
Thanks for your contribution

PR #2462, opened 2024-08-14
South Kesteven UK: fixes (fetch for configured id, date format); support garden waste collections

inconspicuous-name:
Source was fetching for a hard-coded ID: changed to use the configuration value.
Date format was wrong in one of the cases: corrected
Added support for garden waste collections, which have different text (instead of a link like "What goes in the green bin" we get a list of items, the first of which is "Leaves")


5ila5:
Thanks for your contribution

PR #2485, opened 2024-08-17
Update GitHub workflow

tsparber:
Update the version of the github actions used to fix the nodejs deprecation/replacement annotation.
This PR also updates the used python version to 3.12 as it is the version currently used by HomeAssistant.

https://github.blog/changelog/2024-03-07-github-actions-all-actions-will-run-on-node20-instead-of-node16-by-default/

image
image

### DIFF
--- a/.github/workflows/hassfest.yaml
+++ b/.github/workflows/hassfest.yaml
@@ -10,5 +10,5 @@ jobs:
   validate:
     runs-on: "ubuntu-latest"
     steps:
-      - uses: "actions/checkout@v2"
+      - uses: "actions/checkout@v4"
       - uses: home-assistant/actions/hassfest@master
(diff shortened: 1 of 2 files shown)


5ila5:
I used the "old" python version on porpoise to check if the scripts work with python 3.9 as some people do run this python version and cannot run the update_doculinks script if there are any python 3.12 features in any sources

I would like the sources to be at least a few python version backwards compatible and if the tests pass the 3.9 tests they should also pass the 3.12 tests.

What do you think?

tsparber:
Good point, I updated the commit to use python 3.9 as it is (soon) the oldest officially supported python version.

python 3.8 will be EOL in 10/2024.
https://devguide.python.org/versions/

5ila5:
Thanks for your contribution

PR #2492, opened 2024-08-18
Config flow: add howto and custom translations, descriptions

5ila5:
Changes:
- Add custom Exceptions that are treated differently by the config flow to show error messages directly at the faulty argument and even provide alternate suggestions
- Add the ability to translate ICS howtos into different languages (only the english version is shown in the md file)
- config_flow: every source has it's own 
- Added `HOW_TO_GET_ARGUMENTS_DESCRIPTION` parameter to add a howto section to normal sources
- Howto Sections are shown in the config_flow GUI above the input fields
- Add source level translation using the `PARAM_TRANSLATIONS` argument, so every source can decide by itself how the parameters should be translated. The old approach would translate some parameters badly, as arguments such as `district` were used with different meanings.
- Added default translations for parameter not translated by `PARAM_TRANSLATIONS`
- Added the parameter `PARAM_DESCRIPTIONS` to add (translated) descriptions for arguments shown below the input field
- added default param description for `uprn`
- update pytest to test newly added arguments


5ila5:
@mampfes do you want to take a look?
@ReneNulschDE do you want to take a look as you built/maintained a few sources
@fabio-garavini do you want to take a look at the Italian translations

fabio-garavini:
the word `details` is misspelled as `detailes` and `Go to` as `Goto`

custom_components/waste_collection_schedule/config_flow.py
```diff
@@ -283,12 +291,32 @@ class WasteCollectionConfigFlow(ConfigFlow, domain=DOMAIN):  # type: ignore[call
...
         self._sources = self._get_source_list()
         self._options: dict = {}
 
+        async def args_method(args_input):
+            return await self.async_step_args(args_input)
+
+        async def reconfigure_method(args_input):
+            return await self.async_step_reconfigure(args_input)
+
+        for _, sources in self._sources.items():
```
mampfes, line 307: `values()` instead of `items() `? To get rid of `_`

PR #2521, opened 2024-08-21
Update chiltern_gov_uk.py

craigsblackie:
This should fix the source after it was changed.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/chiltern_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/chiltern_gov_uk.py
@@ -5,6 +5,9 @@ import requests
 from datetime import datetime
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your work,
I would really like it to be an extra itouchvision source so it's easy to integrate potential other municipalities using this "new" itouchvision API I know NEWPORT does use this by default, but we could use the old implementation with the same ID

Do you want to give it a try or should I redo the source to a dedicated second itouchvion source?

craigsblackie:
It would make sense if other councils are usig this new endpoint to create a new generic itouchvision source. I'm not sure I have the time to implement this currently, but hopefully this provides a solid base from which to build that. 

Why would they encrypt the parameters?!

5ila5:
Ok I will do the new source

I have no clue why they do this, maybe it's some react library they are using the key's name is REACT_APP_ENCRYPT_PUBLIC_KEY

5ila5:
Closed for #2522

PR #2532, opened 2024-08-22
Added source for Napier City Council NZ

redjungle:
New source for the Napier City Council in New Zealand, including .py, .md changes etc.

image


5ila5:
Your waste Type is always `Collection` as you match the h2 heading which is Collection days instead of the th element of the table

5ila5:
Thanks for your contribution.

Just did some minor reformatting based on .pre-commit-config.yaml

PR #2577, opened 2024-08-27
Update renfrewshire_gov_uk.py to handle multiple bin collections on a single day

JamiePhonic:
Hello! 
I have attempted to update the code to handle locations in Renfrewshire that have multiple bin collections on 1 day (See below image)
I'm not fluent with Python yet, so this represents my best effort. I have tested the code and verified it works as far as `schedule` having the expected data.

image


5ila5:
Looks like you forgot to commit your changes properly. You only changed one test case

JamiePhonic:
Whoops! Should be sorted now.

5ila5:
LGTM, Thanks for your contribution

PR #2579, opened 2024-08-28
Added City of Ryde, Australia

harrytheeskimo:
This uses the same API as Campbelltown City Council and Blacktown City Council (NSW).

Let me know if I've missed any required file updates please.


custom_components/waste_collection_schedule/waste_collection_schedule/source/ryde_nsw_gov_au.py
```diff
@@ -0,0 +1,118 @@
...
+
+        data = json.loads(r.text)
+
+        # Find the ID for our suburb
+        for item in data["Items"]:
+            locationId = item["Id"]
+            break
+
+        if locationId == 0:
+            return []
```
5ila5, line 75: Please raise an Exception here as people using the integration should know what went wrong (Exceptions will show in the logs) instead of just getting an empty calendar

harrytheeskimo: Thanks for the review - I've updated both of these.

5ila5 (approved):

5ila5:
Thanks for your contribution.

Did some reformatting based on .pre-commit-config.yaml and run the update_docu_links.py script

PR #2584, opened 2024-08-29
Add koeniz_ch

Hiekkan:
There is still a small bug. I get duplicated entries from all types.


5ila5:
Thanks for your contribution. I fixed the duplicate bug: this site has two ics Links one for the text and one for the icon I just added a break after finding the first one.

Issue #2643, opened 2024-09-09
[Bug]: source AWR_DE defective


pehebo:
### I Have A Problem With:

A specific source

### What's Your Problem

Since latest update of HACS source awr_de isnt loaded with error:
^ TypeError: can't compare datetime.datetime to datetime.date

Before everything workes fine.

### Source (if relevant)

awr_de

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

9H3SC:
Same here for source: awsh_de

TheDuffman85:
Same issue here with source Abfall.IO / AbfallPlus

Logs:
2024-09-09 10:02:12.716 ERROR (SyncWorker_2) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source Abfall.IO / AbfallPlus:
Traceback (most recent call last):
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 145, in fetch
    entries = self._source.fetch()
              ^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/abfall_io.py", line 177, in fetch
    dates = self._ics.convert(ics_file)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py", line 48, in convert
    events: List[Any] = icalevents.events(
                        ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/icalevents/icalevents.py", line 57, in events
    found_events += parse_events(
                    ^^^^^^^^^^^^^
[...]

PR #2645, opened 2024-09-09
#2643 upgrade icalevents to 0.1.29

dm82m:
icalevents 0.1.28 is broken, setting dependency to 0.1.29 to force an upgrade on all affected instances

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 beautifulsoup4>=4.12.2
 DateTime>=4.9
 icalendar>=4.0.9
-icalevents>=0.1.27
+icalevents>=0.1.29
 python-dateutil>=2.8.2
 pytz>=2021.3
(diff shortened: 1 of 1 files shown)


dm82m:
not sure if this is the correct way to force the upgrade, treat this PR with care

icompi:
Thx, it worked!
I just changed in the manifest.json to "icalevents==0.1.29" and it's finally working again.

saschaludwig:
It would be super nice if this gets merged. All iCal based schedules seem to be broken at this moment.

5ila5:
Thanks for your contribution,
I will release a new version shortly to fix the issue for everyone affected

5ila5:
Just noticed this is not the right spot, these are just the development requirements used to set up the development environment and used for the GitHub actions. 

Actually fixed here: #2652

PR #2652, opened 2024-09-10
fix icalevents version 0.1.28 causing problems

5ila5:
Actually fixes problem addressed here: #2645, Which just changes the development dependencies


dm82m:
That’s why I wrote „not sure if this is the correct way to force the upgrade, treat this PR with care“ 😂

but thanks for this fix here! 🤘

PR #2690, opened 2024-09-16
Update recyclecoach source to include Richmond Hill, Ontario

gregprosser:
I noticed that if I use the recyclecoach integration, pick "Vaughan", and enter my address (which is in Richmond Hill) it appears to work.  So, add Richmond Hill as a listed city.

I'm having trouble testing this in homeassistant - but the test scripts successfully fetch data for recyclecoach after my change, and if I add my address as a test case and ask it to list days, it's returning the correct information.


gregprosser:
I managed to get my custom component to run/work, here are some screenshots of it finding my pickup schedule.
Screenshot 2024-09-16 212355
Screenshot 2024-09-16 212522

5ila5:
Thanks for your contribution

5ila5:
I made some additional improvements to the source in combination with the GUI in #2697

Issue #1479, opened 2023-12-05
[Feature]: Export calendar in icalendar format


schklom:
### I propose a feature for:

The integration in general

### Describe your wanted feature

Ideally, Home-Assistant would provide a way to subscribe from outside to its local calendars. Sadly, it does not.

I want to subscribe from outside (e.g. from Google / Microsoft / Nextcloud / etc) to the waste collection schedule made on Home-Assistant.
A way to do that is by making a file e.g. waste_collection.ics available in /config/www/waste_collection.ics which would be available from outside Home-Assistant via the URL http://your.domain:8123/local/waste_collection.ics (c.f. https://www.home-assistant.io/integrations/http/#hosting-files)

This would solve #1105.

bbr111:
you can use https://github.com/JosephAbbey/ha_calendar_export to export HA Calendar

Issue #1105, opened 2023-07-19
Google Calendar instead/as well as local Calendar


JonReed:
I had a quick look but couldn't see an easy to do this.
Would it be possible to add the events to a google calendar defined in the config?

As a "non-technical partner" of a HA/waste_collection_schedule user
I want to see in my Google calendar when the waste collections are
So that google assistant can also alert me with my daily schedule

markvp:
Thanks for the suggestion, JonReed.

This project is focused on *fetching* waste collection schedules from providers and surfacing them as local Home Assistant calendar and sensor entities. Writing events out to an external service like Google Calendar is outside that scope — it would require OAuth credential handling and ongoing Google API maintenance that doesn't belong here.

The good news is that Home Assistant already has a native [Google Calendar integration](https://www.home-assistant.io/integrations/google/) that supports creating events. You could wire that up with an automation (triggered by this component's sensors or calendar) to push events across to Google Calendar — no changes to this component required.

I'm going to close this as out of scope, but feel free to continue the conversation if you find a HA-native approach that works well for you.

PR #2710, opened 2024-09-19
Added uid to events added to calendar

JonReed:
uid was missing from events being added to the local calendar. This leads to other systems (eg. Google calendar) not being able to import the calendar due to all events having the same uid of 'None'

https://github.com/mampfes/hacs_waste_collection_schedule/issues/1479 and https://github.com/mampfes/hacs_waste_collection_schedule/issues/1105 are blocked without a change such as this

With this change, something like https://github.com/JosephAbbey/ha_calendar_export can be used to export events to other systems


5ila5:
Thanks for your contribution

PR #2713, opened 2024-09-20
Fixes herefordshire_gov_uk

Jab2870:
Change to the website meant it failed to get the dates.


5ila5:
Thanks for your contribution, the test case did not return any data, so I changed it to an address, that shows a schedule on the website and added a custom error message if there is no schedule to be found

Issue #2671, opened 2024-09-12
[Source Request]: City of Saanich, British Columbia, Canada


Uninexus:
### Municipality / Region

Saanich, British Columbia, Canada

### Collection Calendar Webpage

https://www.saanich.ca/EN/main/community/utilities-garbage/garbage-organics-recycling/collection-schedules.html

### Example Address

1565 Craigiewood Court, Victoria, BC, Canada

### Collection Data Format

Dedicated IOS/Android APP

dt215git:
![image](https://github.com/user-attachments/assets/9f8d2be8-3294-4651-9216-f7208ca71e3d)
 The "Subscribe to Calendar" button looks to be using the [ReCollect](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/recollect.md) source.

markvp:
This municipality is already supported via the **ReCollect** platform using the generic **ICS** source. See documentation for how to extract your calendar URL: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/recollect.md

Closing as already supported — feel free to reopen if you hit a specific issue.

Issue #2642, opened 2024-09-08
[Source Request]: Halton Region, Ontario, Canada


PeterHCa:
### Municipality / Region

Halton, Ontario, Canada

### Collection Calendar Webpage

https://www.halton.ca/For-Residents/Recycling-Waste/Online-Waste-Collection-Schedule

### Example Address

2057 Blairholm Ave., Burlington, ON, L7R1V5

### Collection Data Format

As html on a webpage

dt215git:
Does anyone else have difficultly viewing that site?

![image](https://github.com/user-attachments/assets/1d4b3ec5-6c06-4ff3-8037-1e790e8d9c2e)

lit-af:
![image](https://github.com/user-attachments/assets/e8df5eb8-492e-41d8-aa79-d60f62e068ae)
The "Subscribe to Calendar" button looks to be using the [ReCollect](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/recollect.md) source.

mike391:
This has been completed via Recollect:

https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/recollect.md#halton-region-on-canada

markvp:
This municipality is already supported via the **ReCollect** platform using the generic **ICS** source. See documentation for how to extract your calendar URL: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/recollect.md

Closing as already supported — feel free to reopen if you hit a specific issue.

PR #2720, opened 2024-09-21
Additional ReCollect examples

dt215git:
ics.md and recollect.md updates with examples for Saanich, BC (magenta) and Halton Region, ON (blue) addresses given in #2671 and #2642

image


dt215git:
Not sure what's happening with the update_docu_links script, but it keeps removing the additional examples I add to the recollect.md file. Is there another file I need to update?

image

5ila5:
For ICS source: The examples in the md are the test cases of the YAML. So you need to add them to the test cases so they will show up in the example section

dt215git:
Ah, hadn't spotted the yaml file needed updating in two places - think I've resolved all the issue now.

PR #2741, opened 2024-09-25
Add support fo "Renhållningen Kristianstad" in Kristianstad, Sweden

hansbacklund42:
Added support for "Renhållningen Kristianstad" in Sweden to HACS waste collection service


custom_components/waste_collection_schedule/waste_collection_schedule/source/krab_se.py
```diff
@@ -0,0 +1,64 @@
...
+    def fetch(self):
+        params = [{"query": self._street_address}]
+        headers = {"Module": "universal", 
+                   "Accept": "application/json, text/plain, */*", 
+                   "Unit": "dd905ce7-b16d-4422-be36-564169af4035"}
+        response = requests.get(
+            "https://api-universal.appbolaget.se/waste/addresses/search",
+            headers=headers,
+            params=urllib.parse.urlencode(params),
+        )
```
5ila5, line 32: This does not work you cannot `urllib.parse.urlencode` a list, You don't need to encode anything just make your `params` a dict and pass it as `params` the request's library does the converting for you (probably with the same method)

custom_components/waste_collection_schedule/waste_collection_schedule/source/krab_se.py
```diff
@@ -0,0 +1,64 @@
...
+
+        params = [building_id]
+        headers = {"Module": "universal", 
+                   "Accept": "application/json, text/plain, */*", 
+                   "Unit": "dd905ce7-b16d-4422-be36-564169af4035"}
+        response = requests.get(
+            "https://api-universal.appbolaget.se/waste/addresses",
+            headers=headers,
+            params=urllib.parse.urlencode(params),
+        )
```
5ila5, line 50: building_id is not passed as a param at all but as part of the url:

```py
response = requests.get(
    "https://api-universal.appbolaget.se/waste/addresses" +building_id,
    headers=headers
)
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/krab_se.py
```diff
@@ -0,0 +1,64 @@
...
+
+        data = json.loads(response.text)["services"]
```
5ila5, line 52: its `data = json.loads(response.text)["data"]["services"]`

5ila5 (changes requested):
Why did you name the source krab_se? We typically name the source after the website ULR so `renhallningen_kristianstad_se` here.

Did you do any testing before submitting this? The source is fundamentally broken at multiple points. Please test your source using the [test script (test_sources.py)](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md#test-the-new-source-file) before submitting a PR

5ila5 (approved):

5ila5:
Thanks for your contribution

Issue #2569, opened 2024-08-26
[Source Request]: Luxembourg - SICA


Seb3rus:
### Municipality / Region

Mamer / Luxembourg

### Collection Calendar Webpage

https://sicaapp.lu/

### Example Address

rue de l'ecole 
L-8226
Mamer
Luxembourg

### Collection Data Format

As html on a webpage, Dedicated IOS/Android APP, As PDF Download

5ila5:
Added in #2575 

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA

install master with HACS version < 2.0: (3-dot menu -> redownload -> change version to master -> fully restart HA (this will increase update cycle by a lot)

Foxi352:
That is a duplicate. I committed Sica already in #2520.
I think that was even before the Source Request.

5ila5:
@Foxi352 sorry I somehow missed this

Your source seem to support some municipalities not supported by my implementation while your implementation seem to not support Habscht

How id you find the API/JSON? In the App that no longer exists (com.sica_app)? The new app ([lu.cuco.sica](https://play.google.com/store/apps/details?id=lu.cuco.sica)) uses yet another API (https://dashboard.sicaapp.lu/api/api/app/{ENDPOINT})

If this is from the old app I wonder how long the API still works.

Foxi352:
@5ila5 I hooked an MITM proxy between my iPhone and the Internet. I used the APP i had on my iPhone for years. I only realized now, reading your comment, that there was a new APP. There was never an information or message in the old app. And it still works.

As for the municipalities, the old App got them from here: https://sicaapp.lu/wp-json/wp/v2/locations/
Hobscheid and Septfontaines merged 2018 into Habscht. And only villages from previous Septfontaines use services from SICA.
So your API's "Habscht" should be exactly the same data as the old API's "Septfontaines"

But to make a long story short: I guess if your code uses the newer API, you should remove my implementation and keep yours.

Foxi352:
@5ila5 Was very quick:
List of municipalities with sometimes villages as children: https://dashboard.sicaapp.lu/api/api/app/community
List of pickup types: https://dashboard.sicaapp.lu/api/api/app/pickup-type
List of pickup dates: https://dashboard.sicaapp.lu/api/api/app/pickup-date

[...]

PR #2766, opened 2024-09-30
New sica lu api

Foxi352:
As discussed in #2569 i migrated the sica_lu source to the new API.

Output of test run:
```
Testing source sica_lu ...
  found 103 entries for Habscht
  found 104 entries for Steinfort
```

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sica_lu.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sica_lu.py
@@ -1,4 +1,6 @@
 import datetime
+from typing import TypedDict
+
 import requests
 from waste_collection_schedule import Collection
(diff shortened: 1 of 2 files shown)


5ila5:
Thanks for your contribution.

I fixed your type hints:
- any is not a type, but a python built-in function you need to use `typing.Any`-
-  your fetch_json might return a list (collections) or a dict (municipalities)

The easiest way to fix this would be to just use Any (or no type hints) there, but I like that my IDE knows what could be in the dict and does some sanity checks for me, so I added these `TypedDicts`. (maybe run mypy next time to check your types or make sure your IDE does these checks)

and I did some reformatting (black, isort)

You can run all these checks using pre-commit

Foxi352:
Thanks @5ila5 for your fixes, appreciated. I work in cybersecurity and am far from an experienced developer :-)
I am always open to learn something new. Also thanks for the merge.

PR #2770, opened 2024-10-01
Fix typo in value template

marcjay:


5ila5:
Thanks for your contribution

PR #2772, opened 2024-10-01
Updated bir_no.py 

tskulbru:
The waste disposal company have started collecting compost units, so I've added a new icon for this type of waste. 
- Added mapping to new compost type icon


5ila5:
Does the source work for you? It looks like the source requires the python module cssselect to be installed, which is not the case in my test HA install.

I would fix this here: #2773

tskulbru:
I saw that it required modules that werent in requirements, but im not a python guy so i didnt know what to do with it. The module works in my hassio at the very least

5ila5:
I'll close this for #2773. Thanks for your work anyway.

This will be part of the next release or is available now as master (git) version 

<p>
</p>
</details>

Issue #2813, opened 2024-10-10
[Feature]: Unify the icons for different waste-types


dala318:
### I propose a feature for:

The integration in general

### Describe your wanted feature

Every service is now proposing icons for every waste type, and user can then modify the icon per sensor. No problem in itself.

But during service development i see that different services has chosen different proposed icons for same type ow waste. I propose some "standard icons" are defined in a const.py for the most common waste types so that all integrations create similar behavior.

markvp:
Thanks for raising this, DanielV — it's a well-observed inconsistency. You're right that source authors have converged on different MDI icons for the same logical waste categories (e.g. \`mdi:bottle-soda\` vs \`mdi:glass-fragile\` for glass; \`mdi:food-apple\` vs \`mdi:food\` for food waste).

There's already quite a bit of overlapping work in this area worth pulling together:

- **PR #2814** — your own implementation attempt at exactly this, opened the same day. Still open and a natural home for the icon-mapping work itself.
- **#5752** — a broader meta-issue I opened to discuss global type-name normalisation, which explicitly ties the icon question (#2814) and the type-normalisation question (e.g. PR #5569) together.
- **#4272** — a related request to ship default icons/images with the integration.
- **#3249** — related styling customisation (per-type colours).

Rather than fragmenting the discussion across all of these, I'd suggest:

1. Keep the **icon-mapping itself** in PR #2814 — that's where the actual code lives.
2. Use **#5752** for the cross-cutting design discussion (canonical icon set + canonical waste-type names, whether opt-in per source or global).
3. I'll leave #2813 open as the user-facing tracker so it stays visible in search.

A good starting point would be to propose the canonical icon-per-waste-category mapping in #5752 so it can be agreed before #2814 lands.

PR #2814, opened 2024-10-10
Unify usage of Icons

dala318:
Proposal of implementation for #2813


dala318:
For now don't want to spend a whole afternoon to copy-pasting icon usage in all services, specially if it's not something you want implemented or in a different way.

5ila5:
I generally do like this. It does not have a very high priority, but having an ENUM for the waste types is quite nice

Note to also update https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md

I thought about implementing this/or some fallback in the Collection class. The advantage would be that ICS sources would profit from this, as they currently only get the default ICON and you have to set your own if you want to change it.  But I don't think this is a good fit as names vary drastically between sources (languages/bin colors/ waste types / bin size ....). 

I also thought about implementing a default ICON_MAP to import into sources, but it has basically the same problems/limitations as above

dala318:
Yeah, my idea with this was not to automate icon assignment but only to select one for every basic type and then in every service promote selecting from those when the actual translation is done.

dala318:
> Note to also update https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md

Updated templet

dala318:
The `fix_icon_use.sh` script may only be needed once, so could remove it once the majority of sources have been fixed.

script/check_icon_use.sh
```diff
@@ -0,0 +1,498 @@
...
+    "wyre_gov_uk.py"
+    "wyreforestdc_gov_uk.py"
+    "yarra_ranges_vic_gov_au.py"
+    "york_gov_uk.py"
+)
+
+# Build a git-exclude and git-include strings
+grep_exclude=""
+grep_include=""
+for excluded in "${whitelist[@]}"
```
dala318, line 452: This loop (and above list of excluded files) could be replaced by adding a specific string or in each source file that shall be excluded by check and then replace the loop with the following:

```sh
for excluded in $(git grep --name "CUSTOM_ICON_WHITELIST = True" -- $checkpath); do
    base=$(basename $excluded)
    grep_exclude="${grep_exclude} :(exclude)*${base}"
    grep_include="${grep_include} ${checkpath}${base}"
done
```

bbr111:
I'm currently updating validate und pre-commits to ruff.
After that is done, we will go forward here.

PR #2876, opened 2024-10-19
Update merton_gov_uk.py - Add support for garden waste

bulletprooffool:
Tiny change - simply add support for garden waste which was mised by the original implementation (I assume the original developer did not have garden waste?)


5ila5:
This fails if the address does not have garden collections

It looks like the test cases don't work anymore, could you maybe update them and test your changes with them?

Test wtith `python custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s merton_gov_uk -lit`

Issue #2871, opened 2024-10-18
[Source Request]: For Mid and East Antrim Council


marcusoft7:
### Municipality / Region

Mid and East Antrim / United Kingdom

### Collection Calendar Webpage

https://midandeastantrim.gov.uk/resident/waste-recycling/collection-dates/

### Example Address

10 Antrim Road Ballymena

### Collection Data Format

Something different (add to additional information)

### Additional Information

I believe the information is taken out via python script. The collection is the same as Antrim and Newtownabbey Council, i was able to get the data by changing the script from code in the Antrim and Newtownabbey. Would like it added to the integration so it doesnt get copied over when integration updates
[mideastantrim.txt](https://github.com/user-attachments/files/17436960/mideastantrim.txt)

markvp:
This council is already supported via the existing **`midandeastantrim_gov_uk`** source. See documentation: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/midandeastantrim_gov_uk.md

Closing as already supported — feel free to reopen if you hit a specific issue with this source.

PR #2904, opened 2024-10-22
Midandeastantrim gov uk

AHARVEY99:


AHARVEY99:
#This pull request addresses the request for Mid and East Antrim Council as per [Source Request]: #2871.

custom_components/waste_collection_schedule/waste_collection_schedule/source/midandeastantrim_gov_uk.py
```diff
@@ -0,0 +1,118 @@
...
+    "187262293": {"uprn": 187262293},
+}
+
+
+ICON_MAP = {
+    "grey": "mdi:trash-can",
+    "glass": "mdi:bottle-soda",
+    "green": "mdi:leaf",
+    "blue": "mdi:recycle",
+}
```
5ila5, line 22: This seems not to match your collection types, so all collections have icon None

AHARVEY99: Thanks for pointing out, have updated now

PR #2920, opened 2024-10-25
Adding west lothian

alitheg:
Adding parsing from West Lothian Council. They use the forms which several councils use where we need to do some requests and BeautifulSoup work to get data, and then *normally* return ICal compatible data which we can parse. However their ICal data is currently broken (download from the website and gives me a file containing [object Object]), so I've now added parsing from the webpage displayed as a fallback.


alitheg:
Ah, let me get the latest README and info files and update

5ila5:
Thanks for your contribution

swwebb1:
When configuring West Lothian Council I get the following error

The source returned an invalid response: “‘str’ object has not attribute ‘get’”. Please check the provided argument and try again.

From what I can tell, all arguments are correct.

PR #2971, opened 2024-11-02
Fix warszawa19115_pl response schedule new field name

tomek-sienicki:
warszawa19115_pl has recently changed the field name for waste schedule from `harmonogramy` to `harmonogramyZ`.

* Changed the field name + small refactoring
* Added check the field is not null (to avoid an exception)

Test run results:

```
Testing source warszawa19115_pl ...
  found 5 entries for Street Name
  found 5 entries for Geolocation ID
```


5ila5:
Thanks for your contribution

m-wyz:
Thank you!

Issue #2889, opened 2024-10-20
[Bug]: Max retries exceeded


alessandroguerra:
### I Have A Problem With:

Sensor configuration

### What's Your Problem

Max retries exceeded

### Source (if relevant)

na

### Logs

```Shell
The source returned an invalid response: "HTTPSConnectionPool(host='www.ilrifiutologo.it', port=443): Max retries exceeded with url: /ajax/archivio_ilrifiutologo_ajax.php (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: certificate has expired (_ssl.c:1000)')))". Please check the provided arguments and try again.
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: ilrifiutologo_it
      args:
        town: Padova
        street: VIA BERNARDO TREVISAN
        house_number: 8
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

alessandroguerra:
The source returned an invalid response: "Caught blocking call to putrequest with args (<urllib3.connection.HTTPSConnection object at 0x7f6025400440>, 'POST', '/ajax/archivio_ilrifiutologo_ajax.php') inside the event loop by custom integration 'waste_collection_schedule' at custom_components/waste_collection_schedule/waste_collection_schedule/source/ilrifiutologo_it.py, line 170: return requests.post(. (offender: /usr/local/lib/python3.12/site-packages/urllib3/connection.py, line 219: return _HTTPConnection.putrequest(self, method, url, *args, **kwargs)), please report it to the author of the 'waste_collection_schedule' custom integration For developers, please see https://developers.home-assistant.io/docs/asyncio_blocking_operations/#putrequest". Please check the provided arguments and try again.

[...]

PR #2976, opened 2024-11-03
fix: Prevent blocking calls in source __init__ methods (GUI mode)

ReneNulschDE:
When sources have blocking calls in the __init__ method, a warning is shown when adding the source via the UI mode. Potential exceptions on init are overwritten with the blocking warning.

Example: #2889 

This PR moves the initialization into the hass.async_add_executor_job


5ila5:
Thanks for your contribution

PR #2980, opened 2024-11-03
Resolve broken proxy for nottingham city and simplify json processing

TurnrDev:
Not sure exactly when proxy broke, but it's broken.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/nottingham_city_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/nottingham_city_gov_uk.py
@@ -1,5 +1,4 @@
 import datetime
-import json
 
 import requests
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

TurnrDev:
When can we expect the next release?

5ila5:
I release a new version ruffly every month, or when I think it happened enough to justify a release, but you can use the master/git version with HACS which is the current development status. As far as I noticed, it will prompt you to update to the normal releases when there is a new release


install master/git version: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA


<p>
</p>
</details>

TurnrDev:
Thank you, I never knew I could do that!

PR #2983, opened 2024-11-04
Fix static schedule params validation

zolakt:
Validation for the static schedule is wrong. It works when specifying weekdays as strings, but no when specifying weekdays as dictionary


5ila5:
Thanks for your contribution

zolakt:
No problem, glad to help. Thanks for adding the GUI for this, so I can get rid of YAML

PR #2989, opened 2024-11-04
add ics provider for wetzlar de

arbue:
Adding ics provider for wetzlar.de.

Result of:

`./test_sources.py -y wetzlar_de -l`

image


5ila5:
Thanks for your contribution.

I fixed your country argument and run the update_docu_links.py script

PR #3008, opened 2024-11-09
publidata_fr: improve date range parsing and add EXTRA_INFO values

vinzd:
Follow-up for f87fe88c0a with more presets for additional communities handled by publidata

cc @Crazyscorpio88


5ila5:
Thanks for your contribution

PR #3032, opened 2024-11-14
Fix: Blocking call to open sources.json in config_flow initialization

ReneNulschDE:
This is PR tackles the blocking call to load the sources.json in the config_flow initialization. I moved the the _sources and method creation logic out of __init__. 

I'm not sure if this is the best approach. Whenever a new process entry method (other then new, reconfigure) is created, `_async_setup_sources` needs to be called...

Fixes: #3031

### DIFF
--- a/custom_components/waste_collection_schedule/config_flow.py
+++ b/custom_components/waste_collection_schedule/config_flow.py
@@ -298,13 +298,21 @@ class WasteCollectionConfigFlow(ConfigFlow, domain=DOMAIN):  # type: ignore[call
     _country: str | None = None
     _source: str | None = None
 
+    _options: dict = {}
     _sources: dict[str, list[SourceDict]] = {}
     _error_suggestions: dict[str, list[Any]]
 
-    def __init__(self, *args: list, **kwargs: dict):
-        super().__init__(*args, **kwargs)
-        self._sources = self._get_source_list()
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/config_flow.py
```diff
@@ -298,13 +298,21 @@ class WasteCollectionConfigFlow(ConfigFlow, domain=DOMAIN):  # type: ignore[call
     _country: str | None = None
     _source: str | None = None
 
+    _options: dict = {}
     _sources: dict[str, list[SourceDict]] = {}
     _error_suggestions: dict[str, list[Any]]
 
-    def __init__(self, *args: list, **kwargs: dict):
-        super().__init__(*args, **kwargs)
```
5ila5, line 305: Removing the constructor with the super call here is probably a bad idea and was probably an accident?

ReneNulschDE: No it was not, if a base class has a explicit implementation then it would get called, or? But happy to add it again. (Checked all the base classes and they have not an explicit `__init__`

5ila5: Ah, you're right, missed that.

5ila5:
Thanks for your contribution

Issue #3027, opened 2024-11-13
[Bug]: Invalid config for 'waste_collection_schedule' - 'day_offset' is an invalid option for 'waste_collection_schedule',


RedNo7:
### I Have A Problem With:

The integration in general

### What's Your Problem

`day_offset` is not working.  The integration works fine if I remove the `day_offset` line.

I am using v 2.4.0 but only just started using so not sure is a new problem, old problem or just me!

### Logs

```Shell
Logger: homeassistant.config
Source: config.py:940
First occurred: 17:22:20 (1 occurrences)
Last logged: 17:22:20

Invalid config for 'waste_collection_schedule' at JCHH/packages/integrations/waste_collection_schedule.yaml, line 6: 'day_offset' is an invalid option for 'waste_collection_schedule', check: waste_collection_schedule->sources->0->day_offset, please check the docs at https://github.com/mampfes/hacs_waste_collection_schedule#readme
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: cherwell_gov_uk
      args:
        uprn: !secret uprn
      day_offset: -1
      calendar_title: CDC waste collection
  day_switch_time: "23:00"
```
```


### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
Fixed in #3033

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA


<details><summary>Older versions of HACS:</summary>
<p>
install master with HACS version < 2.0: (3-dot menu -> redownload -> change version to master -> fully restart HA (this will increase update cycle by a lot)
</p>
</details>

PR #3033, opened 2024-11-14
Fix: Day_Offset missing in yaml schema

ReneNulschDE:
The "day_offset" setting is missing in the yaml-schema. Looks like that got lost around june-2nd.

fixes: #3027


5ila5:
Thanks for your contribution

Issue #2870, opened 2024-10-18
[Bug]: sbazv_de Not working anymore


reentier:
### I Have A Problem With:

The integration in general

### What's Your Problem

I only receive an error-Message from this Integration singe today.

<img width="1072" alt="image" src="https://github.com/user-attachments/assets/aeab12c7-288d-4c60-9d86-102355b4fd4a">

While trying to reconfigure and double-checking with website of the provider it seems that they have changed the requested paramaters for the API-Call. In the past there asked for City, Part of the City and Street. But now there are asking for City, Street and Number of the House.

<img width="1007" alt="image" src="https://github.com/user-attachments/assets/806eca40-ff57-4bab-8ace-43c9d2560b3d">

If there is any more Information i should provide to solve this issue, please ask.

Thanks in advance for looking into this.

### Source (if relevant)

Sbazv_de

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

maxgisel:
+1

reentier:
The provider has completely changed the way in which the data can be retrieved. Instead of city, district and street, a location ID is now required. You can then use additional parameters to specify which waste you want to retrieve. You can choose between residual waste, paper, yellow bags, leaf bags and Christmas trees.
The URL for the query has also changed completely. The URL for the address from the example in the data now looks like this:
https://fahrzeuge.sbazv.de/WasteManagementSuedbrandenburg/WasteManagementServiceServlet?ApplicationName=Calendar&SubmitAction=sync&StandortID=1448162001&AboID=495&Fra=P;R;WB;L;GS

I have temporarily solved the problem for myself by hard-coding the URL for the query at my address in the sbazv_en.py module. For the moment it works again, but I hope that the changes will soon be included in the official part.

regards

[...]

PR #3041, opened 2024-11-15
Fix SBAZV Provider

Maddimax:
The "api" of SBAZV has changed (see #2870). I used @rmi1974 solution and changed it so the user can simply enter the URL he copied from the Website.


5ila5:
This source just fetches the ICS url, the user need to get anyway. We allready have mechanism for providers providing a ICS file. As it looks like we cannot migrate users gracefully, there is not really a need for a dedicated source.

We probably should remove this source in favor of a ICS source which is a YAML file which generates documentation on how to configure the ICS source for this service provider. A script also creates links to the README.md and in the configuration GUI based on this yaml file. here is a guide on how to create a ICS source: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md

5ila5:
Thanks for your contribution

Issue #3052, opened 2024-11-17
[Bug]: Region doesn't work


HJE0815:
### I Have A Problem With:

A specific source

### What's Your Problem

Die Quelle hat eine ungültige Antwort zurückgegeben: "HTTPSConnectionPool(host='wastemanagement.awg.de', port=443): Max retries exceeded with url: /WasteManagementDonauwald/WasteManagementServlet?SubmitAction=wasteDisposalServices&InFrameMode=true (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1000)')))". Bitte überprüfe die Argumente und versuche es erneut.

### Source (if relevant)

https://www.awg.de/abfallentsorgung/abfuhrkalender

### Relevant Configuration

```YAML
city: Achslach
        street: Aign
        hnr: "1"
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

ReneNulschDE:
This is a problem on the provider side and should be fixed on their side... They have an interesting system config... 

I will prepare a workaround...

HJE0815:
install master/git: Developer tools -> actions -> select update: Install update -> click Coose entity and select WCS -> select version and type master in the input field -> run the action -> restart HA

doesn't work for my, sry

5ila5:
I completely disabled SSL verification as it still seem not to work.

Please repeat the process above and try again

HJE0815:
doesn't work at all

HJE0815:
sry, now i found the right, new Code ->

now it works

PR #3053, opened 2024-11-17
Fix: Switch Awg_de to legacy session handling

ReneNulschDE:
Problem: The ssl config of wastemanagement.awg.de has changed and the cert-chain is [incomplete](https://www.ssllabs.com/ssltest/analyze.html?d=wastemanagement.awg.de).
Solution: Switch to legacy_session
Fixes: #3052 

Tests:
```
Testing source awg_de ...
  found 16 entries for Achslach Aign 1 
  found 24 entries for Böbrach Bärnerauweg 10A
  found 40 entries for Am Bäckergütl 1, 94094 Malching
```

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awg_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awg_de.py
@@ -6,9 +6,9 @@ import random
 import re
 import string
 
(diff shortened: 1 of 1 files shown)


HJE0815:
#3052 doesn't work
even with adapting it with "nano"

Issue #3081, opened 2024-11-23
[Bug]: Warwick DC calendar not loading


jayhix123:
### I Have A Problem With:

A specific source

### What's Your Problem

Calendar stopped loading in home assistant. Went to set it up again from fresh and it's reporting the following issue:

The source returned an invalid response: "time data 'Sat 04/01/2025' does not match format '%d/%m/%Y'". Please check the provided arguments and try again.
![Screenshot_20241123_210146_Home Assistant](https://github.com/user-attachments/assets/23db2099-ce6d-40a1-a517-9627ca092da2)

### Source (if relevant)

Warwick District Council

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
    sources:
    - name: warwickdc_gov_uk
      args:
        uprn: *************
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

jeverley:
Same issue here:

```
This error originated from a custom integration.

Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
integration: Waste Collection Schedule (documentation)
First occurred: 09:14:31 (1 occurrences)
Last logged: 09:14:31

[...]

PR #3088, opened 2024-11-25
Fix: Warwickdc_gov_uk date parsing

ReneNulschDE:
Fixes: #3081 
Error: Exceptional collection are shown with a different date format
Solution: Additional date parsing

Test:
```
Testing source warwickdc_gov_uk ...
  found 16 entries for Test_001
  found 16 entries for Test_002
  found 16 entries for Test_003
```


LeonArmston:
Just tested this by pasting the fix into custom_components/waste_collection_schedule/waste_collection_schedule/source/warwickdc_gov_uk.py and now works fine!

Issue #3078, opened 2024-11-23
[Bug]: Kingston Gov Date format change


the-jeffski:
### I Have A Problem With:

A specific source

### What's Your Problem

It seems the date format in the API has changed to provide / instead of - as a separator.    I amended my local copy for the date comparisons and this fixed the errors.   Assume this is permanent or needs a if loop in case they revert.

### Source (if relevant)

kingston_gov_uk

### Logs

```Shell
Traceback (most recent call last):
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch
    entries: Iterable[Collection] = self._source.fetch()
                                    ^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py", line 76, in fetch
    date=datetime.strptime(
         ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/_strptime.py", line 554, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/_strptime.py", line 333, in _strptime
    raise ValueError("time data %r does not match format %r" %
ValueError: time data '2024/11/29 12:00:00' does not match format '%Y-%m-%d %H:%M:%S'
```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

ilbarone87:
My Kingston API seems broken as well since a few days due to change in time data format.
I've tried to edit kingston_gov_uk.py file and set the correct format to "%Y/%m/%d %H:%M:%S" but after restarting HA I'm getting the below error

[...]

PR #3096, opened 2024-11-26
fix kingston uk waste collection ssl error and new date format

ilbarone87:
Fixes: https://github.com/mampfes/hacs_waste_collection_schedule/issues/3078


5ila5:
Thanks for your contribution

PR #3104, opened 2024-11-28
Fixed wrong URL for abfallwirtschaft-enzkreis.de

JavanXD:
abfallwirtschaft-enzkreis.de is the new domain


5ila5:
For ICS sources you need to modify the YAML file and call the update_docu_links.py script to update the MD file as the modified MD file will get overwritten when someone calls the script in the future.

More details here: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md

PR #3122, opened 2024-12-01
Change wuerzburg_de to ICS

rikroe:
The city of Würzburg, Germany now supports ICS calendars.

This PR adds the required info for the ICS integration and adds deprecation warnings to the existing, custom source.

What would be the process to fully remove the existing source (although it does still work)?


5ila5:
Thanks for your contribution.

I would keep it as long as it returns correct data and delete it if it stops working

Issue #3131, opened 2024-12-02
[Bug]: AWB Emsland changed to 2025. Actual 2024 will be ignored


StKnoll:
### I Have A Problem With:

A specific source

### What's Your Problem

AWB Emsland added 2025 as default.
Waste Collection Schedule use the new year and ignores 2024.

### Source (if relevant)

custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_emsland_de.py

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
fixed in  #3134

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA


<details><summary>Older versions of HACS:</summary>
<p>
install master with HACS version < 2.0: (3-dot menu -> redownload -> change version to master -> fully restart HA (this will increase update cycle by a lot)
</p>
</details>

PR #3134, opened 2024-12-02
AWB Emsland request this AND next years data instead of only next years

niklasdoerfler:
This PR tries to fix https://github.com/mampfes/hacs_waste_collection_schedule/issues/3131 by adding a new (optional) parameter, in which the corresponding request period can be entered to override the default request period.

Therefore, for the specific issue, you can enter "Jahresübersicht 2024" to get the waste collection schedule for the current year.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_emsland_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_emsland_de.py
@@ -3,6 +3,7 @@
 from html.parser import HTMLParser
 
 import requests
+from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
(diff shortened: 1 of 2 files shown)


5ila5:
If it's always `Jahresübersicht {YEAR}` I think it is better to set this in the code instead of the user. In your approach, the user needs to change the config regularly. I think the best way to approach this is to rename the fetch method to `get_collections(self, year: int) -> List[Collections)` and set the year according to the parameter and use a new fetch method like this:

```py
def fetch(self) -> List[Collections]:
  now = datetime.now()
  entries = self.get_collections(now.year)
  if now.month != 12:
    return entries
  try:
    return entries + self.get_collections(now.year + 1)
  except Exception:
    return entries
```

5ila5:
Thanks for your contribution. I like this approach

PR #3197, opened 2024-12-13
Create sbazb_de.py

tj551955:
This is the missing source file for SBAZV, Germany


tj551955:
this request can be deleted. wrong .py file name

PR #3199, opened 2024-12-13
Create sbazv_de.md

tj551955:
this is the missing .md file for SBAZV, Germany


tj551955:
there is an orphaned file with the same name. This old file can be deleted. I have updated the information with this new file here.

PR #3205, opened 2024-12-14
Fix lrasha_de API URL

toXel:
The API URL for lrasha_de has changed. I changed it to the new one.


5ila5:
Thanks for your contribution

Issue #3208, opened 2024-12-14
[Bug]:  Salford Council returning 403 error


DaveR66:
### I Have A Problem With:

A specific source

### What's Your Problem

Has been working perfectly until the last couple of days, now giving 403 error, but the web page/http call returns the correct info when accessed via a browser. Have tried rolling back HA and Integration but no joy.

### Source (if relevant)

salford.gov.uk

### Logs

```Shell
Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
integration: Waste Collection Schedule (documentation)
First occurred: 15:32:06 (1 occurrences)
Last logged: 15:32:06

fetch failed for source Salford City Council: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/salford_gov_uk.py", line 35, in fetch r.raise_for_status() ~~~~~~~~~~~~~~~~~~^^ File "/usr/local/lib/python3.13/site-packages/requests/models.py", line 1024, in raise_for_status raise HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.salford.gov.uk/bins-and-recycling/bin-collection-days/your-bin-collections/?UPRN=xxxxxxxxxxxxx
```

### Relevant Configuration

```YAML
waste_collection_schedule:
    sources:
    - name: salford_gov_uk
      args:
        uprn: "xxxxxxxxxxxxx"
      customize:
        - type: Brown recycling (bottles and cans)
          alias: Recycle Bin
        - type: Domestic waste
          alias: General Bin       
        - type: Food and garden waste
          alias: Green Bin
        - type: Blue recycling (paper and card)
          alias: Paper Bin
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

[...]

PR #3221, opened 2024-12-16
Fix Salford_gov_uk source

FilHarr:
Salford have recently updated their site. Now requires a user-agent in the header otherwise gets a 403 response. Also needed slight amend to check for finding relevant data on page.

Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/3208


5ila5:
Thanks for your contribution

PR #3225, opened 2024-12-17
Fix date parsing for kingston.gov.uk. Fixes #3224

### LINKED ISSUE #3224: Kingston.gov.uk appears to have changed date formats again
@ilbarone87 I think they might have changed something again, I'm now seeing:

```
2024-12-16 09:47:12.097 ERROR (SyncWorker_8) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source The Royal Borough of Kingston Council:
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch
                                    ~~~~~~~~~~~~~~~~~~^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py", line 79, in fetch
    date=datetime.strptime(
         ~~~~~~~~~~~~~~~~~^
        data["echo_refuse_next_date"], "%Y/%m/%d %H:%M:%S"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).date(),
    ^
  File "/usr/local/lib/python3.13/_strptime.py", line 573, in _strptime_datetime
    tt, fraction, gmtoff_fraction = _strptime(data_string, format)
                                    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.13/_strptime.py", line 352, in _strptime
    raise ValueError("time data %r does not match format %r" %
                     (data_string, format))
ValueError: time data 'Friday 20 December' does not match format '%Y/%m/%d %H:%M:%S'
```

_Originally posted by @cmsj in https://github.com/mampfes/hacs_waste_collection_schedule/issues/3078#issuecomment-2545096194_

cmsj:
Date parsing is now in its own function because it has grown in complexity. Kingston has stopped including years in their dates, so we have to add them manually and account for year boundaries.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
@@ -40,14 +40,14 @@ class Source:
         self._uprn = str(uprn)
 
     def interpret_date(self, date_string):
-        now = datetime.now()
-        date = datetime.strptime(date_string + f" {now.year}", "%A %d %B %Y")
+        now = datetime.now().date()
+        date = datetime.strptime(date_string + f" {now.year}", "%A %d %B %Y").date()
 
         if date < now:
             # This means the next collection crosses a year boundary
             date = date.replace(year=now.year+1)
 
-        return date.date()
+        return date
 
     def fetch(self):
         s = requests.Session()
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
@@ -1,6 +1,6 @@
 import logging
 import time
-from datetime import datetime, timedelta
+from datetime import datetime
 
 import requests
 from waste_collection_schedule import Collection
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
```diff
@@ -39,6 +39,16 @@ class Source:
     def __init__(self, uprn: str):
         self._uprn = str(uprn)
 
+    def interpret_date(self, date_string):
+        now = datetime.now()
+        date = datetime.strptime(date_string + f" {now.year}", "%A %d %B %Y")
+
+        if date < now:
```
5ila5, line 46: This is also true for collections on the current date so they will be moved to the next year. You could resolve this by converting `now` and `date` from `datetime` objects to `date` objects before this instead of at the end of the function.

5ila5:
You could also use the `dateutil.parse.parse` function, which might be more resilient against future changes

custom_components/waste_collection_schedule/waste_collection_schedule/source/kingston_gov_uk.py
cmsj: Done.

cmsj:
Interesting point about using `dateutil`, but I'll leave that up to maintainers to decide if they want to add another dependency. Thanks for the review!

5ila5:
We use `dateutil` at a lot of places already.


Thanks for your contribution

PR #3229, opened 2024-12-17
Add Los Angeles, CA

cdiaz2799:
Added source for City of Los Angeles waste collection.

- Uses LA City's GeoQuery API to retrieve collection schedules
- Returns black (trash), blue (recycling), and green (organics) bin pickups
- All bins for a given address are collected on the same day
- Schedules return collection dates for next 2 weeks


5ila5:
Thanks for your contribution, I just renamed your md file.

PR #3240, opened 2024-12-19
updated with apps_ci_minneapolis_mn_us

jimternet:
hope I did ok here - open to feedback


custom_components/waste_collection_schedule/waste_collection_schedule/source/apps_ci_minneapolis_mn_us.py
```diff
@@ -0,0 +1,120 @@
...
+        Args:
+            APN: Area Parcel Number in Minneapolis
+
+        Raises:
+            SourceArgumentRequired: If APN is not provided
+        """
+        if not APN:
+            raise SourceArgumentRequired(
+                "street_address",
+                "A valid street address in Los Angeles is required to look up collection schedule",
```
5ila5, line 49: `Los Angeles` copy-paste leftovers?

custom_components/waste_collection_schedule/waste_collection_schedule/source/apps_ci_minneapolis_mn_us.py
```diff
@@ -0,0 +1,120 @@
...
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+from waste_collection_schedule.service.SSLError import get_legacy_session
+from waste_collection_schedule.exceptions import (
+    SourceArgumentException,
+    SourceArgumentRequired,
+)
+
+TITLE = "Minneapolis MN USA" # Title will show up in README.md and info.md
+DESCRIPTION = "Source script for apps_ci_minneapolis_mn_us"  # Describe your source
+URL = "https://www.minneapolismn.gov/resident-services/garbage-recycling-cleanup/collection-day-lookup/"  # Insert url to service homepage. URL will show up in README.md and info.md
```
5ila5, line 13: Please only use the main URL not the path to the lookup here.

custom_components/waste_collection_schedule/waste_collection_schedule/source/apps_ci_minneapolis_mn_us.py
jimternet: yes, thank you fixing

custom_components/waste_collection_schedule/waste_collection_schedule/source/apps_ci_minneapolis_mn_us.py
jimternet: e.g. `https://www.minneapolismn.gov/`?

5ila5 (commented):
Do you think people using this source know what an APN is?
`APN` will show up as `Apn` in the GUI configuration screen, and does not follow python naming conventions. I'd prefer using `apn` in the __init__ (and readme) and adding 

```python
PARAM_TRANSLATIONS = {
    "en": {
        "apn": "APN"
    }
}
```


Does the source work for you. It does not work for me as I get the Cloudflare captcha as answer from the server (but maybe due to my german IP?)

custom_components/waste_collection_schedule/waste_collection_schedule/source/apps_ci_minneapolis_mn_us.py
5ila5: yes as it shows up in the Readme.md and using the full url would take up more than one line a lot of the times.

jimternet: got it- updated

jimternet:
It does work for me - I had to put in the right combination of headers or I was getting a 403 - I suspect you're right and it doesn't work outside the US. I can have a friend test

jimternet:
I'm fine with that - or I could replace the acronym with "area_parcel_number"  wherever I have APN today - what do you prefer?

5ila5:
If you think everyone using this knows what an APN is, that's fine. I didn't know what that is, but as I'm from Germany I'm not the target of this source.

I'd still prever the argument to be all lower case in the python file (and .md file) and adding 

```py
PARAM_TRANSLATIONS = {
    "en": {
        "apn": "APN"
    }
}
```

To the top of the py file

jimternet:
updated

5ila5:
Thanks for your contribution

jimternet:
Thanks for maintaining  the project!

Issue #3105, opened 2024-11-28
[Bug]: Afval Wijzer doesn't support ICS anymore


rugebiker:
### I Have A Problem With:

A specific source

### What's Your Problem

Since some time ago, Afval Wijzer doesn't support ICS anymore. It only shows the information in their own website: https://www.mijnafvalwijzer.nl/

### Source (if relevant)

https://www.mijnafvalwijzer.nl/

### Relevant Configuration

```YAML
As an example, feel free to use in the website the following address:

Postcode: 5613JZ
Huisnummer: 25
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

gilaberticus:
Just created a pull request to fix this: https://github.com/mampfes/hacs_waste_collection_schedule/pull/3244

5ila5:
fixed by #3244

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA


<details><summary>Older versions of HACS:</summary>
<p>
install master with HACS version < 2.0: (3-dot menu -> redownload -> change version to master -> fully restart HA (this will increase update cycle by a lot)
</p>
</details>

PR #3244, opened 2024-12-19
Adding dedicated mijnafvalwijzer nl

gilaberticus:
mijnafvalwijzer nl does not offer ics support anymore: https://github.com/mampfes/hacs_waste_collection_schedule/issues/3105#issue-2702474142
Therefore I added support via dedicated source (parsing html)


custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
```diff
@@ -0,0 +1,122 @@
...
+    "it": "COME OTTENERE GLI ARGOMENTI",
+}
+
+PARAM_DESCRIPTIONS = { # Optional dict to describe the arguments, will be shown in the GUI configuration below the respective input field
+    "en": {
+        "postcode": "Postcode",
+        "number": "House number",
+        "add": "Addition"
+    },
+}
```
5ila5, line 44: Remove if there is nothing to describe. (This text is shown below the input field, but your text is the same as the placeholder text already populated by the name.)

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
```diff
@@ -0,0 +1,122 @@
...
+                    "december":12
+                }
+        
+        date_day = []
+        date_month = []
+        date_year =[]
+        types_list = []
+        for year in years:
+            year_int = int(year.get('id')[-4:])
+            print(year_int)
```
5ila5, line 92: remove print

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
```diff
@@ -0,0 +1,122 @@
...
+                    Collection(
+                        date = datetime.date(date_year[ii], date_month[ii], date_day[ii]),  # Collection date
+                        t = types_list[ii],  # Collection type
+                        icon = ICON_MAP.get(types_list[ii]),  # Collection icon
+                    )
+                )
+            except:
+                pass
+#                raise Exception("Error in date: " + str(date_year[ii]) + "-" + str(date_month[ii]) + "-" + str(date_day[ii]))
+            ii = ii+1
```
5ila5, line 120: id prefer

```py
for day, month, year, bin_type in zip(date_day, date_month, date_year, types_list):
    ...
```

than this weird while loop, or at least a real for loop

```py
for ii in range(len(date_da)):
    ....
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
```diff
@@ -0,0 +1,122 @@
...
+    },
+}
+
+#### End of arguments affecting the configuration GUI ####
+
+class Source:
+    def __init__(self, postcode, number, add=""):
+        self._postcode = postcode
+        self._number = number
+        self._add = add
```
5ila5, line 60: Maybe check if add is none and set it to "". Your example in the .md file would set add to None and `self._add.replace(" ","")` would result in a runtime error

doc/source/mijnafvalwijzer_nl.md
```diff
@@ -0,0 +1,40 @@
+# Afval Wijzer
+
+Afval Wijzer is no longer supported by the generic [ICS](/doc/source/ics.md) source. For all available configuration options, please refer to the source description.
```
5ila5, line 3: > For all available configuration options, please refer to the source description.

Does not make any sense anymore as THIS is the source description

5ila5 (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
gilaberticus: Removed

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
gilaberticus: removed

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
gilaberticus: Changed

doc/source/mijnafvalwijzer_nl.md
gilaberticus: Changed md file with more info

gilaberticus:
Hi @5ila5 , just committed with requested changes

5ila5 (approved):

5ila5:
Thanks for your contribution

gilaberticus:
Thanks for the review and good suggestions. When will it be released to HACS?

5ila5:
I think I will release a new version soon. But I'd like to address some more issues for now.

You can install the master/git/development version with HACS

<p>
</p>
</details>

custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
```diff
@@ -0,0 +1,113 @@
...
+        for year in years:
+            year_int = int(year.get("id")[-4:])
+            dates = soup.find_all("span", class_="span-line-break")
+            for date in dates:
+                date_day.append(int(date.string.split()[1]))
+                date_month.append(dict_month[date.string.split()[2]])
+                date_year.append(year_int)
+            types = soup.find_all("span", class_="afvaldescr")
+            for type in types:
+                types_list.append(type.string)
```
ricklambrechts, line 95: Thanks for the contribution. Just having HA running for a few days and just found this integration and tested the latest changes.

Currently it is showing 2 entries because we are looping through the years `for year in years` but the dates found by `dates = soup.find_all("span", class_="span-line-break")` are not related to the looped year.

image

My suggestion would be to remove the `for year in years:` loop and use the year of the date string or only search for the dates in the year div.
```py
dates = soup.find_all("span", class_="span-line-break")
for date in dates:
    date_day.append(int(date.string.split()[1]))
    date_month.append(dict_month[date.string.split()[2]])
    date_year.append(int(date.string.split()[3]))
types = soup.find_all("span", class_="afvaldescr")
for type in types:
    types_list.append(type.string)
```

ricklambrechts: Sadly this is changed again. The year selection is removed and the year can only be parsed from a class string because the date string does not contain the year anymore.

I have emailed them if they can add back the ICS file feature.

PR #3246, opened 2024-12-19
Fetch all calendars offered by awm_muenchen_de

mhelff:
At the end of the year, awm_muenchen_de offers two calendar links (2024 and 2025).
Instead of fetching the first one, fetch all offered calendars.
Solves the problem that only the 2024 schedule was fetched.


5ila5:
Thanks for your contribution

PR #3290, opened 2024-12-26
fix: Merton UK has changed the format of the dates

robertmarsal:
Also updated the test cases as they don't seem to be valid IDs any more.

Tested with `python custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s merton_gov_uk -lit` and the tests now pass successfully


5ila5:
Thanks for your contribution

PR #3309, opened 2024-12-28
Update calendar.py

mynkpandey:
Updated code for older versions (older than Python 3.9+) of Python if required anywhere.


5ila5:
Your fix does not fix compatibility with older versions of python, as the type hint is `list` not `List`

Even if this would fix this for this specific place, it would not increase compatibility as we use 3.9 type hinting at a lot of places and even some other type hinting introduced much more recently.

Home Assistant dropped 3.8 support in 2022.1, so I don't see a good reason to keep compatibility.

Python 3.8 is already end-of-life (https://devguide.python.org/versions/) so this is not necessary even outside the context of Home Assistant

PR #3329, opened 2024-12-29
Add Publidata.ca source

MrNunu:
Add publidata Canada, which covers at least Marguerite D'Youville MRC.


custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_ca.py
```diff
@@ -0,0 +1,482 @@
...
+    "enc": "Volumineux",
+    "dv": "Feuilles",
+    "bio": "Compost",
+    "verre": "Verres",
+    "sapin": "Sapin de Noël",
+}
+
+HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
+    "en": "Your address / Votre adresse",
+}
```
5ila5, line 72: Please remove this. This will be shown in the configuration GUI header and does not need to be there. Only use this if there is actually something useful to fill in to the header

custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_ca.py
```diff
@@ -0,0 +1,482 @@
...
+            for schedule in waste_data["schedules"]:
+                if schedule["schedule_type"] in ("regular", "exception"):
+                    my_rruleset.rrule(self._parse_regular(schedule))
+                elif schedule["schedule_type"] in ("closed", "closing_exception"):
+                    my_rruleset.exrule(self._parse_closure(schedule))
+            for entry in my_rruleset:
+                entries.append(
+                    Collection(
+                        entry.date(),
+                        LABEL_MAP.get(waste_type),
```
5ila5, line 477: Maybe use LABEL_MAP.get(waste_type, waste_type) so it will not get None even if the data change and it is not part of the LABEL_MAP anymore

custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_ca.py
```diff
@@ -0,0 +1,482 @@
...
+    },
+    "Saint-Amable City Hall": {
+        "address": "575 Rue Principale, Saint-Amable, QC J0L 1N0"
+    },
+    "Varennes City Hall": {
+        "address": "175 Rue Sainte-Anne, Varennes, QC J3X 1R6"
+    },
+    "Verchères City Hall" : {
+        "address": "581 QC-132, Verchères, QC J0L 2R0"
+    }
```
5ila5, line 46: Maybe reduce your test cases to 3 and comment out the rest. Having a lot of tests, cases accumulates when testing all sources

MrNunu: Done. Kept the first three.

custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_ca.py
MrNunu: Done

custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_ca.py
MrNunu: Done

5ila5:
Thanks for your contribution

PR #3338, opened 2024-12-30
Fix tauranga_govt_nz date

conork123:
Bugfix -  Add a missing else clause which is causing new year pickups to show in the incorrect year.


5ila5:
Thanks for your contribution

PR #3344, opened 2024-12-30
Update Conwy County Borough Council endpoint

tomusher:
It appears the endpoint used to retrieve data for Conwy County Borough Council has recently been updated. This PR updates the provider to use the new endpoint.


5ila5:
Thanks for your contribution

PR #3345, opened 2024-12-30
Fix liverpool council dates in the next year

sysmoon14:
The line to push dates into the next year wasn't working, so I've rewritten it

Tested on my local system and all working well.


5ila5:
The line probably would have worked, but it did not reassign the new date 

```py
# not working
date.replace(year = today.year + 1)
# what would work
date = date.replace(year = today.year + 1)
```

but I like your approach

sysmoon14:
Great, thanks!

PR #3354, opened 2024-12-31
Fix source cederbaum_de due to homepage/data changes

pogross:
The provider made changes to the homepage / data source. These changes should fix that.


5ila5:
thanks for your contribution

PR #3372, opened 2025-01-02
[montreal_ca] Updating to support different sectors

mfortin:
Waste routes could be named differently for a given post code.
This PR allows specifying different sectors per waste types, all while retaining the existing behaviour of only having "sector" being set.
In the event only sector is defined, it will be used for all waste management.
Updated documentation to reflect changes.
Added test cases.

### DIFF
--- a/custom_components/waste_collection_schedule/translations/de.json
+++ b/custom_components/waste_collection_schedule/translations/de.json
@@ -2549,7 +2549,11 @@
         "description": "Konfiguriere deinen Service Provider. Mehr details: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/montreal_ca.md",
         "data": {
           "calendar_title": "Kalender Titel",
-          "sector": "Sector"
+          "bulky": "Sector (Sperrgut)",
+          "food": "Sector (Lebensmittelverschwendung)",
+          "green": "Sector (GreensGrün und Blätter)",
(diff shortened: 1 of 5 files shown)


mfortin:
@julienboriasse This was originally written by you, I had to modify as there are 5 different sector ids in my area.

lit-af:
Nice, j'utilisais l'option multiple sources jusqu'à présent, mais cette solution est meilleure!

mfortin:
@mampfes can this gets a review ? I have been using this updated code in my local HA installation and won't update for the time being

mfortin:
@5ila5 care for a review ?

5ila5:
`PARAM_TRANSLATIONS` and `PARAM_DESCRIPTIONS` currently do not work with fresh as we currently only support languages English, German, Italian.


You can comment out the French translations for now.


## If anyone is interested in adding support for the French language:

To support french, we'd need `fr.json` file in `/custom_components/waste_collection_schedule/translations` translating all the not source specific values. An `fr` entry in the `/default_translations.py` file containing some default translations for common source arguments. And adding `fr` to the `LANGUAGES = ["en", "de", "it"]` line in the `/update_docu_links.py` file.

mfortin:
@5ila5 Done

5ila5:
Thanks for your contribution

PR #3373, opened 2025-01-02
Update gipsprojekt_de.md

jandechent:
Temporary workaround: As the city of Heidelberg (Germany) retired gips project, this offers the temporary workaround for users.


5ila5:
This does not work like this. You need to modify the `doc/ics/yaml/gipsprojekt_de.yaml` and call the `update_docu_links.py` to update the md file, or it will just get overwritten the next time anyone uses update_docu_links.py

jandechent:
Thanks. Then let's focus on the actual issue that Heidelberg broke and leave this be.

5ila5:
Dealt with in #3477 and #3479

Issue #3307, opened 2024-12-27
[Bug]: Lichfield (UK) source breaks on dates in next year


Troon:
### I Have A Problem With:

A specific source

### What's Your Problem

Lichfield DC website doesn't specify the year in the dates on its page. It's currently December: the parsing logic naively assumes the "January" dates shown for the next collections are this year rather than next, and related sensors are thus `unavailable`.
 
I'm just noting this here as the source is mine in the first place. I intend to submit a PR in the next few days when I can find time although it won't make it through to release in time for the New Year, whereupon the problem should go away anyway. If they don't change the site structure by Dec 2025 then it should work for next year.

If you want to fix it locally in the meantime, find:

```
custom_components/waste_collection_schedule/waste_collection_schedule/source/lichfielddc_gov_uk.py
```

add `import datetime` at the top of the file, and add the last two lines of this snippet after the first line:

```
            date = parser.parse(dates[i].text).date()
            if date.month == 1 and datetime.date.today().month == 12:
                date = date.replace(year=date.year+1)
```

then restart HA. Works here: I have the black and blue sensors now showing correctly.

### Source (if relevant)

lichfielddc_gov_uk

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #3374, opened 2025-01-02
Fixing Lichfield year boundary issue.

Troon:
Fixing incorrect parsing in December of January collection dates provided without a year. Resolves issue #3307.


5ila5:
Thanks for your contribution

Issue #3365, opened 2025-01-01
[Bug]: Invalid Response - South Holland District Council (UK)


Lewis1352:
### I Have A Problem With:

A specific source

### What's Your Problem

![image](https://github.com/user-attachments/assets/7c45fc8a-9054-4780-b133-e04694bb5196)

Invalid response when attempting to fetch the data from the South Holland website.

Also when attempting to use HACS instead sensors are populated with unknown, these worked until around the 29th of December (UK time):

![image](https://github.com/user-attachments/assets/d0381c92-dead-482f-966e-c3b518d392e6)
![image](https://github.com/user-attachments/assets/fc7e97d4-7f5a-4316-90dd-9a584a77dcee)

I haven't double checked but I suspect this has to do with the garden waste. Because it's the end of the year the garden waste collection takes a break so they don't put anything on the website (if they do it's very meaningless, see below image).

![image](https://github.com/user-attachments/assets/966dc1e4-87b9-4372-8638-f16d613b8521)

### Source (if relevant)

sholland_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
    sources:
    - name: sholland_gov_uk
      args:
        uprn: -removed-
        postcode: -removed-
      customize:
        - type: Refuse
          alias: Refuse
        - type: Recycling
          alias: Recycling
        - type: Garden
          alias: Garden
    
sensor:
  - platform: waste_collection_schedule
    name: sensor for refuse collection
    types:
      - Refuse
    
  - platform: waste_collection_schedule
    name: sensor for recycling collection
    types:
      - Recycling
      
  - platform: waste_collection_schedule
    name: sensor for garden collection
    types:
      - Garden
```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
[...]

PR #3382, opened 2025-01-02
Update sholland_gov_uk.py

JDoades:
Added path for code when no date is present (This happens during the winter in South Holland, as Garden Bin Waste is not collected all year round)

This has stopped the service from crashing when attempting to add the integration. I am not 100% on whether the integration will still bug when someone already has a garden waste entity created, and then SHDC stop collections. But it at least allows for re-adding the integration.

This was raised in issue https://github.com/mampfes/hacs_waste_collection_schedule/issues/3365

Some addresses in Spalding (PE11) already have new dates, but others (such as my own address, which I am not particularly comfortable sharing publically) do not, and the info is blank currently.
image


5ila5:
Thanks for your contribution

Closed for #3446

Did reformat and moved the if down to the date parsing. I could not push to your branch correctly, so I opened a separate pr in #3446

Issue #3271, opened 2024-12-24
[Bug]: Integration can't load - showing invalid config


ChristophCaina:
### I Have A Problem With:

The integration in general

### What's Your Problem

Since today, the integration is showing an invalid configuration in HomeAssistant - and won't load anymore.

### Logs

```Shell
Logger: homeassistant.setup
Source: setup.py:416
First occurred: 11:39:24 AM (1 occurrences)
Last logged: 11:39:24 AM

Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 416, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/init_yaml.py", line 109, in async_setup
    await hass.async_add_executor_job(
    ...<6 lines>...
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/waste_collection_schedule/waste_collection_api.py", line 88, in add_source_shell
    new_shell = SourceShell.create(
        source_name=source_name,
    ...<3 lines>...
        day_offset=day_offset,
    )
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 217, in create
    source_module: SourceModule = importlib.import_module(
                                  ~~~~~~~~~~~~~~~~~~~~~~~^
        f"waste_collection_schedule.source.{source_name}"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 200, in protected_loop_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
[...]

Issue #3409, opened 2025-01-04
[Source Request]: AttributeError: module 'icalendar' has no attribute 'InvalidCalendar'


LordHayne:
### Municipality / Region

Austria / Klagenfurt am Wörthersee

### Collection Calendar Webpage

https://app.muellapp.com/ical/195684?area_filter=152236%2C152281%2C152331%2C152384&reminder

### Example Address

Finkstraße 44, 9020 Klagenfurt am Wörthersee

### Collection Data Format

As ICS/ICAL file

### Additional Information

Logger: homeassistant.setup
Quelle: setup.py:416
Erstmals aufgetreten: 19:37:37 (1 Vorkommnisse)
Zuletzt protokolliert: 19:37:37

Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 416, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/init_yaml.py", line 109, in async_setup
    await hass.async_add_executor_job(
    ...<6 lines>...
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/waste_collection_schedule/waste_collection_api.py", line 88, in add_source_shell
    new_shell = SourceShell.create(
        source_name=source_name,
    ...<3 lines>...
        day_offset=day_offset,
    )
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 217, in create
    source_module: SourceModule = importlib.import_module(
                                  ~~~~~~~~~~~~~~~~~~~~~~~^
        f"waste_collection_schedule.source.{source_name}"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 200, in protected_loop_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
[...]

Issue #3397, opened 2025-01-04
[Bug]: error after update to HA 2025.1


mkurz0662:
### I Have A Problem With:

The integration in general

### What's Your Problem

After upgrade to HA 2025.1. the integration does not work anymore.

Error message: Unable to prepare setup for platform 'waste_collection_schedule.sensor': Unable to set up component.

### Logs

```Shell
Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 416, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/init_yaml.py", line 109, in async_setup
    await hass.async_add_executor_job(
    ...<6 lines>...
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/waste_collection_schedule/waste_collection_api.py", line 88, in add_source_shell
    new_shell = SourceShell.create(
        source_name=source_name,
    ...<3 lines>...
        day_offset=day_offset,
    )
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 217, in create
    source_module: SourceModule = importlib.import_module(
                                  ~~~~~~~~~~~~~~~~~~~~~~~^
        f"waste_collection_schedule.source.{source_name}"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 200, in protected_loop_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 1026, in exec_module
[...]

Issue #3390, opened 2025-01-03
[Bug]: Error during setup of component waste_collection_schedule


screamer3003:
### I Have A Problem With:

The integration in general

### What's Your Problem

After Update to Hassio 2025.1.0 I get this error

### Logs

```Shell
Logger: homeassistant.setup
Quelle: setup.py:416
Erstmals aufgetreten: 21:40:34 (1 Vorkommnisse)
Zuletzt protokolliert: 21:40:34

Error during setup of component waste_collection_schedule
Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/setup.py", line 416, in _async_setup_component
    result = await task
             ^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/init_yaml.py", line 109, in async_setup
    await hass.async_add_executor_job(
    ...<6 lines>...
    )
  File "/usr/local/lib/python3.13/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/config/custom_components/waste_collection_schedule/waste_collection_api.py", line 88, in add_source_shell
    new_shell = SourceShell.create(
        source_name=source_name,
    ...<3 lines>...
        day_offset=day_offset,
    )
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 217, in create
    source_module: SourceModule = importlib.import_module(
                                  ~~~~~~~~~~~~~~~~~~~~~~~^
        f"waste_collection_schedule.source.{source_name}"
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/src/homeassistant/homeassistant/util/loop.py", line 200, in protected_loop_func
    return func(*args, **kwargs)
  File "/usr/local/lib/python3.13/importlib/__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1387, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1360, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1331, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 935, in _load_unlocked
[...]

PR #3421, opened 2025-01-05
fix: Setup of integration with generic ICS fails since HA 2025.1

z-master42:
In my opinion, removes unused and legacy code that prevents the integration from being set up since HA 2025.1 when using a generic ICS file or URL.

See #3259, #3271, #3390, #3397, #3409

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
@@ -13,7 +13,7 @@ from waste_collection_schedule.exceptions import (
     SourceArgumentNotFoundWithSuggestions,
 )
 from waste_collection_schedule.service.ICS import ICS
-from waste_collection_schedule.service.ICS_v1 import ICS_v1
+# from waste_collection_schedule.service.ICS_v1 import ICS_v1
 
 TITLE = "ICS"
 DESCRIPTION = "Source for ICS based schedules."
@@ -141,7 +141,6 @@ class Source:
         regex: str | None = None,
         title_template: str = "{{date.summary}}",
         split_at: str | None = None,
-        version: int = 2,
         verify_ssl: bool = True,
         headers: dict = {},
     ):
@@ -151,19 +150,11 @@ class Source:
             raise SourceArgumentExceptionMultiple(
                 ("url", "file"), "Specify either url or file"
             )
-        if version == 1:
-            self._ics = ICS_v1(
-                offset=offset,
-                split_at=split_at,
-                regex=regex,
-                title_template=title_template,
-            )
-        else:
-            self._ics = ICS(
-                offset=offset,
-                split_at=split_at,
-                regex=regex,
-                title_template=title_template,
+        self._ics = ICS(
+            offset=offset,
+            split_at=split_at,
+            regex=regex,
+            title_template=title_template,
             )
         self._params = params
         self._year_field = year_field  # replace this field in params with current year
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
@@ -13,7 +13,6 @@ from waste_collection_schedule.exceptions import (
     SourceArgumentNotFoundWithSuggestions,
 )
 from waste_collection_schedule.service.ICS import ICS
-# from waste_collection_schedule.service.ICS_v1 import ICS_v1
 
 TITLE = "ICS"
 DESCRIPTION = "Source for ICS based schedules."


affeldt28:
Why not delete the `ICS_v1.py` like you mentioned in your comment?

z-master42:
Yes, you can do this, but it is not really necessary as it is not integrated anywhere and therefore the code is not used.

Finest (approved):

schmidtfx:
It is a backwards incompatible fix, is there anyway to set version=1?

z-master42:
As far as I can read the code, version == 2 is hardcoded.

ChristophCaina:
I can confirm that this pr has fixed the issue #3271 which I have reported.

Thanks :)

TDCroPower:
the else in line 161 must remain, otherwise the code will not work.
I also made the mistake first and it didn't work, today I looked at the Python code again and the else is necessary for the first if from line 150 to 153.

z-master42:
The if in line 150 has no else. Please pay attention to the indentations. And I don't think this is the same "version".

custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
```diff
@@ -141,7 +140,6 @@ def __init__(
         regex: str | None = None,
         title_template: str = "{{date.summary}}",
         split_at: str | None = None,
-        version: int = 2,
```
5ila5, line 144: This breaks existing installations. Just keep the argument

5ila5:
> The if in line 150 has no else. Please pay attention to the indentations. And I don't think this is the same "version". The version is not set elsewhere in the class, in contrast to url or file, for example.

@z-master42 I don't know what you mean there is no `else` for the `if` in line 150 and there never was one.

Does anyone know if the v1 produces different results than the "new" one? I could not find any differences in the results for the ICS sources I tested.

If we remove it we should probably be consistent and remove it from everywhere: remove v1 file completely, remove the version description from the `ics.md` file, the ics sources `hws_halle_de`, `muensingen_ch` and remove the `recurring_ical_events` dependency all together (requirements.txt, manifest.json, the note in the `README.md`

TDCroPower:
> The if in line 150 has no else. Please pay attention to the indentations. And I don't think this is the same "version". The version is not set elsewhere in the class, in contrast to url or file, for example.

I have just tested it again.
When I remove else: and try to configure my webcal address in the GUI I get an error message.
With else: the configuration via GUI works as usual.

5ila5:
applying the patch removes the else but keeps the self._ics = ..

```py
        if bool(self._url is not None) == bool(self._file is not None):
            raise SourceArgumentExceptionMultiple(
                ("url", "file"), "Specify either url or file"
            )

        self._ics = ICS(
            offset=offset,
            split_at=split_at,
            regex=regex,
            title_template=title_template,
        )
        self._params = params

**Comment by TDCroPower** [comment]
@5ila5 correct, with this variant I get an error message in the configuration GUI after entering my webcal URL and pressing OK.
With the else: line, the GUI continues to run as usual and it works.
I can't quite explain why the behavior is like this.

**Comment by TDCroPower** [comment]
Here the error message without the else: line...

```
Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
Integration: Waste Collection Schedule ([Dokumentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
Erstmals aufgetreten: 14:01:56 (1 Vorkommnisse)
Zuletzt protokolliert: 14:01:56

fetch failed for source ICS: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 208, in fetch return self.fetch_url(self._url, self._params) ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 236, in fetch_url return self._convert(r.text) ~~~~~~~~~~~~~^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 251, in _convert dates = self._ics.convert(data) ^^^^^^^^^ AttributeError: 'Source' object has no attribute '_ics'
```

**Comment by z-master42** [comment]
> @z-master42 I don't know what you mean there is no else for the if in line 150 and there never was one.

That's what I mean.

The
```
if bool(self._url is not None) == bool(self._file is not None):
            raise SourceArgumentExceptionMultiple(
                ("url", "file"), "Specify either url or file"
```

has no `else`-block. But @TDCroPower claims that `else` removed by me would belong to this.

File: custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py (hunk 1 above)
Comment (z-master42) [reply custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py:144]: At what point? "version" is no longer used within the class. This is exactly how it works for me (with an ICS file an some test URL).

**Comment by TDCroPower** [comment]
@z-master42 I can't explain what the error is at the moment, but the results are different if you leave the line out or leave it in.

File: custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py (hunk 1 above)
Comment (devployment) [reply custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py:144]: attention: I don't know much about python and even less about this code base. But to me it looks like `version: int = 2` just declares the default value as version 2. For my limited understanding, the `ics` datasource configuration can configure `version: 1`. Shown in the "test comments" above for example.

https://github.com/mampfes/hacs_waste_collection_schedule/blob/b44e14e73dbd681e0234513ad71c6c08bd579aac/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py#L37

By that, my assumption is that there might be installations out there, that require this code for backwards (version) compatibility. In this class the parameter might not be used any more, but the configuration of these instances seems pretty dynamic. 

So, this might be a hack/fix to make it work locally. But probably not a general fix for the problem. But again, not knowing much about this particular code here, just having the same problem with my installation.

Comment (z-master42) [reply custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py:144]: The version can only be selected if WCS is configured via YAML. When using the UI, which is the preferred and default way, it is always version 2. The version only determines which parser is used.

```
version
(integer) (optional, default: 2)

Selects the underlying ICS file parser:

version: 1 uses recurring_ical_events
version: 2 uses icalevents
```
For more see configuration via YAML: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc%2Fsource%2Fics.md

aseith:
> Here the error message without the else: line...
> 
> ```
> Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
> Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
> Integration: Waste Collection Schedule ([Dokumentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
> Erstmals aufgetreten: 14:01:56 (1 Vorkommnisse)
> Zuletzt protokolliert: 14:01:56
> 
> fetch failed for source ICS: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 208, in fetch return self.fetch_url(self._url, self._params) ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 236, in fetch_url return self._convert(r.text) ~~~~~~~~~~~~~^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 251, in _convert dates = self._ics.convert(data) ^^^^^^^^^ AttributeError: 'Source' object has no attribute '_ics'
> ```

I've ran into the same issue. Check your indentation. The `self._ics` part needs to be on the same level like the `if bool(...)`, like this:

        if bool(self._url is not None) == bool(self._file is not None):
            raise SourceArgumentExceptionMultiple(
                ("url", "file"), "Specify either url or file"
            )
        self._ics = ICS(
             offset=offset,
             split_at=split_at,
             regex=regex,
             title_template=title_template,
        )

BoGnY:
maybe requirement `recurring_ical_events` should be removed as is not used anymore

z-master42:
> > Here the error message without the else: line...
> > 
> > ```
> > Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
> > Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
> > Integration: Waste Collection Schedule ([Dokumentation](https://github.com/mampfes/hacs_waste_collection_schedule#readme))
> > Erstmals aufgetreten: 14:01:56 (1 Vorkommnisse)
> > Zuletzt protokolliert: 14:01:56
> > 
> > fetch failed for source ICS: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 208, in fetch return self.fetch_url(self._url, self._params) ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 236, in fetch_url return self._convert(r.text) ~~~~~~~~~~~~~^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py", line 251, in _convert dates = self._ics.convert(data) ^^^^^^^^^ AttributeError: 'Source' object has no attribute '_ics'
> > ```
> 
> I've ran into the same issue. Check your indentation. The `self._ics` part needs to be on the same level like the `if bool(...)`, like this:
> 
>         if bool(self._url is not None) == bool(self._file is not None):
>             raise SourceArgumentExceptionMultiple(
>                 ("url", "file"), "Specify either url or file"
>             )
>         self._ics = ICS(
>              offset=offset,
>              split_at=split_at,
>              regex=regex,
>              title_template=title_template,
>         )

That's what I talk about all the time. Indentations...

TDCroPower:
@aseith & @z-master42  Perfect, that was the problem, now it works and makes sense again.
I'm not so good at Python and didn't know that it takes indentation so seriously.

custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
5ila5: removing this argument breaks every existing configuration using the ics source as the integration tries to set up the source with an argument not accepted by the __init__


The version arguemnt is registered in every ics source configuration. Additionally it can be set in the GUI it just defaults to 2 but you CAN change it:

image

z-master42:
So can we close this PR here?

custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
5ila5: That would still be a problem when removing it from the `__init__` as it would throw `TypeError: Source.__init__() got an unexpected keyword argument 'version'` when setting starting with an existing configuration having the version attribute to anything.

PR #3430, opened 2025-01-05
Deduplicate entries in ALW-WF

envy:
The ALW Wolfenbüttel API has started to return every entry twice (since 2025-01-01). This causes this integration to also add them to the calender and sesnors twice.

This PR deduplicates the returned entries by creating a "hash" (haha) from the timestamp + bin type so we add every bin type and timestamp combination only once.


5ila5:
Thanks for your contribuiont

PR #3461, opened 2025-01-06
citiesapps_com: Update to citiesapp v2 api

### LINKED ISSUE #3460: [Bug]: CitiesApp does not return 2025 collection dates
### I Have A Problem With:

A specific source

### What's Your Problem

Since the end of 2024 I don't see any collection dates anymore.
After some investigation it looks like CitiesApp hast moved to api v2 for the waste collection dates. I verified this using their app.

I also verified this using one of the test sources provided in the tests folder. As well as by @tobik176 in this comment: https://github.com/mampfes/hacs_waste_collection_schedule/issues/3377#issuecomment-2571577084.

Might be related to #3377 - I did not verify this, however.

Solution:
Update the integration to work with the CitiesApp v2 api

### Source (if relevant)

citiesapps_com

### Logs

_No response_

### Relevant Configuration

_No response_

### Checklist Source Error

### Checklist Sensor Error

### Required

lbloder:
Fixes #3460

Updates citiesapp related code to use v2 api.
Also updates wizard to be compatible.

@mampfes Thank you and all the contributors for your work on this integration. Its much appreciated. Also note that I am not a python developer, so any suggestions are welcome :)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
@@ -1418,7 +1418,7 @@ class CitiesApps:
         if email is not None and phone is not None:
             raise Exception("Only provide one of email or phone not both")
 
-        # get authentication as guest
+        # get authentication
         self._session = requests.Session()
         self._session.headers.update(
             {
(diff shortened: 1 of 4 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
```diff
@@ -1463,47 +1463,28 @@ def get_specific_citiy(self, search: str) -> dict:
...
     def get_specific_calendar(self, city_id: str, search: str) -> dict:
         calendars = self.get_garbage_calendars(city_id)
         for calendar in calendars:
-            if calendar["name"].lower().strip() == search.lower().strip():
+            if calendar["street"].lower().strip() == search.lower().strip():
                 return calendar
         raise SourceArgumentNotFoundWithSuggestions(
-            "calendar", [calendar["name"] for calendar in calendars]
+            "calendar", [calendar["street"] for calendar in calendars]
         )
```
5ila5, line 1478: Looks like the Excption call is wrong (from before you touched it) it should be 
```py
        raise SourceArgumentNotFoundWithSuggestions(
            "city", search, [city["name"] for city in cities]
        )
```
 in get_specific_citiy

and 
```py
        raise SourceArgumentNotFoundWithSuggestions(
            "calendar", search, [calendar["street"] for calendar in calendars]
        )
```
here.

lbloder:
Thanks for the feedback @5ila5, will look into it and update the PR.

overide6:
I made the changed, but after reconfiguration, I still don't get the waste type in the calendar.
what do I need to do to get the waste types into the calendar and then setup the sensors again?
City: Str. Ruprecht an der Raab
Calendar: Unterfladnitz

thx

lbloder:
@overide6 Thank you for testing this. This seems to be the same issue as for Fürstenfeld, which I already have a fix for. In my initial code I did not handle the pagination, so Unterfladnitz is not returned from the API.

I will update the PR here once I'm done testing.

lbloder:
Short update from my side:
Looks like `Rudersdorf` is still using the v1 api. There is an api call which we can use to figure out which garbage collection version is used. I'm currently trying to figure out how to implement both v1 and v2 in a nice manner.

I'm not fully done yet, due to time constraints. I hope to have a working version by Monday.

@overide6 As a quick workaround you can replace:
```
https://api.v2.citiesapps.com/waste-management/by-city/{city_id}/areas
```

with

```
https://api.v2.citiesapps.com/waste-management/by-city/{city_id}/areas?pagination=limit:100
```

in `CitiesAppsCom.py`

overide6:
@lbloder 
thank you, but I don't have this line in my `CitiesAppsCom.py` file located in the "service" folder
`https://api.v2.citiesapps.com/waste-management/by-city/{city_id}/areas`

could you tell me where to find it?
Thank you

lbloder:
@5ila5 
I reworked the `citiesapps_com` source and corresponding service to handle both v1 and v2 cities. It would be great if you could have another look. I tried to keep the separation between source and service, if you have suggestions for improvements, I'd be happy to adapt the code.

@overide6 maybe you could try the new implementation to see if that works for you.

overide6:
@lbloder 
I now see the waste types in the calendar again, so it seems to work. thank you.

custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
```diff
@@ -1459,39 +1459,42 @@ def get_specific_citiy(self, search: str) -> dict:
...
+        if is_v2:
+            api = self.GarbageApiV2(self._session)
 
-        return self.get_garbage_plans(specific_calendar)
+        return {
+            "is_v2": is_v2,
+            "data": api.fetch_garbage_plans(city_dict, calendar)
+        }
+
+    def get_garbage_calendars(self, city_id: str) -> list:
```
5ila5, line 1487: Type hinted as returning list, but returns `dict[str, bool|list]`

5ila5:
The `Rudersdorf Rudersdorf 3` test case seems to fail. If this test case is not compatible with the api_v2, and it cannot be made to work, it's best to remove it and replace it with a working one

lbloder:
> The `Rudersdorf Rudersdorf 3` test case seems to fail. If this test case is not compatible with the api_v2, and it cannot be made to work, it's best to remove it and replace it with a working one

Looks like they migrated to the calendar v2 api in the meantime 🤦 . This was basically the reason I built it so we can handle both v1 and v2 :).
I'll keep the v1 compatibility in the code, not sure when they will migrate all cities to v2. Maybe I can find another city that still uses v1.
I'll update the test case.

custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
lbloder: Ah, thank you, will update.

lbloder:
Updated the Rudersdorf test case and added a suggestion to re-check the App if no calendar could be found.

Also, looks like they are migrating all cities to v2, as of this writing, there are only 7 cities left on v1. Did some spot checking and it seems like they don't return any up-to-date data. Guessing they will also be migrated to v2 when new data is available. 

Once all cities are on v2 we can probably remove all the v1 code.

PR #3468, opened 2025-01-07
remove ICS_v1 support and dependency recurring-ical-events dependency

5ila5:
Remove ICS_v1 and dependency recurring-ical-events
as it caused some problems and did not serve a good porpuse anymore.

migrate GUI configs to remove the version argument from the config

replaces / closes #3421


DerDreschner:
@5ila5 : Just for completeness, you can also downgrade (change the requirements.txt) the following packages as the problem occurs due to incompatible new versions:

```
icalendar >= 4.0.9, < 6.0.0
recurring-ical-events >= 2.0.2, < 3.4.0
```

That solves the problem as well. But yeah, a complete removal is the better solution as it's deprecated anyway.

Kolbi:
Should fix:
https://github.com/mampfes/hacs_waste_collection_schedule/issues/3409
https://github.com/mampfes/hacs_waste_collection_schedule/issues/3495

PR #3491, opened 2025-01-08
Update binzone_uk.py

danw87:
WARNING: BREAKING CHANGE if you have custom rules or rely on matching the GARDEN WASTE events
"GARDEN WASTE" will change to "GARDEN WASTE BIN"

Change ICON_MAP "GARDEN WASTE" to "GARDEN WASTE BIN". Binzone list the bin as "GARDEN WASTE BIN" but post a message 'No garden waste collection due to xx' for reasons such as Christmas / new year which caused false positives


5ila5:
Thanks for your contribution

PR #3497, opened 2025-01-08
Heidelberg: Ignore timezones when parsing christmas collecion date

DerDreschner:
Were not fast enough to fix this little bug when parsing the Christmas tree collection date before your merge.

Sorry for any inconvenience!


DerDreschner:
@5ila5 : There are no changes when I run update_docu_links locally with this branch. Different then the result from the validation pipeline. Could you have a look / pull it anyway? Thanks!

PR #3498, opened 2025-01-08
Heidelberg: Ignore timezones when parsing christmas collecion date

DerDreschner:
Were not fast enough to fix this little bug when parsing the Christmas tree collection date before your merge. My branch wasn't up-to-date anyway, that's why I closed #3497 and tried again.

Sorry for any inconvenience!

PR #3502, opened 2025-01-08
Heidelberg: Revert pre-fill of streets and ignore timezone on Christmas tree collection parse

DerDreschner:
During inspection of the failed CI pipeline on my last pull request (#3497 / #3498), I recognized that using `CONFIG_FLOW_TYPES` with a public method call has unexpected side effects. Whenever the file is being imported (e.g., by running `test_sources.py` or `update_docu_links.py`), the `Source.get_available_streets()` method is being called in order to fill the `CONFIG_FLOW_TYPES` dictionary even without access to values of the dictionary at all. This leads to unnecessary and maybe excessive calls to the public API. [It's visible in this pipeline run here](https://github.com/mampfes/hacs_waste_collection_schedule/actions/runs/12674155047/job/35323279820#step:7:12).

First, I thought about answering calls from both test/debug scripts with an empty list using this workaround:
```python
import inspect

[...]

    @staticmethod
    def get_available_streets() -> list[str]:
        if ("update_docu_links.py" or "test_sources.py") in inspect.stack()[-1].filename:
            return list()
```

But, honestly, I'm pretty sure there may be cases the file is being imported by HA/HACS I'm not aware of right now. And checking the stack-trace on runtime in production environments is pretty nasty. Therefore, I decided to revert this change entirely. In case the street entered by the user isn't correct, I call the API anyway and present the result as suggestions.

Additionally, I ignore the local timezone information now when parsing the Christmas tree collection date in order to avoid shifting it to another date unexpectedly.

Sorry for any inconvenience I caused!

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/heidelberg_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/heidelberg_de.py
@@ -216,7 +216,7 @@ class Source:
         for waste_type in WASTE_TYPES:
             if waste_type == WASTE_TYPES["christmas"]:
                 collection_date = date_parser.parse(
-                    raw_collection_data[waste_type][0]["collection_date"]
+                    raw_collection_data[waste_type][0]["collection_date"], ignoretz=True
                 )
                 ruleset = rruleset()
                 ruleset.rdate(collection_date)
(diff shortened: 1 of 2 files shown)


5ila5:
Thanks for your contribution

Custom pre-filled configuration flow are still very bad. They were an afterthought when creating the config_flow as it is now.
I want to look into this sometime, but I currently lack the time to do so.

PR #3521, opened 2025-01-09
Fix dict indexing on error and resolve the fallback now that the ICAL…

alitheg:
… generation is working again (what I originally wanted to fix last time!)


5ila5:
Thanks for your contribution

PR #3527, opened 2025-01-09
Typo on example of montreal_ca.md

adarrous:
Change example config from address to sector


5ila5:
Thanks for your contribution

PR #3539, opened 2025-01-10
Update calendar url for Seon (Switzerland)

starwarsfan:
Update with new calendar url for 2025


5ila5:
Thanks for your contribution

PR #3542, opened 2025-01-10
update tbr_reutlingen_de to support new source url used for 2025

srempfer:
Adjusted documentation and used new url

### DIFF
--- a/doc/ics/yaml/tbr_reutlingen_de.yaml
+++ b/doc/ics/yaml/tbr_reutlingen_de.yaml
@@ -3,11 +3,18 @@ title: Technischer Betriebsdienst Reutlingen
 url: https://www.tbr-reutlingen.de
 howto:
   en: |
-    - Go to <https://www.tbr-reutlingen.de/entsorgungskalender> and select your street.  
-    - Right-click on `Abfuhrtermine (iCal) herunterladen` and copy link address.
-    - Replace the `url` in the example configuration with this link.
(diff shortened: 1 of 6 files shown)


doc/ics/yaml/tbr_reutlingen_de.yaml
```diff
@@ -3,11 +3,18 @@ title: Technischer Betriebsdienst Reutlingen
 url: https://www.tbr-reutlingen.de
 howto:
   en: |
-    - Go to <https://www.tbr-reutlingen.de/entsorgungskalender> and select your street.  
-    - Right-click on `Abfuhrtermine (iCal) herunterladen` and copy link address.
+    - Open <https://www.tbr-reutlingen.de/entsorgungskalender> and go to section `Entsorgungskalender 2025`.
+    - Select your street, desired waste types and timeperiod.
+    - Right-click on `ICS` and copy link address.
     - Replace the `url` in the example configuration with this link.
```
5ila5, line 9: I don't like this line anymore and try to remove it every time I touch an ICS source. As it also shows up in the GUI configuration where there is no example. I'd prefer something like "Use this URL as the `url` parameter". (same for the German one)

srempfer: Changed the text as requested

srempfer:
Should I also change the example text in [https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md) ?

5ila5:
Sorry for the delay I don't have mcuh time ATM, It will get better next month.



> Should I also change the example text in https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md ?

I'd appreciate it in an additional PR.

PR #3543, opened 2025-01-10
AppAbfallplusDe - Add TBR as supported service

srempfer:
Add Technischer Betriebsdienst Reutlingen as supported service for the Abfall+ App


5ila5:
Thanks for your contribution

PR #3594, opened 2025-01-15
[cc-montesquieu_fr] Fix sources errors

melankh:
- Disable planning for "dechets verts et encombrants" since the 2025's planning is not available
- Fix function to manage malformed html source
- List all cities in sources.json


5ila5:
Is there a better way to do this than to comment out the code, so we don't have to re-add the code once it is available?

melankh:
The Table's HTML code is broken, and I don't know how the 2025 schedule will be published.
So I don't think there's a better solution than waiting some days (weeks ?) for the new web page to be updated.

5ila5:
I dont't like the commented out code. And would prefer an if like (not tested might need to look different)

```py
if parsed_source.find_all("div", class_="block-openclose skin-openclose"):
    global_planning2 = self.get_planning_table_dechets_verts_et_encombrants(
        parsed_source
    )
```

markvp:
@melankh This PR has been open since early 2025. Is the "dechets verts et encombrants" schedule now available on the website? 

As @5ila5 noted, the commented-out code should be replaced with a conditional guard (e.g. `if` check for the section existing) rather than commenting out the block. Could you update this PR to address that feedback?

If the source is no longer relevant or you're unable to update, please let us know and we can close this.

melankh:
Hi, 
I'll try to work on this PR next week.

markvp:
@melankh checking in — any progress on the conditional guard around `dechets verts et encombrants`? Also needs a rebase on master (last commit is from Jan 2025, so there'll be conflicts). Happy to keep this open if you're still working on it; otherwise we'll close for now and welcome a fresh PR once the website's 2026 schedule shape is settled.

melankh:
It should be good now.

markvp (approved):
Thanks for sticking with this and getting it over the line — the PDF-based parsing for the dechets verts schedule is a solid long-term solution, and the normalised commune-name matching handles the encoding edge cases cleanly. Made a couple of small tweaks: reverted the auto-generated files (they're rebuilt by CI on merge) and added the missing `url` key to the `EXTRA_INFO` entries so all 13 communes appear correctly in the source listing. Great work!

Issue #3181, opened 2024-12-11
[Bug]: Source "Abfallwirtschaft Neckar-Odenwald-Kreis" not working


Bomber-G:
### I Have A Problem With:

A specific source

### What's Your Problem

The source "Abfallwirtschaft Neckar-Odenwald-Kreis" is not woring since the beginning of december

### Source (if relevant)

awn_de

### Logs

```Shell
Dieser Fehler stammt von einer benutzerdefinierten Integration

Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
Integration: Waste Collection Schedule (Dokumentation)
Erstmals aufgetreten: 08:10:29 (1 Vorkommnisse)
Zuletzt protokolliert: 08:10:29

[...]

PR #3597, opened 2025-01-15
fixed arg ApplicationName for awn_de

GyroGearl00se:
fixed the issue with awn_de source.
The ApplicationName parameter was wrong / has changed.

Tested successfully


GyroGearl00se:
just added also the option to preselect given waste types which should be imported.

By default all are preselected.
image

5ila5:
Thanks for you contribution

- fixes #3181

PR #3630, opened 2025-01-18
New Source: Simbio si

archon-dev:
Adding the new Source for Slovenia country, Simbio.


custom_components/waste_collection_schedule/waste_collection_schedule/source/simbio_si.py
```diff
@@ -0,0 +1,75 @@
...
+            if not emb_date is None:
+                bin_type = BIN_TYPES.get("E", "E")
+                icon = ICON_MAP.get("E")
+                entries.append(Collection(date=emb_date, t=bin_type, icon=icon))
+
+            bio_date = self.get_date(waste_data.get("next_bio"))
+            if not bio_date is None:
+                bin_type = BIN_TYPES.get("B", "B")
+                icon = ICON_MAP.get("B")
+                entries.append(Collection(date=bio_date, t=bin_type, icon=icon))
```
5ila5, line 64: something like 
```py
        if isinstance(response_data, list) and response_data:
            waste_data = response_data[0]
            for key, value in waste_data.items():
                if not key.startswith("next_"):
                    continue
                bin = key.split("_")[1]
                bin_type, icon = BIN_TYPES.get(bin, bin), ICON_MAP.get(bin)

                mes_date = self.get_date(value)
                if mes_date is not None:
                    entries.append(Collection(date=mes_date, t=bin_type, icon=icon))
```

would remove duplicate code and be more resilient against potential bin name changes

archon-dev: Thank you for suggesting the improvements of code. Code was updated.

5ila5:
Thanks for your contribution.

Just did some minor changes.

PR #3661, opened 2025-01-21
fix(Bordeaux-metropole/Villenave d'ornon): Allow cities without geo_shape

Chuckame:
There are a few cities without geoshape, which means that all the city is impacted by the scheduling (like Villenave d'Ornon).

It has been tested on my HA instance


5ila5:
Thanks for your contribution

PR #3676, opened 2025-01-24
Update zva_wmk_de.py

Demel75:
- convert umlauts
- fix upper case
- new test case


5ila5:
Thanks for your contribution, sorry for the delay. I hadn't had much time the past months

Issue #2925, opened 2024-10-26
[Source Request]: Finland Rosk'n Roll


jfornander:
### Municipality / Region

Inkoo, Finland

### Example Address

You need to have a registered account in "MinRoskis, e-tjänsten" (https://web.rosknroll.fi) self-service portal. Then the information (date for next waste collection) is written as html directly on the webpage along with price, size of container and other things. 

### Collection Data Format

As html on a webpage

### Additional Information

Web page in Finnish or Swedish only I think

jfornander:
![image](https://github.com/user-attachments/assets/d13cb3e6-80e3-450a-8920-176a31079a60)

5ila5:
added in #3678

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA


<details><summary>Older versions of HACS:</summary>
<p>
install master with HACS version < 2.0: (3-dot menu -> redownload -> change version to master -> fully restart HA (this will increase update cycle by a lot)
</p>
</details>

PR #3678, opened 2025-01-24
New provider - RosknRoll, Finland

timbba80:
test_sources.py -s rosknroll_fi
Testing source rosknroll_fi ...
  found 1 entries for Test1

Confirmed to be working with two Home Assistant instances. This is almost 100% identical to Kiertokapula Finland (URL's obviously different and it seems that mixed waste in Kiertokapula is SEK when as in rosknroll it's SE). 

I was not able to validate other waste types as I have only mixed waste container. 

In contribution guidelines there was no instructions to  update sources.json, thus it's not included in the pull request.


cunners:
Fixes #2925

5ila5:
Thanks for your contribution.
Sorry for the delay, I hadn't had much time lately

PR #3693, opened 2025-01-27
Updating poznan source

czyzniek:
Updating poznan source to create calendar's events for all available dates


5ila5:
Thanks for your contribution

PR #3695, opened 2025-01-27
Update WAS Wolfsburg

kenodai:
WAS Wolfsburg updated their website.

- Data can be retrieved via JSON
- City/District is no longer supported
- Street is mandatory now


cunners:
Fixes #3242

PR #3711, opened 2025-01-31
fix missing data for AWB Bad Kreuznach

### LINKED ISSUE #3381: [Bug]: AWB Bad Kreuznach dosn`t work
### I Have A Problem With:

A specific source

### What's Your Problem

Source awb_bad_kreuznach_de dosn`t work, no data

### Source (if relevant)

awb_bad_kreuznach_de

### Logs

```Shell
i dont no were to finde
```

### Relevant Configuration

_No response_

### Checklist Source Error

### Checklist Sensor Error

### Required

alex-voigt:
fixes #3381

The API returns calendar entries from 2024. An extended `until` for calculation is necessary to get entries for 2025.
According to the [documentation](https://dateutil.readthedocs.io/en/stable/rrule.html) `until` and `count` are deprecated when used together, so I removed the count completely.


ElfenB:
I tested it locally by applying the changes and it seems to work. Would be great if we can merge :)

5ila5:
Thanks for your contribution.
Sorry for the delay, I hadn't had much time lately

Issue #3714, opened 2025-02-01
[Bug]: SSL Issue for Cheshire East in UK


greghesp:
### I Have A Problem With:

A specific source

### What's Your Problem

Looks like Cheshire East integration for UK has an SSL issue.  Will investigate myself if I get chance. This was working until I updated Unraid to v7, so may be very specific to some setup here

Tested validating the certificates using curl -v, no issues locally but do have the issues in venv.  Wondering if there's an issue with HA certificates or something 

### Source (if relevant)

cheshire_east_gov_uk

### Logs

```Shell
The source returned an invalid response: "HTTPSConnectionPool(host='online.cheshireeast.gov.uk', port=443): Max retries exceeded with url: /MyCollectionDay/SearchByAjax/GetBartecJobList?uprn=XXXXXXX (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1018)')))". Please check the provided arguments and try again.
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

nerd8192:
Hi, I am also getting the same issue whilst setting up the integration on HAOS (running on Proxmox).

![Image](https://github.com/user-attachments/assets/d5c9c402-58f5-4f81-8e27-73e455e410f4)

nerd8192:
> [@nerd8192](https://github.com/nerd8192) remove your UPRN - it identifies your exact house

👍

ajmorr-is:
Same issue here. Had it all working and at some point (can't say when, but I only realised yesterday) it stopped.
Removed all old lines from configuration.yaml and trying to add from the UI I get the same errors as above.

[...]

PR #3720, opened 2025-02-02
fix: Cheshire East Certificate Verification errors

greghesp:
Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/3714

Seems some HA installs are having issues with certification verification


5ila5:
Thanks for your contribution

Issue #3683, opened 2025-01-25
[Bug]: SSL Certificate Verification Error for MOHU Budapest (former FKF Budapest)


bakoscsaba:
### I Have A Problem With:

A specific source

### What's Your Problem

Source provider for Budapest changed from FKF to MOHU. The new source address is: https://mohubudapest.hu/hulladeknaptar
In the fkf_bp_hu.py file I changed every fkf.hu entry to mohubudapest.hu. It was working for 1-2 days then the integration stopped working.

In the log I found the following error message: 
requests.exceptions.SSLError: HTTPSConnectionPool(host='www.mohubudapest.hu', port=443): Max retries exceeded with url: /hulladeknaptar (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1018)')))

### Source (if relevant)

fkf_bp_hu

### Logs

```Shell
2025-01-22 20:00:23.957 ERROR (SyncWorker_2) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source MOHU Budapest:
Traceback (most recent call last):
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 716, in urlopen
    httplib_response = self._make_request(
        conn,
    ...<5 lines>...
        chunked=chunked,
    )
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 404, in _make_request
    self._validate_conn(conn)
    ~~~~~~~~~~~~~~~~~~~^^^^^^
  File "/usr/local/lib/python3.13/site-packages/urllib3/connectionpool.py", line 1061, in _validate_conn
    conn.connect()
    ~~~~~~~~~~~~^^
  File "/usr/local/lib/python3.13/site-packages/urllib3/connection.py", line 419, in connect
    self.sock = ssl_wrap_socket(
                ~~~~~~~~~~~~~~~^
        sock=conn,
        ^^^^^^^^^^
    ...<8 lines>...
        tls_in_tls=tls_in_tls,
        ^^^^^^^^^^^^^^^^^^^^^^
    )
    ^
  File "/usr/local/lib/python3.13/site-packages/urllib3/util/ssl_.py", line 458, in ssl_wrap_socket
    ssl_sock = _ssl_wrap_socket_impl(
[...]

PR #3721, opened 2025-02-02
fix: fkf_bp_hu - Rename and fix fkf_bp_hu to mohu_bp_hu

greghesp:
Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/3683, and adds option SSL verification for users with issues


TomiEckert:
Will this get merged at some point, or should we just manually install from the fork?

5ila5:
Thanks for your contribution,
I added a migration to migrate all GUI configuration automatically. YAML based configuration need to change the name manually

PR #3749, opened 2025-02-07
[New Source] Panda Waste, Ireland

### LINKED ISSUE #2894: [Source Request]: Add support for Panda Waste in Ireland
### Municipality / Region

Cork, Ireland

https://pandagateway.azurewebsites.net/api/domestic/Account/next-collections

### Example Address

_No response_

### Collection Data Format

As html on a webpage, Dedicated IOS/Android APP

### Additional Information

Couldn't find a site with the calendar that doesn't require authentication:

https://my.panda.ie/waste/next-collections

gcacace:
Implemented source for Panda Waste, Ireland.

Fixes #2894.


gcacace:
Question for the maintainers of this repo: I'm wondering whether there is a better way to manage the authorization token: wouldn't be better storing it in HA until it's about to expire, so every waste schedule refresh request won't call the authentication API all the time?

cunners:
Fixes #2894

5ila5:
It's once a day, so I don't think it's that big of a deal, but you could store the token as an attribute of the source object so it would not call the API every day but just on HA reboot or if it expires. Storing it directly in HA so it isn't called even on reboot isn't that straight forward, and we do not have a good approach for that.

PR #3768, opened 2025-02-11
Support for waste collection schedules [PreZero Bielsko-Biała]

bladowski:
Support for waste collection schedules provided by [PreZero Bielsko-Biała](https://prezero-bielsko.pl/harmonogram-odbioru-odpadow/), serving the city of Bielsko-Biała, Poland.


5ila5:
Thanks for your contribution

I fixed your test case (typo in house_number), I remove self._year as this would fetch the collection from the year HA was last restarted which might cause some problems at the end of a year. I replaced your use of SourceArgumentException as this exception needs to get the problematic source argument as first argument, but you used it as a normal exception.

PR #3792, opened 2025-02-16
Fix: Sunderland_gov_uk

### LINKED ISSUE #3791: [Bug]: Sunderland City Council
### I Have A Problem With:

A specific source

### What's Your Problem

Sunderland City Council (sunderland_gov_uk). "The source returned an empty response."  Seems to have broke since 2025.2.3, 2025.2.4 does not fix the issue.

### Source (if relevant)

sunderland_gov_uk

### Logs

```Shell
No relevant log entries
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

cunners:
Fixes #3791 

Currently returning no results. Looks like a slight change to the API. 
-The postcode no longer requires a + in the space.
-The list of addresses that are returned includes a comma after the house number.

This PR fixes it, but requires users to enter the address in a specific way, ie with the comma included. The address matching would also work with just the house number or name so it may be worth changing to use that rather than include the road name too. The documentation also included a reference to house_number rather than address so it may have been tried like this previously?

```
Testing source sunderland_gov_uk ...
  found 2 entries for Test_001
    2025-02-19 : Recycle
    2025-02-26 : House
  found 2 entries for Test_002
    2025-02-21 : Recycle
    2025-02-28 : House
  found 2 entries for Test_003
    2025-02-20 : Recycle
    2025-02-27 : House
```

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sunderland_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sunderland_gov_uk.py
@@ -11,16 +11,16 @@ URL = "https://www.sunderland.gov.uk/"
 API_URL = "https://webapps.sunderland.gov.uk/WEBAPPS/WSS/Sunderland_Portal/Forms/bindaychecker.aspx"
 HEADERS = {"user-agent": "Mozilla/5.0"}
 TEST_CASES = {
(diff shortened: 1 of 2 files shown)


cunners:
Seems like the website changes have been reverted so the original code now works again.

PR #3795, opened 2025-02-16
Add tsceskybrod_cz

moneytoo:
Hi,

Here's a schedule I created for my local city including all municipal parts with only little over 7 thousand inhabitants. Unlike other sources, there is no API source or calendar. There is only a [table](https://www.tsceskybrod.cz/pravidelny-svoz-odpadu) and a few text rules. 
I already use calendars to track the schedule, but it gets more and more complicated. Some schedules are based on even/odd weeks but some collections are weekly instead and based on the current month. To further complicate it, in some months it's only the first even/odd week in a month.
Because of that, I decided to create a source. It supports all listed streets or locations and all types of waste that is collected from households.

I ran tests as well as tried my fork directly in HA:

image

image


5ila5:
Thanks for your contribution, I changed the exception thrown when the street does not match any street from the list. This way you'll see a list of available streets when using the GUI (Or in the HA logs)

PR #3832, opened 2025-02-25
Added support for Komorniki and Kostrzyn commune in Poland (zys_harmonogram_pl) - fix for loop and add auto NAME_MAP

sematuszewski:
Hello,

i fixed for loop and added auto NAME_MAP with uff-8 fix for this source.


custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
```diff
@@ -0,0 +1,125 @@
...
+
+
+    def fetch(self):
+        try:
+            return self.get_data(API_URL.format(self._commune_name,datetime.datetime.now().year))
+        except Exception:
+            _LOGGER.debug(
+                f"fetch failed for source {TITLE}: trying different API_URL ..."
+            )
+            return self.get_data(API_URL.format(""))
```
5ila5, line 55: requires two format arguments only one given

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
```diff
@@ -0,0 +1,125 @@
+####Mod by SEMATpl - semat.pl - based on sepan_remondis_pl ###
+import datetime
+import json
+import logging
+import defusedxml.ElementTree as XMLDEFUSE
```
5ila5, line 5: This is not a dependency of this module and adding addtional dependencies should only be done when really neccesary. You could use BeatuifulSoup with "xml" parsing, It should probably work pretty much the same:
```py
from bs4 import BeautifulSoup
[...]
tree = BeautifulSoup(table, "xml")
```

5ila5:
The test case only works when fixing the format issue.

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
sematuszewski: I replaced XMLDEFUSE to BeautifulSoup

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
sematuszewski: I fixed this :)

custom_components/waste_collection_schedule/waste_collection_schedule/source/zys_harmonogram_pl.py
```diff
@@ -0,0 +1,125 @@
...
+    def get_data(self, api_url):
+        r = requests.get(f"{api_url}/addresses/cities")
+        r.raise_for_status()
+        city_id = 0
+        cities = json.loads(r.text)
+        for item in cities:
+            if item["value"] == self._city:
+                city_id = item["id"]
+        if city_id == 0:
+            raise Exception("city not found")
```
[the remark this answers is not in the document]
sematuszewski: Thank you for your suggestion, I will improve the code when I have some free time :)

5ila5:
Thanks for your contribution


I implemented the Exception improvement, did some reformatting and run the  update_docu_links.py script

sematuszewski:
> Thanks for your contribution
> 
> I implemented the Exception improvement, did some reformatting and run the update_docu_links.py script

Thank you for your help :)

PR #3841, opened 2025-02-27
ecoharmonogram_pl: return if sides_matcher is empty when creating entries in loop

amozeo:
fixes a logical error, where it loop over every side when side_matcher is not provided and it would return on first found side when side_matcher is provided

that meant that if side_matcher didn't match against first found side,
there would be no entries found

one schedule provider in test cases also updated names for sides, so I updated this test accordingly


5ila5:
thanks for your contribution

PR #3850, opened 2025-03-02
update_docu_links.py - add cs as supported language

danoh:
We have new collection system, so the code has to be modified
I made also czech translation for the config flow

### DIFF
--- a/custom_components/waste_collection_schedule/translations/de.json
+++ b/custom_components/waste_collection_schedule/translations/de.json
@@ -9961,6 +9961,7 @@
         "description": "Konfiguriere deinen Service Provider. Mehr details: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/fkf_bo_hu.md",
         "data": {
           "calendar_title": "Kalender Titel",
+          "ssl_verify": "Ssl Verify",
           "street": "Straße"
         },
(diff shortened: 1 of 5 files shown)


custom_components/waste_collection_schedule/translations/cs.json
```diff
@@ -0,0 +1,17777 @@
...
+          "additional": "If checked you will be prompted to add another sensor after saving this one."
+        }
+      },
+      "reconfigure": {
+        "title": "Reconfigure Source",
+        "description": "Configure your service provider. {howto}More details: {docs_url}.",
+        "data": {
+          "calendar_title": "Calendar Title"
+        }
+      },
```
5ila5, line 111: needs to be translated manually

custom_components/waste_collection_schedule/translations/cs.json
```diff
@@ -0,0 +1,17777 @@
...
+          "url": "Url",
+          "verify_ssl": "Verify Ssl",
+          "version": "Version",
+          "year_field": "Year Field"
+        },
+        "data_description": {}
+      },
+      "args_ahe_de": {
+        "title": "Configure Source",
+        "description
```
5ila5, line 17777: needs to be translated manually as well

PR #3866, opened 2025-03-05
fix(source/torridge): fix date parts unpacking

tigattack:
Somehow I managed to break the code in #3816 between the tests passing and committing the source. Not sure how that happened, but here's the fix. Sorry and thanks 😄


5ila5:
Thanks, Not quite sure how this slipped through, I thought I tested it as well, but it looks like I did not.

PR #3868, opened 2025-03-05
Add Dinhard Switzerland

DonPablo1010:
Add waste collection source for Municipality of Dinhard.

This adds support for waste collection in the municipality of Dinhard, Switzerland.
Supported waste types: regular waste (with holiday adjustments), organic waste (Grüngutabfuhr), and paper/cardboard collection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2077,6 +2077,7 @@ If your service provider is not listed, feel free to open a [source request issu
 - [Berg](/doc/source/a_region_ch.md) / a-region.ch
 - [Bühler](/doc/source/a_region_ch.md) / a-region.ch
 - [Canton of Zürich](/doc/ics/openerz_metaodi_ch.md) / openerz.metaodi.ch
+- [Dinhard](https://www.dinhard.ch/verwaltung/veranstaltungen.html/167) / dinhard.ch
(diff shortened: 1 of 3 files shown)


5ila5:
I don't really like the hard coded holidays (they are only for 2025) and the ICS parsing could be a lot easier if you'd use the ICS service (`from waste_collection_schedule.service import ICS`). This does not really look that robust. At this point, you're probably better of requesting the search page and parsing the HTML with BeautifulSoup:

```py
entries = []
r = requests.post(
    "https://www.dinhard.ch/verwaltung/veranstaltungen.html/167",
    data={
        "eventSearchParams.category": "12",
        "eventSearchParams.query": "",
        "eventSearchParams.dateFrom": "",
        "eventSearchParams.dateTo": "",
    },
)
r.raise_for_status()
soup = BeautifulSoup(r.text, "html.parser")
events = soup.select("div.mod-entry-header")
for event in events:
    title = event.select_one("h2").text
    date_str = event.select_one("time").attrs.get("datetime")
    if date_str is None:
        continue
    event_date = date.fromisoformat(date_str)
    entries.append(Collection(event_date, title))
return entries
```

Could not really test this as the site behaves very weirdly for me not shure whats up with it.

PR #3871, opened 2025-03-06
fix(source): GVU Melk (umweltverbaende_at)

AustrianBarrel:
Currently, trying to use the source "Austria > GVU Melk" gives an empty response error.
Screenshot 2025-03-06 094027

Checking [their website](https://melk.umweltverbaende.at/fuer-die-bevoelkerung/abfuhrterminkalender/), I saw that they use a different URL path.

Adding the path makes it work in my Home Assistant.


5ila5:
Thanks for your contribution,

I added some additional fix and clean up in #3883, so I'll merge this

PR #3883, opened 2025-03-08
update umweltverbaende_at to fix GVU Melk and GVA Baden

5ila5:
also updated EXTRA_INFO and Test cases

closes #3871

@AustrianBarrel thanks for your work


kevinjanisch:
With the master branch installed I get the Error: Could not find nonce for page https://gvabaden.at//fuer-die-bevoelkerung/abholtermine/" 
while configuring via the UI:
image

I guess thats an error on my side? But since manually adding the ics url works i thought i would mention it here.

5ila5:
`baden`, `Berndorf` work for me. Make sure you really installed the master version (the version shown by HACS should be a hash like 7468f02 and not a version number

kevinjanisch:
Thats exactly the hash i have. It seems that all of the municipality with multiple collection zones don't work if you don't add it as calendar. Which is to be expected? "Hernstein" works perfectly. But "Berndorf" or "Baden" failes with that nonce error for me.

Works perfectly if i add Abfuhrbereich II as calendar. I guess this is working as expected? Probably the error message could be more expressive?

5ila5:
Improved errors in #3901

PR #3888, opened 2025-03-09
Add "Krieglach" to citiesapps_com

FilipHoertner:
- Added "Krieglach" to SERVICE_MAP
- Updated docs with "update_docu_links.py"


5ila5:
Thanks for your work and contribution, I updated the whole service map for citeisapps_com, adding some other regions as well in #3904, so I'll close this one in favor of #3904

PR #3911, opened 2025-03-14
New Source: Preston City Council 

RickSeymour:
Supports
Commercial + Domestic

(First commit to a HA project, and first Python commit)


custom_components/waste_collection_schedule/waste_collection_schedule/source/preston_gov_uk.py
```diff
@@ -0,0 +1,134 @@
...
+    "ctl00$MainContent$hdnPropertyEntityValue": "",
+    "ctl00$MainContent$txtAddress": "",
+    "ctl00$MainContent$hdnUPRN": "",
+}
+
+class Source:
+    def __init__(self, street = "", uprn = ""):
+        self._street = street
+        self._uprn = str(uprn)
+        self._session = Session()
```
5ila5, line 59: I don't think keeping the session for days is a good idea. You're doing every request anyway, I think it's a better idea to create a new Session object every time fetch is called

RickSeymour: Thanks. Committed your suggestion. move Session() from __init__() to _get()

5ila5:
Thanks for your contribution

PR #3943, opened 2025-03-21
Updated CCC to public API

camcamnz:
Updated CCC source to use public API, replacing previous call using authenticated API with hard coded credentials


shupershuff:
Legend! Was wondering why I couldn't get it to work

codyc1515:
This doesn't work because the issue is not with that API. I will raise a PR now.

PR #3952, opened 2025-03-24
Fix oxford_gov_uk

rbrunt:
csrf token no longer present on a separate landing page, so now have to load the form page and submit that for the csrf token.

After this change, I've re-run the tests and they now pass:

```
$ git log --oneline | head -n1
e590679d Fix oxford_gov_uk
$ python test_sources.py -s oxford_gov_uk -t
Testing source oxford_gov_uk ...
  found 6 entries for Magdalen Road
  found 8 entries for Oliver Road (brown bin too)
```


josephsellers:
Works for me. Thank you!

davideyre:
Good here too. Thanks!

5ila5:
Thanks for your contribution

PR #3966, opened 2025-03-26
oslokommune: support filtering on pickup point ID

lersveen:
Added an optional field `point_id` to filter by. A single address can have multiple pickup points in the array `HentePunkts` and they can be far apart. At my home address there are 12(!) of them, so the data currently returned contains a lot of noise and does not accurately tell me when collection happens at my closest pickup point.

I also made `house_letter` optional as there is not always one and the API does not require it. Current behaviour is confusing as you're required to enter one, and I had to get around it by entering a space in that field.

Example of a API query for address with multiple pickup points:
https://www.oslo.kommune.no/xmlhttprequest.php?service=ren.search&street=N%C3%A5kkves%20vei&number=5&street_id=15280

Output of `test_sources.py -s oslokommune_no`:

```
Testing source oslokommune_no ...
  found 4 entries for Villa Paradiso
  found 1 entries for Nåkkves vei
```


lersveen:
Anyone able to review? @ccsalvesen @5ila5

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
```diff
@@ -52,11 +83,17 @@ def fetch(self):
             'street_id': self._street_id,
         }
 
+        if self._house_number:
+            args['letter'] = self._house_letter
+
```
5ila5, line 88: This seems wrong. Why are you adding house_letter if house_number is truthy?

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
```diff
@@ -30,14 +37,38 @@
...
 class Source:
-    def __init__(self, street_name, house_number, house_letter, street_id):
-        self._street_name  = street_name
+    def __init__(
+        self,
+        street_name,
+        house_number,
+        street_id,
+        house_letter = None,
+        point_id = None,
```
5ila5, line 65:
```suggestion
        point_id:int|None = None,
```

Adding this type hint will enforce an integer when configuring through the HA GUI


If you're at it, you could type hint the other ones as well (not relevant for config as it assumes str if no type hint given.

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
lersveen: Oops, good catch, that's just a typo.

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
lersveen: Yep, I think I tried this first, but then reverted it when I saw HA's form input for int was the kind with inline arrow buttons, and it seemed wrong for an ID input. Might be some way to avoid that?

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
```diff
@@ -52,11 +83,17 @@ def fetch(self):
             'street_id': self._street_id,
         }
 
+        if self._house_number:
```
lersveen, line 86:
```suggestion
        if self._house_letter:
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
5ila5: We use the default integer input from HA, You can modify it quite a bit by using the Integer Selector (https://www.home-assistant.io/docs/blueprint/selectors/#number-selector)

But I don't think that's really necessary.

I personally think the arrows are fine as you can input with your keyboard OR use the arrows and the errors automatically indicate to the user to input an integer

lersveen: Fixed!

lersveen:
Any idea why the validation is failing @5ila5? Not quite able to see what is wrong.

5ila5:
You modified parameter description and translation
These changes need to be applied in the language jaon files

There is a script that does this for you (`update_docu_links.py`)


Thats why the message before the error is

> looks like the update_docu_links.py did not run

lersveen:
@5ila5 Sorry, took a while to get back to this, but should be fixed now!

5ila5:
Thanks for your contribution,
Sorry for the delay.

Issue #3959, opened 2025-03-25
[Bug]: Redbridge Council - no longer working


oerix:
### I Have A Problem With:

A specific source

### What's Your Problem

All sensors are showing as 'unknown' for 2 weeks+ now. I'm not sure if it's a bug, or the api has changed. However, I did notice that I was required to use an account in order to view the bin dates on their website, otherwise it would show a `Sorry, an error occurred while processing your request` error.

### Source (if relevant)

redbridge_gov_uk

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
I'l close this for #3971 as #3971 has already the PR attached even though this Issue was earlier

PR #3971, opened 2025-03-28
Fix: Redbridge, UK

### LINKED ISSUE #3965: [Bug]: Redbridge_gove_uk no longer working
### I Have A Problem With:

A specific source

### What's Your Problem

This sensor no longer appears to be working - when I noticed it I removed it and when trying to re-add I get the below error

The source returned an invalid response: "500 Server Error: Internal Server Error for url: https://my.redbridge.gov.uk/RecycleRefuse?uprn=[REDACTED]". Please check the provided arguments and try again.

When using the council's website to check the dates manually at https://my.redbridge.gov.uk/RecycleRefuse it no longer appears to work unless logged in.

Once logged in the URL it returns to download the file (which now appears to be a PDF) is https://my.redbridge.gov.uk/RecycleRefuse/GetFile?uprn=
whereas the error has a slightly different URL without GetFile 

### Source (if relevant)

redbridge_gov_uk

### Logs

```Shell
The source returned an invalid response: "500 Server Error: Internal Server Error for url: https://my.redbridge.gov.uk/RecycleRefuse?uprn=[REDACTED]". Please check the provided arguments and try again.
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

num1r0:
Fixes #3965 

Looks like the collections PDF is still available to public (without an account). It requires the UPRN to be provided in the GET request.

Since the response is a PDF file, we are now extracting collections details from the tables downloaded from the council's website. List of changes below:

- Removed BeautifulSoup import (not used anymore)
- Added pdfplumber, BytesIO and logging to the imports
- Updated one of the test cases' UPRN number, due to the old one not returning "Garden" collections anymore
- Updated the API URL

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/redbridge_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/redbridge_gov_uk.py
@@ -41,7 +41,7 @@ class Source:
                 table = page.extract_table()
                 tables.append(table)
 
-        collections = []
+        entries = []
         
         # Month and year regex pattern
         month_year_pattern = r"\b([A-Za-z]+) \d{4}\b"
@@ -62,7 +62,7 @@ class Source:
                 elif isCollection:
                     date_waste_type = e
                     try:
-                        collections.append(
+                        entries.append(
                             Collection(
                                 date=datetime.strptime(f"{date_waste_type.split("\n")[0]} {month_year}", "%d %B %Y"),
                                 t=f"{date_waste_type.split("\n")[1]}",
@@ -71,7 +71,7 @@ class Source:
                         )
                         # In case we have multiple collections of the same day
                         if date_waste_type.split("\n")[1] != date_waste_type.split("\n")[-1]:
-                            collections.append(
+                            entries.append(
                                 Collection(
                                     date=datetime.strptime(f"{date_waste_type.split("\n")[0]} {month_year}", "%d %B %Y"),
                                     t=f"{date_waste_type.split("\n")[-1]}",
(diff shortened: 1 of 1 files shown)


oerix:
Thanks for the fix, but you might want to reference the original issue [here](https://github.com/mampfes/hacs_waste_collection_schedule/issues/3959).

5ila5:
pdfplumber is not a dependency of this project, so we either need to add it as a dependency or find another way to parse this. We had some problems introducing new dependencies in the past. 
We did not do PDF parsing in the past, but I merged a source doing PDF parsing with pydf. Is there any way you could use pydf instead of pdfplumber to do PDF parsing? So we don't introduce another PDF parsing dependency?

oerix:
Issue still persists.
```
HA: 2025.6.1
WCS: 2.8.0
```

oerix:
It's been months. When will this be merged already?

5ila5:
This can not be merged as is it uses the pdfplumber library which is not marked as dependency in the manifest.json, and I'd prefer not to introduce a new dependency but first try to solve this with the existing libraries (pdpdf for PDF parsing)

oerix:
> This can not be merged as is it uses the pdfplumber library which is not marked as dependency in the manifest.json, and I'd prefer not to introduce a new dependency but first try to solve this with the existing libraries (pdpdf for PDF parsing)

Any progress on `pdpdf` parsing please? 
Also, it seems they have updated the frontend and backend recently, resulting in the sensor breaking again.

5ila5:
Redbridge seems to be supported without PDF parsing and fixed in #5214

PR #3986, opened 2025-03-31
Fix: Bracknell Forest source not working with house names

### LINKED ISSUE #3985: [Bug]: Bracknell Forest source not working with house names
### I Have A Problem With:

A specific source

### What's Your Problem

The Bracknell Forest source does not appear to be working when using a house name. Looking at the code, this is because a space is automatically added after the input provided as house number, which is not present when a house name is used.

### Source (if relevant)

bracknell_forest_gov_uk

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

tombotch:
Fixes #3985


5ila5:
Thanks for your contribution

I added a minor change so it does not fail when house_number is an integer (which may happen when configuring through the configuration.yaml and not adding `""` to the house_number

PR #3991, opened 2025-04-02
Added Motala to source avfallsapp_se.py

calleel:
Added municipality Motala as an option in the source avfallsapp_se.py


5ila5:
Thanks for your contribution

Issue #3995, opened 2025-04-04
[Bug]: AHA Region Hannover does not work anymore (again) since last HAS updated


kwithus:
### I Have A Problem With:

A specific source

### What's Your Problem

source does not longer work 
I paste a log below, this quote is than followed by my street name and a suggestion of street names I may use, which does only include names within the city the city of Hannover, not my city in the region. So maybe this is part of the issue, that the city name is not forwarded, as the city of Hannover is pre-populated when I go to the website to check if the source generally does work (which is the fact).
For me it looks like the integration does not change this repopulated city-selection to the correct city
I did not change the source in the config.
When I try to add the source again (exactly the same) the integration tells me, that this source is already existing. When I try to add the source again with a different house number, the integration tells me, that the street does not exist, although I use exactly the same writing as on the website of the source.
And again, the dropdown list only suggests streets from inside of Hannover, not from my selected town. Same true when I select any other city outside of Hannover. 

### Source (if relevant)

Germany, AHA Zweckverband Region Hannover

### Logs

```Shell
Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
Integration: Waste Collection Schedule (Dokumentation)
Erstmals aufgetreten: 3. April 2025 um 17:02:06 (3 Vorkommnisse)
Zuletzt protokolliert: 08:15:39

fetch failed for source Zweckverband Abfallwirtschaft Region Hannover: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/aha_region_de.py", line 94, in fetch raise SourceArgumentNotFoundWithSuggestions( "strasse", self._strasse, [select.text for select in selects] ) waste_collection_schedule.exceptions.SourceArgumentNotFoundWithSuggestions: We could not find values for the argument 'strasse' with the value....
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
[...]

PR #4002, opened 2025-04-08
Fix for Aha website change.

### LINKED ISSUE #3455: [Bug]: AHA Zweckverband Region Hannover doesn't work any work (most likely same issue as for other sources)
### I Have A Problem With:

A specific source

### What's Your Problem

since HA 2025.1 AHA Zweckverband Region Hannover (Germany) is one of the sources which don't work anymore.
Is there any fix on the way?
Many Thanks to all looking into this.

### Source (if relevant)

AHA Abfallzweckverband Region Hannover

### Logs

_No response_

### Relevant Configuration

_No response_

### Checklist Source Error

### Checklist Sensor Error

### Required

Zappdidappdi:
The website of Aha had some changes. Because of the changes you can't get the adresses of all streets anymore, so you have to hand over the first letter of your street.
Furthermore the "ladeort" field has no ID anymore but a name, so we needed to change that.
3 of 5 tests are succesfull, the other tests fail because there are no ICS for these adresses anymore.
Probably fixes #3455 and #3995.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/aha_region_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/aha_region_de.py
@@ -73,7 +73,7 @@ class Source:
     def fetch(self):
         # find strassen_id
         r = requests.get(
-            API_URL, params={"gemeinde": self._gemeinde, "von": "A", "bis": "["}
+            API_URL, params={"gemeinde": self._gemeinde, "von": self._strasse[0]}
         )
         r.raise_for_status()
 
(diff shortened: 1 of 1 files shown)


Xyaren (approved):

modem-man-gmx:
Is the fix merged in now? With most recent version 2.7.0, (HA Core 2025.4.4, Frontend 20250411.0 refreshed to "latest" on 2025-04-25), my street is still not be selectable in the UI:

image
image

5ila5:
Thanks for your contribution

Issue #4029, opened 2025-04-17
Unable to add property ID for Hornsby Shire Council


revixisse:
I am unable to add the property ID during the wizard due to an expired SSL certificate link in the wizard
 
 ![Image](https://github.com/user-attachments/assets/c5dea56e-db1d-4017-a15a-b2c37dd9a977) 

 _Originally posted by @revixisse in [#3529](https://github.com/mampfes/hacs_waste_collection_schedule/issues/3529#issuecomment-2811668871)_

Additionally, the link provided in the documentation for acquiring the property ID is also invalid.

![Image](https://github.com/user-attachments/assets/22222791-c11b-4582-adcf-416e86270df5)
(link: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/hornsby_nsw_gov_au.md)

![Image](https://github.com/user-attachments/assets/7f2ccf34-0e4a-4e84-acc8-55084f75c271)

This is a working link for an online council service to get your property ID:

https://[hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySummaryView.aspx?EnquiryListId=2&ERB=25961](https://hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySummaryView.aspx?EnquiryListId=2&ERB=25961)

![Image](https://github.com/user-attachments/assets/d557e185-e1c6-4a79-8d5c-4c79d96f76ae)

dt215git:
I can't replicate the setup failure -  if I use one of the property ID from the script I get calendar entries. The test scripts are also returning results.

![Image](https://github.com/user-attachments/assets/7691f0f8-72af-47dc-94c8-5ea416fa8e61)

The Property ID search URL will need to be updated, but it looks like the number after  `ERB=` increments with every page visit, so it'll probably need to reference the home new page.

FussySteak:
Hi @dt215git, I have managed to get my property ID from 
https://hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySummaryView.aspx
When I try to add intergration, the system stuck here for over 20 mins

![Image](https://github.com/user-attachments/assets/8c62d8b7-d1c8-4b0a-8f4a-ef2eb74c7c4c)

[...]

PR #4031, opened 2025-04-17
Update: Hornsby, NSW

dt215git:
Updated url in .md file as mentioned in #4029


sh00t2kill:
Any chance we can roll back this commit ? Its wrong; the link provided does not give the correct ID.

dt215git:
I'm not certain that would solve the issue - the original observation in #4029  was that the url in the .md file didn't work.
It still gives the same error for me:

image

If there's a better way to get the property ID needed for the script, wouldn't a new PR with updated instructions be the better approach?

sh00t2kill:
That's the problem. There *is* no way to get it.

I have reworked the API, I now need to re work the provider here.

Issue #4010, opened 2025-04-12
Bug Region Landkreis Osnabrück "AWIGO" is not working


Guido19671:
### Municipality / Region

Landkeis Osnabrück

### Collection Calendar Webpage

https://www.awigo.de/index.php

### Collection Data Format

As html on a webpage

### Additional Information

Hello I cannot get the waste collection dates from Landkreis Osnabrück (AWIGO), I tried via YAML and GUI Interface but get the following Error message from the GUI:

Die Quelle hat eine ungültige Antwort zurückgegeben: "404 Client Error: eID not registered for url: https://www.awigo.de/index.php?eID=awigoCalendar&calendar[method]=getCities&calendar[rest]=1&calendar[paper]=1&calendar[yellow]=1&calendar[brown]=1&calendar[mobile]=1". Bitte überprüfe die Argumente und versuche es erneut.

MatthiasDahlke:
Hello,
i think the Source has changed. Take a look to the AWIGO Website (https://www.awigo.de/index.php) with this source it is possible to download the ics. 

https://www.awigo.de/fileadmin/kalender/AWIGO_Kalender_Rest_Papier_Gelb_Bio_68674151001.ics

It is possible to change this in the Waste Collection Schedule HACS App ?

 Regards

markvp:
This was fixed by PR #4102 (merged May 2025) which updated the source to use the new API endpoint. All test cases are passing. Please update your HACS integration to the latest version and it should work again. Closing.

Issue #3993, opened 2025-04-03
[Bug]: AWIGO doesnt work sice update to HAOS15.1


mircebirke1399:
### I Have A Problem With:

A specific source

### Problem
Suddenly stopped working since Update to 15.1

### LOG
```
fetch failed for source AWIGO Abfallwirtschaft Landkreis Osnabrück GmbH: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/awigo_de.py", line 63, in fetch r.raise_for_status() ~~~~~~~~~~~~~~~~~~^^ File "/usr/local/lib/python3.13/site-packages/requests/models.py", line 1024, in raise_for_status raise HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 404 Client Error: eID not registered for url: https://www.awigo.de/index.php?eID=awigoCalendar&calendar[method]=getCities&calendar[rest]=1&calendar[paper]=1&calendar[yellow]=1&calendar[brown]=1&calendar[mobile]=1
```

### Configuration
```
waste_collection_schedule:
  sources:
    - name: awigo_de
      args:
        ort: Quakenbrück
        strasse: XX.
        hnr: "XX"
  fetch_time: "04:00"
  day_switch_time: "10:00"
```
```
- platform: waste_collection_schedule
  name: "Gelber Sack"
  details_format: "upcoming"
  value_template: "{{ value.daysTo }}"
  add_days_to: True
  types:
    - Gelbe Tonne/Gelben Sack
# Nächste Abholung Restmül
- platform: waste_collection_schedule
  name: "Restmüll"
  details_format: "upcoming"
  value_template: "{{ value.daysTo }}"
  add_days_to: True
  types:
    - Restmülltonne
# Nächste Abholung Bio Tonne
- platform: waste_collection_schedule
  name: "Bio-Abfall"
  details_format: "upcoming"
  value_template: "{{ value.daysTo }}"
  add_days_to: True
  types:
    - Bio-Tonne
#Nächste Abholung Papiermüll
- platform: waste_collection_schedule
  name: "Papiermüll"
  details_format: "upcoming"
  value_template: "{{ value.daysTo }}"
  add_days_to: True
  types:
[...]

PR #4044, opened 2025-04-20
Update awigo_de.py

spechter:
The parameter 'eID' has changed to 'legacy_eID'.

Even though the name implies more changes will be due at one point, local test of this verified that querying works again.

Related to #3993 and #4010; I assume #3447 might have been about a different problem.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awigo_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awigo_de.py
@@ -49,7 +49,7 @@ class Source:
     def fetch(self):
         s = requests.Session()
         args = {
-            "eID": "awigoCalendar",
+            "legacy_eID": "awigoCalendar",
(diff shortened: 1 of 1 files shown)


spechter:
I did actually receive the waste types containing Umlaute incorrectly decoded as well. 
On my instance I've manually added sensors for each type of waste which worked fine.

You have set the encoding explicitly to utf-8 in line 109 or did I miss something?

mircebirke1399:
How can I use this fixed Version in Hassio?

hillbicks:
> How can I use this fixed Version in Hassio?

You have to manually make the changes in your local ha installation, at least I didn't see a way to use a Beta version.

It worked for me, but only for a brief moment. Made the changes, restarted ha, got the calendar back. Couple hours later they were gone again.

Not sure if this is really a reliable way to get this info, since the awigo site seems to be really flaky.

Issue #1709, opened 2024-01-22
[Bug]: Setup failed for custom integration 'waste_collection_schedule': Requirements for waste_collection_schedule not found: ['shapely'].


Ruko2010:
### I Have A Problem With:

The integration in general

### What's Your Problem

After Update to 1.45.0 I get this error in the logs and waste collection is not working any more. Even if I remove the configuration from my configuration.yaml I get this errors.

### Source (if relevant)

ics File

### Relevant Configuration

```YAML
I use Home Assistant in an Docker enviroment.
```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

CedricBojoly:
same issue here https://github.com/mampfes/hacs_waste_collection_schedule/issues/1705

Ruko2010:
ok, so I'm watching the other issue an closing this one.

PR #4045, opened 2025-04-20
Add Support for City of Greater Bendigo VIC, Australia

### LINKED ISSUE #2258: [Source Request]: City Of Greater Bendigo
### Municipality / Region

City Of Greater Bendigo, Australia

https://www.bendigo.vic.gov.au/residents/general-waste-recycling-and-organics/bin-night

### Example Address

555 Napier St, White Hills VIC 3550

### Collection Data Format

As html on a webpage

### Additional Information

On webpage and also via pdf download

iSnackTwoPointOh:
# Pull Request: Add Support for City of Greater Bendigo VIC, Australia
## Description
Added support for the City of Greater Bendigo VIC, Australia. 
No ICS file but they use an app called pozi for geographic overlays including the bin schedule so was able to work with that.

## Key Features
Collection types:
- general waste ✅ 
- recycling ✅ 
- green waste ✅ 
- Note: not including hard waste and commercial waste since the data had no entries for either of these (pretty sure Bendigo has no hard waste collection anymore either)

Accepts input coords and checks with the downloaded geometry which schedule they belong to. Orignally used Shapely to process the geomtry but removed that based on feedback and now using a custom ray tracing algo.

## Testing
Test Script: Verified functionality using test_sources.py:
```python
python3 test_sources.py -s bendigo_vic_gov_au      
Testing source bendigo_vic_gov_au ...
  found 3 entries for Bunnings Epsom
  found 3 entries for Bunnings Bendigo
  found 3 entries for Alfa Kitchen Bendigo
```

closes #2258


sagwefwefga:
Nice work! assuming we have to wait for them to add/approve this and then update the integration?

iSnackTwoPointOh:
@sagwefwefga yeah pretty much. Hopefully there’s not too much of an issue with the added dependency. Otherwise I’ll have to hand roll something

5ila5:
We allready tried adding shaply as dependency in the past and had some trouble with this and decided to remove it again (see  #1705 and #1709)

I'm not the biggest fan of hard-coding the geocode.earth API Key of a commercial third party API

I cannot see the map when visiting the page (probably GeoBlocking issue) So I cannot really look into this.

iSnackTwoPointOh:
@5ila5 i've removed the need for the geocode.earth api in favour of just starting from lat/lon. I've also added a ray tracing algo for the geomtry instead of using shapely. Might not be as robust but for a single source its probably good enough

custom_components/waste_collection_schedule/waste_collection_schedule/source/bendigo_vic_gov_au.py
```diff
@@ -0,0 +1,253 @@
...
+                    "longitude",
+                    str(self._longitude),
+                    f"Longitude must be between {BENDIGO_BOUNDS['min_lon']} and {BENDIGO_BOUNDS['max_lon']}"
+                )
+        except (ValueError, TypeError) as e:
+            raise SourceArgumentNotFound(
+                "coordinates",
+                f"({latitude}, {longitude})",
+                f"Invalid coordinate format. Please provide numeric values. Error: {str(e)}"
+            )
```
5ila5, line 90: This does not work like this just raise a normal exception here

As Ha config will try to match the exception to an input filed to show the error message next to this, but there is no field coordinates 

```suggestion
            raise Exception(
                f"Invalid coordinate format. Please provide numeric values. Error: {str(e)}"
            )
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/bendigo_vic_gov_au.py
```diff
@@ -0,0 +1,253 @@
...
+            zone_names = [zone["properties"]["name"] for zone in found_zones]
+            _LOGGER.debug("Point found in multiple zones: %s", zone_names)
+            raise SourceArgumentNotFound(
+                "coordinates",
+                f"({self._latitude}, {self._longitude})",
+                f"Coordinates are on a boundary between multiple zones: {', '.join(zone_names)}. "
+                "To resolve this, please try adjusting your coordinates slightly (by 0.0001 degrees or less) "
+                "in any direction. You can verify your zone at "
+                "https://www.bendigo.vic.gov.au/residents/general-waste-recycling-and-organics/bin-night"
+            )
```
5ila5, line 176: same

```suggestion
            raise Exception(
                f"Coordinates ({self._latitude}, {self._longitude}) are on a boundary between multiple zones: {', '.join(zone_names)}. "
                "To resolve this, please try adjusting your coordinates slightly (by 0.0001 degrees or less) "
                "in any direction. You can verify your zone at "
                "https://www.bendigo.vic.gov.au/residents/general-waste-recycling-and-organics/bin-night"
            )
```

5ila5 (changes requested):
I'm not that big of a fan of all these helper functions inside the fetch function. I think it would be easier to read if you'd put them in the class (maybe marked as @staticmethod) or just above the 

This is ready to merge when this and the above-mentioned Exceptions are changed

iSnackTwoPointOh:
> I'm not that big of a fan of all these helper functions inside the fetch function. I think it would be easier to read if you'd put them in the class (maybe marked as @staticmethod) or just above the
> 
> This is ready to merge when this and the above-mentioned Exceptions are changed

agreed with the helpers. I've made them private statics on the class as suggested. Thanks for the feedback :)

5ila5 (approved):

5ila5:
Thanks for your contribution

sagwefwefga:
Unfortunately I believe it’s now broken

iSnackTwoPointOh:
> Unfortunately I believe it’s now broken

Looks like there was an API change recently. I've created #5177  to track. Hopefully ready next couple days.

PR #4047, opened 2025-04-21
Add Garden Waste Support for North Somerset

footiefan92:
Simple PR for adding Garden Waste item, newly added to the council data table


5ila5:
Thanks for your contribution

I found another issue with this source which I fixed in #4111

PR #4063, opened 2025-04-26
Update poriruacity_govt_nz.py

Nduva-Hey:
Update website javascript url (JS_URL) to v6 (was v4).


5ila5:
Thanks for your contribution, I noticed that the test cases did not work anymore so I updated them

PR #4065, opened 2025-04-27
Added support for bank holiday overrides in royalgreenwich_gov_uk

timocov:
The page with alternative collection dates during bank holidays: https://www.royalgreenwich.gov.uk/recycling-and-rubbish/bins-and-collections/bank-holiday-collection-dates


5ila5:
Thanks for your contribution

PR #4100, opened 2025-05-08
Fix: Zweckverband Abfallwirtschaft Region Hannover, DE

### LINKED ISSUE #4082: [Bug]: AHA Region Hannover - no data
### I Have A Problem With:

A specific source

### What's Your Problem

Hi, unfortunately, the data retrieval no longer works (without external intervention), most likely in connection with one of the last system updates (message ‘unknown’). The configuration of the source also appears unchanged. The data in YAML is unchanged and correct.  HA Calendar ist empty. Thank you for your support.

### Source (if relevant)

aha_region_de

### Logs

```Shell

```

### Relevant Configuration

```YAML
# Waste_Collection_Schedule
waste_collection_schedule:
    sources:
      name: aha_region_de
      args:
        gemeinde: Hannover
        strasse: xxxxxxxxx. / Kirchrode
        hnr: "xx"
        zusatz: ""
```

### Checklist Source Error

### Checklist Sensor Error

### Required

causa-prima:
Fixes #4082


5ila5:
This was already addressed in #4002 If the additional class matcher is necessary for some edge case let me know and I'll reopen this

causa-prima:
@5ila5: I looked at #4002, and I think my version might be avoiding some errors that the version from #4002 does not:
1. My version does check whether a value for `self._strasse` has been set before calling `self._strasse.upper()[0]` (the `upper()` might not be needed, though). The version from #4002 will throw an `index out of range`, while my version will throw an exception with a more fitting error message (though I'm not sure if there is a better exception pointing to the value being mandatory that should be used instead of the general exception).
2. The version from #4002 does not check the `class` attribute of the Ladeort, but there are multiple `input` elements with name `ladeort`, but only those with `class` `form-control` seemed to contain the id of the Ladeort. Thus, the wrong input might be selected that does not have a value.

Or am I wrong and these things do not matter?

5ila5:
I don't think your first argument really matters as the config GUI requires you to set a non-empty string and for YAML Configuration you'd need to explicitly set the street to be the empty string, or you'd get another Error before even running the script. But you second argument is probably a good idea, even though it worked for me even without this addition

5ila5:
Thanks for your contribution

Issue #4103, opened 2025-05-09
[Source Request]: NL-Gemeente Venray


MikevanBreePXL:
### Municipality / Region

Venray

### Collection Calendar Webpage

https://afvalkalender.venray.nl

### Example Address

Postal Code: 5809EH
House Number: 12

Requesting a GET over `https://afvalkalender.venray.nl/adressen/5809EH:12` returns with address info, including a `bagid`

```json
[
    {
        "bagid": "09xxxxxxxxxxxxxx",
        "postcode": "5809EH",
        "huisnummer": 12,
        "huisletter": "",
        ...
    }
]
```

This `bagid` is futher used to retrieve the list of Waste Collection Dates when requesting a GET over `https://afvalkalender.venray.nl/rest/adressen/<BAG_ID>/afvalstromen` (Bag ID yet to be filled in)

resulting JSON:
```json
[
    {
        "id": 26,
	"parent_id": 0,
	"title": "PMD",
	"slug": null,
	"tags": null,
	"page_title": "PMD",
	"content": "",
	"menu_title": "PMD",
	"icon": "zak-geel-blik-drank",
	"icon_data": <IMAGE IN BASE-64 TEXT>,
       	"ophaaldatum": "2025-05-20" <Or NULL when not on schedule/not requested>
    },
    ...
]

### Collection Data Format

As html on a webpage, Dedicated IOS/Android APP, As ICS/ICAL file

### Additional Information

Dedicated App:
Google Play Store: https://play.google.com/store/apps/details?id=nl.opzet.zcalendar.venray
iOS AppStore: https://apps.apple.com/nl/app/afvalkalender-venray/id1199013983

iCal Apple: webcal://afvalkalender.venray.nl/ical/0984200000024569
iCal Windows/GoogleCalendar: https://afvalkalender.venray.nl/ical/0984200000024569

This is the city coucil's API, rather than HVCGroep which requires the address to be a customer of their service (I am not)

PR #4122, opened 2025-05-13
Add ICS support for [NL] Gemeente Venray

MikevanBreePXL:
Added documentation for ICS support. Issue described in #4103


5ila5:
If You want to provide dutch translation you'd need to create the `custom_components/waste_collection_schedule/translations/nl.json` file then you need to add translations to everything that does not starts with `args_` or `reconfigure_` as these are generated by the update_docu_links.py script (translations you need to touch are at the top and the bottom of this file). You could also provide some common translations for source arguments in the default_translations.py file. Then run the update_docu_links.py script.

Then the whole configuration step has dutch translations and one can add additional translations for sources.

markvp (approved):
Thanks for adding Venray — and for digging up the council's own iCal feed; that's a genuinely useful addition since the existing HVC source needs a paid service subscription.

I've trimmed this down to just the doc/ics/yaml/venray_nl.yaml definition (the only file we hand-maintain for ICS providers — README, sources.json, translations and the doc/ics/*.md page are generated automatically after merge, so I reverted those). I also removed the Dutch (nl) howto block for now, since nl isn't in our supported-language list yet — enabling it needs a bit more plumbing (a translations/nl.json etc.), so I'll open a separate issue for that and we can land the English version now. Everything else looks good.

PR #4157, opened 2025-05-20
Add support for week ranges and missing dates into nvaa_se

### LINKED ISSUE #4052: [Bug]: NVAA_SE, cant handle slam entry with week range.
### I Have A Problem With:

A specific source

### What's Your Problem

Norrtälje Vatten & Avfall,  nvaa_se https://www.nvaa.se/ is returning 1 or 2 entrys called "slam" (septic tanks) which only are specified as a range of weeks with a dash between first and last week.  (v27 - v28)
This makes the integration fail to configure with an error:  "invalid literal for int() with base 10: '-'"
NVAA is working for addresses without the septic tank entries.

There are similar problems reported for other providers but I have not found any reports about week ranges.

Screenshot from nvaa.se:
Image

### Source (if relevant)

nvaa_se.md

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

alu-:
Fixes #4052

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/nvaa_se.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/nvaa_se.py
@@ -40,7 +40,7 @@ def parse_date(next_pickup_date):
     """
     result = re.search(r"v(\d*) - v(\d*) (\d*)", next_pickup_date)
     if result:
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

PR #4201, opened 2025-05-31
Add missing form field for highpeak source

drsgoodall:
The form field "FINDBINDAYSHIGHPEAK_FORMACTION_NEXT" is required to progress to the next form page. Without this each http post is still on the post code entry page.


5ila5:
I cannot really test this as I'm geo-blocked but it looks good

Thanks for your contribution

PR #4202, opened 2025-05-31
Add support for Delta, BC, Canada

infamy:
Add support for Delta BC, even if it has to use fixed dates for holidays. More holidays will be added if excepted (a full year at a time) Once it is published. No ics or other format is available we get a pdf calendar.


5ila5:
Delta, BC is supported by the Recyclecoach source. I added Delta to the documentation in #4204

PR #4211, opened 2025-06-03
Update barrowbc_gov_uk.md

audigex:
The "Add to iCalendar" link seems to have been removed, but the ICS calendar URI still works. Added an instruction for how to manually modify the link

Fixed typo in link


5ila5:
You cannot modify the MD file directly, it will get overwritten. You need to update doc/ics/yaml/barrowbc_gov_uk.yaml and run the `update_docu_links.py` script to update the MD file

PR #4242, opened 2025-06-11
Added support for Kiama Municipal Council, NSW, Australia

daffster:
Added Kiama.
kiama.nsw.gov.au


custom_components/waste_collection_schedule/waste_collection_schedule/source/kiama_nsw_gov_au.py
```diff
@@ -0,0 +1,78 @@
...
+        html_content = data["responseContent"]
+
+        # Parse HTML with BeautifulSoup
+        soup = BeautifulSoup(html_content, "html.parser")
+
+        # Find all waste service result articles
+        results = soup.find_all("div", class_="waste-services-result")
+        entries = []
+        if results:
+            print("Bin Collection Dates:")
```
5ila5: Please remove prints

If you want to add information in logs use logger module

custom_components/waste_collection_schedule/waste_collection_schedule/source/kiama_nsw_gov_au.py
```diff
@@ -0,0 +1,78 @@
...
+
+def parse_date(date_str):
+    """Convert date string like 'Fri 20/6/2025' to ISO format '2025-06-20'."""
+    try:
+        # Parse date string (format: 'Fri DD/M/YYYY')
+        date_obj = datetime.strptime(date_str, "%a %d/%m/%Y")
+        # Convert to ISO format (YYYY-MM-DD)
+        return date_obj.strftime("%Y-%m-%d")
+    except ValueError as e:
+        return f"Invalid date format: {date_str}"
```
5ila5: You parse a string to a datetime object then back to a string and than you parse it again to a date object why not just

```suggestion
def parse_date(date_str) -> date:
    """Convert date string like 'Fri 20/6/2025' to date object"""
    try:
        # Parse date string (format: 'Fri DD/M/YYYY')
        return datetime.strptime(date_str, "%a %d/%m/%Y").date()
    except ValueError as e:
        return f"Invalid date format: {date_str}"
```


Then you don't need to call it with `date.fromisoformat(parse_date(...))` but just `parse_date(...)`

custom_components/waste_collection_schedule/waste_collection_schedule/source/kiama_nsw_gov_au.py
```diff
@@ -0,0 +1,78 @@
...
+                bin_type = result.find("h3")
+                bin_type = bin_type.get_text(strip=True) if bin_type else "Unknown"
+
+                # Get next service date from <div class="next-service">
+                next_service = result.find("div", class_="next-service")
+                day = next_service.get_text(strip=True) if next_service else "No date provided"
+
+                # Only print if a date is available
+                if day and day != "":
+                    print(f"{bin_type}: {day}")
```
5ila5: Remove this print as well

for debugging purposes, you can just put the -l flag to the test script

daffster:
@5ila5 Thanks for the feedback. I've made those changes and tested.

5ila5:
Thanks for your contribution

I did some reformatting and copied the configuration description into the python file so it will show up in the HA configuration GUI


This will be part of the next release or is available now as master (git) version 

Issue #3561, opened 2025-01-12
Source IlRifiutologo not respond anymore


fudu:
### I Have A Problem With:

A specific source

### What's Your Problem

I waws able to configure Integration with source IlRifiutologo, but after a couple of days it gets error.
After a while reconfigured and worked, now it stopped to work again

### Source (if relevant)

IlRifiutologo

### Logs

```Shell
2025-01-12 19:14:15.399 ERROR (MainThread) [homeassistant.config_entries] Error setting up entry Il Rifiutologo for waste_collection_schedule
Traceback (most recent call last):
  File "/usr/local/lib/python3.12/site-packages/requests/models.py", line 974, in json
    return complexjson.loads(self.text, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/simplejson/__init__.py", line 514, in loads
    return _default_decoder.decode(s)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/simplejson/decoder.py", line 386, in decode
    obj, end = self.raw_decode(s)
               ^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/site-packages/simplejson/decoder.py", line 416, in raw_decode
    return self.scan_once(s, idx=_w(s, idx).end())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
simplejson.errors.JSONDecodeError: Expecting value: line 1 column 1 (char 0)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/src/homeassistant/homeassistant/config_entries.py", line 635, in __async_setup_with_context
    result = await component.async_setup_entry(hass, self)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/config/custom_components/waste_collection_schedule/init_ui.py", line 46, in async_setup_entry
    shell = await hass.async_add_executor_job(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 58, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #4247, opened 2025-06-13
fix: Source IlRifiutologo not respond anymore

matteolel:
Porting the fix to #3561 proposed by @iuginP here https://github.com/mampfes/hacs_waste_collection_schedule/issues/3561#issuecomment-2932658908 .

Source IlRifiutologo was not working without the fix.


5ila5:
Thanks for your contribution

Issue #4303, opened 2025-06-24
Swindon has stopped working with a 403 error


psfshr:
Swindon (swindon.gov.uk) has stopped working in the last week or so.

Checking the logs it looks like it's being refused:

`HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.swindon.gov.uk/info/20122/rubbish_and_recycling_collection_days?uprnSubmit=Yes&addressList=xxxxxxxxxxx`

The URL works when pasted directly into a browser.

jamess5:
Same here!

joshwillcock:
As a short term solution you can replace the swindon_gov_uk source with the changes from the above PR. (https://github.com/mampfes/hacs_waste_collection_schedule/pull/4316) #4316

pgale:
Thanks - that solved the issue for me.

5ila5:
This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA

PR #4316, opened 2025-06-26
Swindon: Passing user-agent to bypass detection. Issue #4303

joshwillcock:
Added User Agent to avoid bot detection for the Swindon Source. 
Issue: #4303 4303


5ila5:
Thanks for your contribution

Issue #4310, opened 2025-06-25
[Bug]: Recycle coach stopped working and throws error on trying to configure same source again


muzammilk:
### I Have A Problem With:

The integration in general

### What's Your Problem

The sensors and calendar stopped working so I tried setting up again and now it throws an error while setting it up as new. The address which I am using was working till last week.

### Source (if relevant)

Recycle Coach (recyclecoach_com)

### Logs

```Shell
The source returned an invalid response: "HTTPSConnectionPool(host='pkg.my-waste.mobi', port=443): Max retries exceeded with url: /app_data_zone_schedules?project_id=590&district_id=DELR&zone_id=zone-z7778-z19169 (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7f306b1f8e10>: Failed to establish a new connection: [Errno -5] Name has no usable address'))". Please check the provided arguments and try again.
```

### Relevant Configuration

```YAML
Installation method Home Assistant OS
Core 2025.6.2
Supervisor 2025.06.1
Operating System 15.2
Frontend 20250531.3

Waste collection schedule 2.8.0

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

gregprosser:
This has been failing for a few days for me - it looks like my logs say since June 20:

```
$ docker logs homeassistant |& grep 'fetch failed for source'
2025-06-20 01:54:24.247 ERROR (SyncWorker_10) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source Recycle Coach:
2025-06-20 01:54:42.152 ERROR (SyncWorker_15) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source Recycle Coach:
[...]

PR #4317, opened 2025-06-26
[recyclecoach_com] Use apigw hostname rather than broken pkg.my-waste.mobi, refactor URL code

gregprosser:
`pkg.my-waste.mobi` does not seem to exist in DNS anymore and is causing the source to fail, see #4310

Unfortunately, `us-web.apigw.recyclecoach.com` also no longer seems to find data for all regions.

Thankfully, the city search tool includes an `apigw_prefix` key which hints at the correct endpoint URL, so remember that when doing a city search and use it for future requests.

While I was in here, I refactored out the URL building code into `self._url_builder` to make the API calls more "obvious" and honestly line-wrap better :)

### DIFF
--- a/custom_components/waste_collection_schedule/manifest.json
+++ b/custom_components/waste_collection_schedule/manifest.json
@@ -15,5 +15,5 @@
     "pycryptodome",
     "pypdf"
   ],
-  "version": "2.9.0"
+  "version": "2.10.0"
 }


gregprosser:
So - this is mildly problematic.  It looks like we had prior support for specifying a region/district/zone in `configuration.yaml`, but if we don't look up the `region_prefix` we can't use these anymore. Right now, this introduces a **breaking change** which means that `region_prefix` will now be required alongside the other values. This fixes the test cases as it stands...

HOWEVER, if anyone knows a better way to go from project/district to query the city search API, that would make this a bit more bulletproof.  Querying the API via `curl` seems to suggest it can take a `code` parameter other than `term`, but since this API doesn't seem public I'm not sure what format `code` takes.

gregprosser:
I've now loaded this into my HA and I can confirm this fixes the source for me.

MomSpaghetti01:
Can confirm that this resolves the error, was able to add waste schedule from recyclecoach

Normanras:
Can confirm that the `us-web.apigw.recyclecoach.com` works for me, but I only looked at my region. Where did you find the city search tool with `apigw_prefix`? The RC app requires my state and town in the browser url.

gregprosser:
Honestly I haven't used the app and stupidly I only read `doc/source/recyclecoach_com.md` after I wrote the code.  I found out the "city search API" `https://api-city.recyclecoach.com/city/search?term=` returned `apigw_prefix` when it found a city.  For example, for Canada where I am the prefix is "ca" in the returned response.

rtorchia:
I've been attempting to getting things working again for my Canadian town (Lakeshore, Ontario). When I do a city search using the "city search API" it works. But attempting to get other info from the ca-web.apigw.recyclecoach.com resturns with {"message":"Forbidden"}.

Anyone else finding that all previous api calls no longer work, along with this new one?

gregprosser:
> I've been attempting to getting things working again for my Canadian town (Lakeshore, Ontario). When I do a city search using the "city search API" it works. 

1) Can you confirm this HACS extension with Recycle Coach worked before with your city?  When did it start failing, only recently?

2) Can you provide (maybe pick a random address nearby that's not a business) an address in this city?  Google Maps is showing most addresses I pick as being in Belle River and not living there it's hard to decipher what's really going on here.

3) What happens if you download the Recycle Coach app and try to set it up?  Are you able to see your data in the app?

> But attempting to get other info from the ca-web.apigw.recyclecoach.com resturns with {"message":"Forbidden"}.
> 
> Anyone else finding that all previous api calls no longer work, along with this new one?

I'm definitely not finding that - the source has test cases which were failing and now work post my changes.  I can't duplicate what you're saying (with Forbidden), if I add a test case for a (random guess - see point 1) address in Lakeshore, the search claims it succeeded but finds no results.

Mike5357:
Can also confirm this fixed my integration, alongside adding the region_prefix to my yaml

haimgel:
👍 Tested this, and it does bring the data back for me. Thank you.

rtorchia:
@gregprosser 
> 1. Can you confirm this HACS extension with Recycle Coach worked before with your city?  When did it start failing, only recently?

Not sure when it began to fail, but up to about last month it was working, so I'm not sure with which update it started failing. I was using HA configuration.yaml to set it up with this information:
```
# Waste Collection Schedule
waste_collection_schedule:
  sources:
    - name: recyclecoach_com
      args:
        street: {REMOVED}
        city: lakeshore
        state: ontario
        project_id: "583"
        district_id: LAK
        zone_id: zone-z9942
```

> 2. Can you provide (maybe pick a random address nearby that's not a business) an address in this city?  Google Maps is showing most addresses I pick as being in Belle River and not living there it's hard to decipher what's really going on here.

If you are using Google maps, it will return addresses based on postal codes from Canada Post, which don't align with municipalities. Belle River (along with many others) were all amalgamated into the Town of Lakeshore many years ago.

> 3. What happens if you download the Recycle Coach app and try to set it up?  Are you able to see your data in the app?

Using the Recycle Coach app, I select the Town of Lakeshore, and all the information is displayed correctly.

gregprosser:
> > 2. Can you provide (maybe pick a random address nearby that's not a business) an address in this city?  Google Maps is showing most addresses I pick as being in Belle River and not living there it's hard to decipher what's really going on here.
> 
> If you are using Google maps, it will return addresses based on postal codes from Canada Post, which don't align with municipalities. Belle River (along with many others) were all amalgamated into the Town of Lakeshore many years ago.

OK.  So I can pick any address in Belle River?  Can you provide an example residential address, perhaps of a neighbour or similar?  I am getting errors when I pick random addresses, and I want to pick one that is "known working".

gregprosser:
> @rtorchia
>
> ```
> # Waste Collection Schedule
> waste_collection_schedule:
>   sources:
>     - name: recyclecoach_com
>       args:
>         street: {REMOVED}
>         city: lakeshore
>         state: ontario
>         project_id: "583"
>         district_id: LAK
>         zone_id: zone-z9942
> ```

Can you try adding "region_prefix: ca" to your configuration and see if it works (with my PR)?

I added your configuration as a test case (FYI: "street" seems to be unnecessary) and I get results back.

I can't seem to do address lookups at all - but with your config (a zone, district, project) it seems totally OK (according to the unit test).  If you are still having problems, we probably need debug logs from this integration, with this PR attached - to figure out why you're failing!

5ila5:
As far as I can tell the only valid prefixes are `us` and `ca` so we could just keep the old logic of testing both

gregprosser:
> As far as I can tell the only valid prefixes are `us` and `ca` so we could just keep the old logic of testing both

The old logic would actually bail out completely if a host stopped existing or requests raised an exception - so it would also require a little bit of rework.  Are you opposed to using the region prefix they gave us?  In the 'standard' case of using the UI and doing address discovery it's automagically found - it's only people using `configuration.yaml` who have to change anything.

Without knowing recyclecoach's implementation, I feel like they intentionally split this to save load - some kind of sharding - so hitting both of them seems maybe counter to their wishes?

Plus if they do introduce, say, `eu` endpoints or something, this should handle it I hope.

lk229:
Added region_prefix: us and copied your recyclecoach_com.py to my system. That fixed the problem for me too. Thanks.

rtorchia:
I also copied the recyclecoach_com.py to my system, added region_prefix: ca, and rebooted. Still nothing.
So frome HA integrations I added Waste Collection Schedule. And selected Canada, Chatam-Kent (ON) (which uses Recycle Coach), and when it presented for filling in the details I changed everything to my town with the info I had in my configuration.yaml. Now everything is working like it was!

So yeah, this change is working for me now! Thank-you everyone!

dt215git:
Does the _recyclecoach_com.md_ file also need updating to include the _region_prefix_  arg?

5ila5:
> it's only people using `configuration.yaml` who have to change anything.

That's not quite correct the HA GUI configuration is basically the same as the configuration.yaml config you can do the Project ID configuration in the YAML and GUI.

image



I don't mind the `region_prefix`, I also wouldn't mind removing the Project ID stuff completely if users using the project_id approach need to touch their configuration anyway


It might be a good Idea to add some description to the input fields, so Users Know when they need to fill in which (e.g. PARAM_DESCRIPTIONS) : <https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md#update-links-in-readmemd-and-infomd>


an (admittedly hacky) way to not break existing configuration would be something like this:

```py
        for region_prefix in (
            ["ca", "us"] if not self.region_prefix else [self.region_prefix]
        ):
            schedule_url = self._url_builder(
                host=f"https://{region_prefix}-web.apigw.recyclecoach.com",
                url="/zone-setup/zone/schedules",
                params={
                    "project_id": self.project_id,
                    "district_id": self.district_id,
                    "zone_id": self.zone_id,
                },
            )
            response = requests.get(schedule_url)
            if "not exist in database" not in response.text:
                self.region_prefix = region_prefix
                break
```

gregprosser:
> That's not quite correct the HA GUI configuration is basically the same as the configuration.yaml config you can do the Project ID configuration in the YAML and GUI.

ah - fair point. I only meant that in my case (having completely configured via UI and adding my address) I required no changes to just have it start working again (didn't need to touch the region stuff, it "figured it out").  I admittedly don't know the proportion of users that are doing these, uh, "manual configurations", how many people do we figure do it this way?

> It might be a good Idea to add some description to the input fields, so Users Know when they need to fill in which (e.g. PARAM_DESCRIPTIONS) :

Will do in a commit in a moment.

> an (admittedly hacky) way to not break existing configuration would be something like this:

I'm not particularly interested in the hacky approach, but if it's blocking accepting the PR I'll change it to this - but I like the idea of using the provided API information, personally.  I'm not understanding the "existing configuration" that gets broken though - as far as I know the only people who will need to add anything are anyone who has manually added the zone/project/district/etc rather than using the lookup UI.  As I mentioned, for my use case (and seemingly others on the PR) if you've just used the UI flow to search for an address, no manual changes are necessary to just start working again.  But I could be misunderstanding other people's use cases.

5ila5:
I don't know the percentage, either. I honestly thought the street + city config option came later, and the first version just allowed the ID configuration. But I looked it up, and it was like this the whole time. If the street, city approach works fine, I don't mind just removing the ID arguments. But I can also merge it like this (maybe + added `PARAM_DESCRIPTIONS`).
It's a breaking change for ID configured sources anyway.

You should then remove the arguments from the configs by adding a migration: If you want, I can do it as well

https://github.com/mampfes/hacs_waste_collection_schedule/blob/946474b6fc35aa2c87ac0f7a39b2cdd01df73451/custom_components/waste_collection_schedule/init_ui.py#L99-L192

yapishkahilt:
> I've now loaded this into my HA and I can confirm this fixes the source for me.

Fixed for me as well. Appreciate it

abhi08638:
copied the py file over to my system and rebooted and it brought my config back online too. Huge thanks for this!

5ila5:
@gregprosser

What do you want to do about this, should I just add some migration code and merge this? Or do you want to do the migration or any other changes?

smenzer:
Any updates on this? Thanks!

bnevets27:
Don't have much to contribute, eagerly waiting for the fix to be published but it looks like #4600 is related and has a pull request as well.

5ila5:
I'll merge #4616 as this seem to work with very minor changes.

I'll leave this open as some changes might still be valuable if @gregprosser wants to integrate them into the version after the applied patch from #4616

PR #4342, opened 2025-07-03
Create st-ingbert_abfallkalender.md

djunkee:
Description on how to setup the waste collection calendar for St. Ingbert


5ila5:
There already is a mein_abfallkalender_online description I added st.ingebert to the extra infos so it apears in HA and in the README

just as information: you would have needed to create a yaml file which then generates the ics md file: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md

PR #4343, opened 2025-07-03
Update ics.md to include description for st. ingbert

djunkee:
added link to documentation for st. ingbert (separate pull request)


5ila5:
That does not work like this you need to create a yaml file everything else is handled by the update_docu_links.py script

see: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md

I'll close this for  #4349 as well


For potential next time. Stuff like this should probably be in one PR not split across multiple ones, You can just push in you existing branch and the PR gets updated

PR #4369, opened 2025-07-06
New source for Värmdo kommun Sweden

fatuuse:


custom_components/waste_collection_schedule/waste_collection_schedule/varmdo_se.py
```diff
@@ -0,0 +1,127 @@
...
+            waste_type = "Trash"
+            icon = "mdi:trash-can"
+            
+            while next_pickup_date.weekday() != next_pickup or (next_pickup_date.isocalendar()[1] % 2 != next_pickup_week):
+                next_pickup_date += timedelta(days=1)
+                #print("test date: "+ str(next_pickup_date.strftime('%Y-%m-%d')))
+                #print("test date - weekday: "+str(next_pickup_date.weekday()) +" week number: "+ str(next_pickup_date.isocalendar()[1]))
+                
+            next_pickup_date = next_pickup_date
+            #print("Found day: "+ str(next_pickup_date.strftime('%Y-%m-%d')))
```
5ila5: - `icon` is unused
- Please remove commented out code

custom_components/waste_collection_schedule/waste_collection_schedule/varmdo_se.py
```diff
@@ -0,0 +1,127 @@
...
+            entries.append(Collection(
+                    date=next_pickup_date,
+                    t=waste_type,
+                    icon=ICON_MAP.get(waste_type, "mdi:trash-can")
+                    ))
+            
+            # Contiune for next
+            next_pickup_date += timedelta(days=1)
```
5ila5: You could replace this with

```suggestion
        dates = rule = rrule(
            WEEKLY,
            dtstart=datetime.now(),
            byweekday=weekdays[next_pickup],
            byweekno=range(next_pickup_week+1, 54, 2),
            count=10
        )
        for date_ in dates:
            waste_type = "Trash"
            entries.append(Collection(
                    date=date_,
                    t=waste_type,
                    icon=ICON_MAP.get(waste_type, "mdi:trash-can")
                    ))
```

you need to import 


```py
from dateutil.rrule import rrule, WEEKLY, weekdays
from datetime import datetime
```

markvp (approved):

custom_components/waste_collection_schedule/waste_collection_schedule/varmdo_se.py
markvp: Addressed — removed unused `icon` variable, all commented-out code, and simplified the date generation.

custom_components/waste_collection_schedule/waste_collection_schedule/varmdo_se.py
markvp: Addressed — simplified to a loop with weekday and even/odd week matching. Didn't use rrule to avoid adding a dependency for this single source, but the logic is now much cleaner.

Issue #3805, opened 2025-02-18
[Source Request]: Clarence City Council


chrisnchips:
### Municipality / Region

Australia - Tasmania - Hobart - Clarence council

### Collection Calendar Webpage

https://www.ccc.tas.gov.au/living/waste-recycling/collection-days-areas/

### Example Address

https://assets.ccc.tas.gov.au/uploads/2024/10/Risdon-Vale.pdf

### Collection Data Format

As PDF Download

### Additional Information

https://assets.ccc.tas.gov.au/uploads/2024/10/Risdon-Vale.pdf

churchst28:
Second vote! Many thanks

churchst28:
Third vote :)

owenholloway:
Started work on this one, might have given the api a bit too much to think about for one night but will try again once it's woken up again, progress is here: https://github.com/owenholloway/hacs_waste_collection_schedule

owenholloway:
PR Submitted [New source for Clarence City Council, Tasmania, Australia](https://github.com/mampfes/hacs_waste_collection_schedule/pull/4377)

PeteLawler:
Nice work mate, the latest update in Release telling me I needed to restart HA reminded me something like this had been on my TODO list for way too long. Glad I looked at the pull requests!

PR #4377, opened 2025-07-07
New source for Clarence City Council, Tasmania, Australia

owenholloway:
Relevant issue: https://github.com/mampfes/hacs_waste_collection_schedule/issues/3805


5ila5:
Your result processing is overly complicated why don't you just do something like this:


```py

import re

DATE_REGEX = r"\d{1,2}/\d{1,2}/\d{4}"


def get_dates(value: str) -> list[date]:
    date_strings = re.findall(DATE_REGEX, value)
    dates = []
    if len(date_strings) == 2:
        date1 = datetime.strptime(date_strings[0], "%d/%m/%Y")
        date2 = datetime.strptime(date_strings[1], "%d/%m/%Y")
        dates.append(date1.date())
        dates.append(date2.date())
        diff = date2 - date1
        for _ in range(52):
            date2 += diff
            dates.append(date2.date())
    else:
        for date_str in date_strings:
            dates.append(datetime.strptime(date_str, "%d/%m/%Y").date())

    return dates


def _process_results(results) -> list[Collection]:
    entries = []

    collections: dict[str, list[date]] = {}

    for result in results:
        if result["name"] == "Hardwaste":
            collections["Hardwaste"] = get_dates(result["value"])

        if result["name"] == "Garbage Pickup":
            collections["Rubbish"] = get_dates(result["value"])

        if result["name"] == "Recycling Pickup":
            collections["Recycling"] = get_dates(result["value"])

        if result["name"] == "Green Waste Pickup":
            collections["Greenwaste"] = get_dates(result["value"])

    for bin_type, dates in collections.items():
        entries.extend(
            [
                Collection(
                    date=collection_date, t=bin_type, icon=ICON_MAP[bin_type.upper()]
                )
                for collection_date in dates
            ]
        )
    return entries
```


This would cover Hard waste as well, but I don't know if there is ever a useful date in there. It would also have some kind of "fault protection" if they change to only supply one date or more than two

custom_components/waste_collection_schedule/waste_collection_schedule/source/ccc_tas_gov_au.py
```diff
@@ -0,0 +1,164 @@
...
+        ccc_formatted_address = address_result['results'][0]
+
+        collection_search_data = {'address': ccc_formatted_address}
+        collection_url = "%s/collection-search" % URL
+        collections = session.post(collection_url, data=json.dumps(collection_search_data))
+
+        if address.status_code != 200:
+            raise Exception("Could not find collections")
+
+        collections_result = json.loads(collections.content)
```
5ila5: You can simply this by not using `json.` but the requests built in JSON parsing:

```suggestion
        address = session.post(address_url, json=address_search_data)

        if address.status_code != 200:
            raise Exception("Could not complete addres lookup %i" % address.status_code)
    
        address_result = address.json()

        if not address_result['success']:
            raise Exception("Could not find address")

        ccc_formatted_address = address_result['results'][0]

        collection_search_data = {'address': ccc_formatted_address}
        collection_url = "%s/collection-search" % URL
        collections = session.post(collection_url, json=collection_search_data)

        if address.status_code != 200:
            raise Exception("Could not find collections")

        collections_result = collections.json()
```

owenholloway:
Cleaned it up based on the recommendations, have excluded hard waste as it's never done

5ila5:
Thanks for your contribution

I did some reformatting (+minor type fixes like any is not a valid type hint but the built-in any function)
and run the  ./update_docu_links.py  script to update README.md info.md and the json files for HA config flow

owenholloway:
Thanks for your time taken on the review

Issue #2942, opened 2024-10-29
[Source Request]: NANCRAS / FRANCE


avidflyer17:
### Municipality / Region

NANCRAS / FRANCE

### Collection Calendar Webpage

https://cyclad.org/les-dechets/collecte/calendrier-de-collecte/

### Example Address

NANCRAS

### Collection Data Format

As html on a webpage, As PDF Download

melankh:
City list and their respective id : 

`curl -s  'https://cyclad.org/wp-json/vernalis/v1/communes'  | jq`

waste collection planning for a city id : example for **Surgères centre**, id **232**

`curl -s  'https://cyclad.org/wp/wp-admin/admin-ajax.php' -X POST --data-raw 'action=ajax_calendar_autocomplete&post_id=232' | jq`

PR #4394, opened 2025-07-10
Add Nancras source (and other city nearby)

avidflyer17:
https://github.com/mampfes/hacs_waste_collection_schedule/issues/2942


avidflyer17:
Ignore the translation part which is removing lots of lines ;)

PR #4404, opened 2025-07-11
fix: update kingston_vic_gov_au to use reusable WhatBinDay service

andykelk:
As of July 2025, the Kingston Council website returns a 403 when the API is called from this custom component.

Switch to using the whatbinday.com API which is used by the Binston Android app that Kingston Council have released.

This API requires that the client register a "device" before using the services endpoint. We cache this device key to avoid repeated registrations.


5ila5:
It looks like they support more than this region, Is there a good way to provide this to other sources as well. Maybe by adding most of the API logic into a `custom_components/waste_collection_schedule/waste_collection_schedule/service` script?

I don't like the random json file to be put in the custom_components installation directory. Either register a new device in the __init__ (which would mean to re-reigster on a HA restart) Or use the HA Store, But this would probably need some changes in the setup process of the integration

andykelk:
Still got work to do on this one to get the HA storage working properly.

PR #4433, opened 2025-07-15
Fix: doc where to execute pytest

mpw96:
The actual fix is to execute pytest in `hacs_waste_collection_schedule`, not in `waste_collection_schedule`.

Other than that I changed the formatting of the paragraph. In markdown documents, you usually start each sentence on a new line. This is easier when reviewing changes. The rendered document looks the same.


5ila5:
Thanks for your contribution

PR #4448, opened 2025-07-17
Update app_abfallplus_de.md

battle566:
Added new app_id for "Landkreis Schwäbisch Hall

Closes https://github.com/mampfes/hacs_waste_collection_schedule/issues/4431

PR #4449, opened 2025-07-17
Update lrasha_de.md

battle566:
To be removed as no longer supported. Alternativ e API is in place

Use alternative API "app_abfallplus_de with app_id "de.cmcitymedia.shawaste"
https://github.com/mampfes/hacs_waste_collection_schedule/edit/master/doc/source/app_abfallplus_de.md


steffenrapp:
I think this file `lrasha_de.md` should be removed completely and the `README.md` should be updated to link to `app_abfallplus_de.md` for "Landkreis Schwäbisch Hall" as it does for example for "Kreis Biberach". See this for reference:
Bildschirmfoto 2025-07-17 um 21 21 23
Then the link would directly refer to your change from #4448 

Also, in my opinion all of the obsolete source code should be removed as it's no longer needed with the switch to "Apps by Abfall+". You can find it here:
`custom_components/waste_collection_schedule/waste_collection_schedule/source/lrasha_de.py`

5ila5:
Thanks for your contribution

I just made extra clear that the other source is broken now, so the error message properly reflect it. We can then remove this source all together in the near future

PR #4457, opened 2025-07-20
fix: update kingston_vic_gov_au to use reusable WhatBinDay service

andykelk:
## Summary
- Switch Kingston source from broken Council API to working whatbinday.com API
- Extract reusable WhatBinDay service for other councils to use
- Add proper Home Assistant storage integration for device key caching
- Enable Home Assistant instance passing through source initialization chain

## Changes
1. **Extract reusable WhatBinDay service** - Creates a service other sources can use for whatbinday.com API integration
2. **Add HA instance passing** - Enables sources to access Home Assistant APIs for storage and configuration  
3. **Update Kingston source** - Switch to WhatBinDay service, reducing code from ~110 to ~40 lines

## Background
As of July 2025, the Kingston Council website returns a 403 when the API is called from this custom component. The new implementation uses the whatbinday.com API which is used by the Binston Android app that Kingston Council officially supports.

## Test plan
- [x] Verify Kingston source works with new API
- [x] Confirm device keys are properly cached using HA storage
- [x] Test fallback to file-based storage when HA storage unavailable
- [x] Ensure parameter format change doesn't break existing configs


5ila5:
I like this approach, but is there any reason you pass the hass instance around? I could not find a use of it in the source itself, or am I missing something?

Have you tested this? Looks good, but I should probably still test this before merging.


The check passes, as the environment for this test does not install the homeassistant package you'd need to add it to the requirements.txt file

andykelk:
> I like this approach, but is there any reason you pass the hass instance around? I could not find a use of it in the source itself, or am I missing something?

Thanks for the feedback! The `hass` parameter is used in the implementation - here's the flow:

  1. The Kingston source passes `hass=self._hass `to the `WhatBinDayService` (line 65 in `kingston_vic_gov_au.py`)
  2. The `WhatBinDay` service uses it via `_get_storage()` to access the Home Assistant storage system (lines 68-70 in `WhatBinDay.py`)
  3. The `DeviceKeyStore` requires the `hass` instance to create the Home Assistant `storage.Store` for persistent device key caching (lines 17-20 in `DeviceKeyStore.py`)

The purpose is device key persistence - when the WhatBinDay API registers a device, it receives a key that should be reused across Home Assistant restarts. Without the hass parameter, device keys would be lost on restart, causing unnecessary re-registrations with the API.

Let me know if you'd like me to add more comments to make this clearer!

> Have you tested this? Looks good, but I should probably still test this before merging.

Yes, I have tested this and it works correctly with device key persistence across HA restarts.

5ila5:
I see that the `DeviceKeyStore` needs hass for initilatation, but you only do this in `init_ui.py` the WhatBinDay just gets the already initialized `DeviceKeyStore` and does therefore not need the hass argument.

> The WhatBinDay service uses it via _get_storage() to access the Home Assistant storage system (lines 68-70 in WhatBinDay.py)

You call `get_device_key_store()` which does not require the hass instance


So the self.hass attribute of `kingston_vic_gov_au.py` is only used to pass it to `WhatBinDayService` where it's saved in `self.hass` but never used

andykelk:
> You call `get_device_key_store()` which does not require the hass instance
> 
> So the self.hass attribute of `kingston_vic_gov_au.py` is only used to pass it to `WhatBinDayService` where it's saved in `self.hass` but never used

Ah, gotcha. Yeah, you're right. I tried a couple of different ways to pass around before I got it working and missed cleaning it up. Have removed the redundant parameter passing and retested.

5ila5:
Thanks for your contribution

PR #4472, opened 2025-07-23
Fix: TBV Velbert, DE

pl4st3:
Updated API_URL

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/tbv_velbert_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/tbv_velbert_de.py
@@ -13,7 +13,8 @@ TEST_CASES = {
     "Rathaus": {"street": "Thomasstraße 1"},
 }
 
-API_URL = "https://www.tbv-velbert.de/abfall/abfalltrennung/abfallkalender"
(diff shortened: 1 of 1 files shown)


Marvthedark:
Hi, hab das gleiche Problem gehabt! 
Fix:
Ändere in der Datei "tbv_velbert_de" findet man i.d.R. unter: config\custom_components\waste_collection_schedule\waste_collection_schedule\source

Die Zeile: 
API_URL = "https://www.tbv-velbert.de/abfall/abfallkalender-und-abfuhrtermine/abfallabfuhr-suche"

in

API_URL = "https://www.tbv-velbert.de/abfall/abfallkalender"

Und schon funktionierts!

5ila5:
I don't think that's enough. Did you test this?

I needed to retrieve some more post arguments first so it works.

I'll push this change to this PR and merge it with the additional information.

pl4st3:
> I don't think that's enough. Did you test this?
> 
> I needed to retrieve some more post arguments first so it works.
> 
> I'll push this change to this PR and merge it with the additional information.

Tested it last week. But TBV Velbert changed their website again. Not sure when will release the final version. Raised a request to TBV Velbert to get more information.

PR #4480, opened 2025-07-24
Fixed URL for royalgreenwich

timocov:
https://www.royalgreenwich.gov.uk/info/200171/recycling_and_rubbish/2436/black_top_bin_collections returns 404 so fixing it to the correct one


5ila5:
Thanks for your contribution

PR #4429, opened 2025-07-15
Fix: Wiltshire, UK

dt215git:
Fixes #4425 
```bash
Testing source wiltshire_gov_uk ...
  found 31 entries for standard_uprn
    2025-07-04 : Household waste [mdi:trash-can]
    2025-07-18 : Household waste [mdi:trash-can]
    2025-07-10 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-07-24 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-08-01 : Household waste [mdi:trash-can]
    2025-08-15 : Household waste [mdi:trash-can]
    2025-08-29 : Household waste [mdi:trash-can]
    2025-08-07 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-08-21 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-09-12 : Household waste [mdi:trash-can]
    2025-09-26 : Household waste [mdi:trash-can]
    2025-09-04 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-09-18 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-10-10 : Household waste [mdi:trash-can]
    2025-10-24 : Household waste [mdi:trash-can]
    2025-10-02 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-10-16 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-10-30 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-11-07 : Household waste [mdi:trash-can]
    2025-11-21 : Household waste [mdi:trash-can]
    2025-11-13 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-11-27 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
    2025-12-05 : Household waste [mdi:trash-can]
    2025-12-19 : Household waste [mdi:trash-can]
    2025-12-11 : Mixed dry recycling (blue lidded bin) and glass (black box or basket) [mdi:recycle]
[...]

PR #4509, opened 2025-07-30
Update wiltshire_gov_uk.py

BeardyMike:
Updated to work with the Wiltshire Council's July 2025 update to this page.

(This is my first pull to this project, so it will probably need a good look over before it's accepted.)

I have a PHP project that pulls the same data, and I've applied/translated my own fixes to this projects file.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/wiltshire_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/wiltshire_gov_uk.py
@@ -1,4 +1,6 @@
 from datetime import date, datetime
+import json
+import re
 
 import requests
 from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


5ila5:
What does this actually improve?

It seems to work after just updating the URL (#4429)

BeardyMike:
> What does this actually improve?
> 
> It seems to work after just updating the URL (#4429)

I hadn't spotted this fix when I submitted mine. 

Mine is a port of a solution from one of my PHP projects, it doesn't require Beautiful Soup.

5ila5:
I don't actually think this is a big improvement but adding code and complexity. I don't think the regex search => json decode is more robust. 

So I'll close this.

BeardyMike:
> I don't actually think this is a big improvement but adding code and complexity. I don't think the regex search => json decode is more robust.
> 
> So I'll close this.

No argument from me :D

PR #4500, opened 2025-07-28
Update mijnafvalwijzer_nl.py

jankees0212:
The website has a new layout with both mobile and desktop elements. The change I made is only line 64 to select only the element of te mobile view.

PR #4515, opened 2025-07-31
Fix mijnafvalwijzer_nl for website changes

j0ran:
The website of mijnafvalwijzer.nl has changed a little bit. This pull request fixes that. 

1. Removes checking for a class that is not there anymore. 
2. Removes double entries from the mobile entries.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/mijnafvalwijzer_nl.py
@@ -61,7 +61,7 @@ class Source:
         r.raise_for_status()
 
         soup = BeautifulSoup(r.text, "html.parser")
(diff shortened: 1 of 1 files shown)


j0ran:
I think the change of @jankees0212 is better.

5ila5:
> I think the change of @jankees0212 is better.

Do they work for you? I get an issue with them, but your changes seem to work


(#4500)

j0ran:
> > I think the change of @jankees0212 is better.
> 
> Do they work for you? I get an issue with them, but your changes seem to work
> 
> (#4500)

I tested it, and it works for me. I appreciated @jankees0212's solution for its smaller changeset. However, after closer inspection, I believe my solution is slightly more robust due to targeting more specific elements. That said, web scrapers are inherently fragile and can break with any website change.

5ila5:
Duplicate of #4500

5ila5:
I merged @jankees0212 changes as they work now, Maybe short term outage yesterday when I tested the changes

Issue #4145, opened 2025-05-17
[Bug]: North Herts (UK) bin collection no longer updates, council website address has changed


Fish1-1:
### I Have A Problem With:

A specific source

### What's Your Problem

The North Herts bin collection web address has changed, the integration no longer updates the bin schedule as a result.

The new web address is : https://waste.nc.north-herts.gov.uk/w/webpage/find-bin-collection-day-input-address

I'm able to go this web address and get my bin collection information, the page to enter house address entry information has also changed format.

I hope someone is able to update the source request information for North Herts so this will start working again.

### Source (if relevant)

northherts_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
Waste Collection Schedule 2.7.0
Home Assistant Operating System 14.0
Home Assistant Core Update 2024.12.1
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

oliverclarke:
I've just looked (I noticed it'd stopped working too) at the parser and I think it might be a bit of a headache to fix. The new app that North Herts have used to host the collection dates (Netcall Liberty Create) doesn't render the full HTML so Beautiful Soup can't parse the content. 

This sucks for two reasons, 1. it makes the content much harder to scrape because you've got to run a headless browser library and associated web driver such as selenium or playwright, and 2. it shows that accessibility and automation are not prioritised. Their own [accessibility statement](https://www.north-herts.gov.uk/accessibility-statement#interactive-tools-and-transactions) admits that forms are hard to navigate, not well-labeled, and built by a third party with limited flexibility.

Without delving more into the code, I don't know if there is another way around this or if its been encountered with one of the other providers.

dt215git:
[...]

PR #4521, opened 2025-08-02
Fix North Hertfordshire, UK after council API change

Zepheus:
This pull request fixes issue #4145 

## Context
North Hertfordshire changed providers which broke the previous API integration.

The new portal is similar to `croydon_gov_uk.py` with a few minor adjustments.
I have kept the original parameters as to not to break existing configurations, although address search and matching works differently in this implementation.

## Test plan
```
$ python3 ./custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s northherts_gov_uk
Testing source northherts_gov_uk ...
  found 5 entries for Example
  found 5 entries for Example No Postcode Space
```

Ensured `pytest` still passes

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/northherts_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/northherts_gov_uk.py
@@ -1,5 +1,7 @@
 from datetime import datetime
 
+import re
+
 import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution.
This looks like it could break again very soon, as there are a lot of "magic" values that look they can change any time

sipuncher:
I've installed this in HACS (copied the full waste_collection_schedule) but it doesn't work.  When trying to add this source, whether I provided address input values or not I get the same error every time "The source returned an invalid response: "Expecting value: line 2 column 1 (char 1)". Please check the provided arguments and try again."  Is it still working for you?

5ila5:
This is broken again.

Probably caused by the "magic" variables throughout the code

@Zepheus can you look into retrieving them at runtime?

PR #4522, opened 2025-08-02
Fix: AWN ApplicationName for iCal download

scoop:
Some time this week the `awn_de` iCal download broke. It seems to be related to an upstream change of the `ApplicationName` parameter.


5ila5:
Thanks for your contribution

zimbo86:
can confirm that its broke after restarting my ha today..
Will test as soon as I find some time.

Edit: works, thank you.

PR #4535, opened 2025-08-03
Add `opole` app to the ecoharmonogram service

bogdal:
This PR adds a new app to the `Ecoharmonogram` service that contains the city of `Opole`.

ScreenShot 2025-08-03-13 32 46UTC@2x


5ila5:
Thanks for your contribution

PR #4598, opened 2025-08-17
Fix Lancaster (UK) source

WillFantom:
This fixes the Lancaster (UK) source by validating the html list of collections prior to processing (this has been erroring for a week do a minor change).

This also removes the `[i + 2]` indexers I put in there that, in hindsight, make me question my career and my wider existence.

(I'll be back in 2yrs to update this source to replace the parts that inevitably will embarrass me at that time)


5ila5:
Thanks for your contribution

PR #4599, opened 2025-08-17
Fix KAEV Niederlausitz [DE] source

aschobba:
fix(source): Handle invalid timezone data for KAEV Niederlausitz

The ICS files provided by the KAEV Niederlausitz source contain an empty VTIMEZONE block, which leads to an `IndexError` during parsing.

This commit implements a workaround that checks for the faulty data before processing and sanitizes it if necessary.


5ila5:
Thanks for your contribution

PR #4604, opened 2025-08-18
Adding Nårab - Sweden

### LINKED ISSUE #1243: Request for adding Nårab (Sweden)
I was thinking building something of my own, but there is no easy API or anything that seems to be public.
I think you have to parse the website generated at each request to find the dates for each collection.
Any help would be much appreciated! :)

https://www.narabtomningskalender.se/
Search for e.g., my address: "Snyggatorpsvägen 1"

When incpecting the network traffic, this is the page being loaded:
https://www.narabtomningskalender.se/basfiler/online_kalender_skapa.php?hsG=Snyggatorpsv%C3%A4gen%201&hsO=KLIPPAN&knR=31811&abK=VILLOR&nrA=10012108007&lang=sv&clid=e97828682dc80c2b36df990778fb41a1

GET variables used in the query:
* hsG: Street name and number (yes, with spaces...)
* hsO: City
* knR: Customer number (as seen on the agreement/invoice, or just by searching for an address and inspect the traffic...)

It might also be nescessary to have a paremeter for the GET variable abK, which have at least the following options:
* VILLOR (e.g., your own house)
* FLERFAMILJ (e.g., apartment building)
* VERKSAMHET (e.g., company)

Although, probably VILLOR is the only necessary option as you don't care much when the collection in aparment buildings is.

victordiges:
Adding source for Norra Åsbo Renhållnings AB (Nårab) servicing the southern region of Skåne in Sweden.

It parses a generated calendar provided by Nårab for each customer. No ICS or JSON source are available so the generated html needs to be parsed.

The kundNr (customer number) is sometimes necessary, in case the address search returns multiple results. For example searching for "Streetname 1" will also return "Streetname 11", "Streetname 12" and so on. This argument is a bit tricky to obtain and it needs _some_ computer knowledge as it requires the user to find it using the "Inspect" tools in Chrome or Firefox. Alternatively a wizard script is provided to allow the user to find his address almost the same way the web UI does.

Closes #1243

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/narab_se.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/narab_se.py
@@ -1,12 +1,12 @@
 import datetime
+
+import requests
+from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection
 from waste_collection_schedule.exceptions import (
     SourceArgumentException,
     SourceArgumentNotFound,
 )
-import requests
-from bs4 import BeautifulSoup
-import re
 
 # Title will show up in README.md and info.md
 TITLE = "Nårab - Norra Åsbo Renhållnings AB"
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

LucasHokerberg:
Thank you so much! Much appreciated. 🙂

victordiges:
you're welcome!

I just saw in my invoice that the customer number is also there, so it may be easier for some people to obtain it that way. I may do a PR to update the documentation

PR #4624, opened 2025-08-23
Fixed Hume City Council

seedzero:
Hume City Council changed to new API

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/hume_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/hume_vic_gov_au.py
@@ -1,4 +1,3 @@
-import datetime
 import json
 import re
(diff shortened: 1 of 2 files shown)


5ila5:
The old API seems to still work, and your change would break all existing configurations.
I guess they will discontinue the old one sometime?

Is there a way to provide more seamless transition, maybe writing a migration?

seedzero:
Strange that the old API is working for you, it's not working on my end. Good point about the migration, I sometimes forget that I'm not the only one using this. I'll have another look.

seedzero:
@5ila5 changed the args back to same format as used with the old API. Config should be largely backwards compatible now.

seedzero:
@5ila5 could you please have another look at this?

5ila5:
LGTM
Thanks for your contribution

I did some reformatting  + fixed 2 typos in the md file

Issue #4620, opened 2025-08-21
[Bug]: Cardinia, Victoria, Australia provider error


vytux-com:
### I Have A Problem With:

A specific source

### What's Your Problem

Cardinia provider seems to have changed / broken something.

Looking at their website it would seems like the moved geocoder lookups to a different site e.g.

```
https://geocoder.cit.api.here.com/6.2/search.json?gen=9&app_id=pYZXmzEqjmR2DG66DRIr&app_code=T-Z-VT6e6I7IXGuqBfF_vQ&country=AUS&state=VIC&searchtext=16+Main+St%2C+Pakenham+Vic&bbox=-37.86%2C145.36%3B-38.34%2C145.78
``` 
is now something like 
```
https://corp-geo.mapshare.vic.gov.au/arcgis/rest/services/Geocoder/VMAddressEZIAdd/GeocodeServer/findAddressCandidates?magicKey=GST7YMc0AM9UOsKtGTyVGST7YMc0AM9UOsExAi9XOc50YTc2KQWmObktGMytaikZQDoKUD8GQNFF&SingleLine=16+Main+Street+PAKENHAM+3810&f=json
```

### Source (if relevant)

cardinia_vic_gov_au

### Logs

```Shell
This error originated from a custom integration.

Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
integration: Waste Collection Schedule (documentation)
First occurred: 14:18:01 (1 occurrence)
Last logged: 14:18:01

[...]

PR #4625, opened 2025-08-23
Update cardinia_vic_gov_au.py

seedzero:
New "Magic Key"


seedzero:
Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/4620

5ila5:
This is not sustainable, the old one was added less than 2 months ago, this source will probably break very soon again. We probably need to extract the magic key from their website

I'll merge this for now, but I guess this will break again very soon and we should develop another way to dynamically get this key

PR #4641, opened 2025-08-28
Add Fuquay Varina, NC source

RyanJamesCaldwell:
This PR adds Fuquay Varina, North Carolina, USA as a source for waste collections, which pulls the data from a town ArcGIS server.


custom_components/waste_collection_schedule/waste_collection_schedule/source/fuquay_varina_nc_us.py
```diff
@@ -0,0 +1,123 @@
...
+            "outFields": "garbage_next_pickup_date,recycling_next_pickup_date"
+        }
+
+        response = requests.get(API_URL, params=params)
+        response.raise_for_status()
+        
+        data = response.json()
+        
+        if not data.get("features"):
+            return []
```
5ila5: do not return []

raise an actual exception

RyanJamesCaldwell: Fixed in 64a25572039625fff7398e851b93effd3bd7629d. I had a merge conflict with iTouchVision changes, so I rebased back onto latest master and squashed the change in with the original source commit.

5ila5:
Thanks for your contribution

I get a Cloudflare error message, but this might be due to geo-blocking issues. I guess this works on your side.

I did some reformatting (and fixed the `-> Optional[datetime.date]` type hint as this is `datetime.datetime.date` (due to import) which is a function not the `datetime.date` class)).

PR #4686, opened 2025-09-05
New Sources: Yarra City Council (VIC, AUS) + Newcastle City Council (NSW, AUS)

stumcd:
Adds Yarra City Council (VIC, AUS).

- Weekly: Rubbish + FOGO
- Fortnightly: Recycling (from 2025-07-04), Glass (from 2025-07-11)
- Arg: `zone` (1–10)
- Holiday shifts: 26 Dec 2025 → 27 Dec, 3 Apr 2026 → 4 Apr

## Example config
```yaml
waste_collection_schedule:
  sources:
    - name: yarracity_vic_gov_au
      args:
        zone: 10
```

## References
- https://www.yarracity.vic.gov.au/residents/bins-waste-recycling-and-cleansing/recycling-and-rubbish/bin-collection
- Zone 10 collection calendar PDF

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/newcastle_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/newcastle_nsw_gov_au.py
@@ -1,0 +1,109 @@
+from datetime import date, timedelta
+from waste_collection_schedule import Collection
+
+TITLE = "City of Newcastle (NSW)"
+DESCRIPTION = "Newcastle (NSW) waste schedule – weekly rubbish; recycling and organics alternate on the collection day."
+URL = "https://newcastle.nsw.gov.au/living/waste-and-recycling/collection-services/bin-collection-days"
+TAGS = ["nsw", "newcastle", "australia"]
+
+ICON_MAP = {
+    "Rubbish": "mdi:trash-can",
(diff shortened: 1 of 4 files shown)


5ila5:
I don't like these sources, they basically just re-implment the [static source](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/static.md). I'd at least use the functionality from this source, But I don't think these sources serve much of a purpose.

We reverse engineered the agris map for some sources but that's not the same everywhere and pretty complicated. Maybe parsing the PDF would be a possibility. Or just sticking with the static source

markvp:
Hi @stumcd, thanks for the contribution! The maintainers have noted that static/hardcoded schedule approaches essentially replicate the existing `static` source and aren't suitable for inclusion as standalone sources. For these councils, the ideal approach would be to reverse-engineer their API or parse their website/PDFs. Since this PR has been open since Sep 2025 with no path forward on the approach, we'd suggest closing it. Please feel free to open a new issue or PR if you find a dynamic data source for either council in the future!

markvp:
Closing as the implementation approach isn't viable for inclusion (see prior review feedback) and there's been no further activity. Happy to reopen if you find a dynamic data source (API/ICS/scrapable site) — please open a fresh issue or PR with that approach.

PR #4696, opened 2025-09-08
fix: clamp days to generate

GingerSnap-xx:
We did not check if it is 0 or lower so the default 0 value in the UI was passing the None check.

Also this adds a business that's operating in a historic house that has a residential address so we can have a valid Id in the repo but its not violating anyone's PII (as far as I can tell)


GingerSnap-xx:
Thanks to @byrdman42 for reporting this issue

5ila5:
Thanks for your contribution

I set the actual default value, so our HA config flow fills in 3 instead of 0 as the default value

PR #4700, opened 2025-09-09
chore: add cincinnati support

tylergets:


custom_components/waste_collection_schedule/waste_collection_schedule/source/cincinnati_gov.py
```diff
@@ -0,0 +1,91 @@
...
+ICON_MAP = {
+    "Trash": "mdi:trash-can",
+    "Recycling": "mdi:recycle",
+    "Street Sweeping": "mdi:road-variant",
+}
+
+# ### Arguments affecting the configuration GUI ####
+
+HOW_TO_GET_ARGUMENTS_DESCRIPTION = {  # Optional dictionary to describe how to get the arguments, will be shown in the GUI configuration form above the input fields, does not need to be translated in all languages
+    "en": "Find your address ID by searching for your address on Cincinnati's service website. The address ID is needed to retrieve collection schedules.",
```
5ila5: maybe explain that you can find the id in the URL after searching for the address

custom_components/waste_collection_schedule/waste_collection_schedule/source/cincinnati_gov.py
```diff
@@ -0,0 +1,91 @@
...
+        response.raise_for_status()
+
+        # Parse JSON response
+        data = response.json()
+
+        entries = []
+
+        # Check if the response has the expected structure
+        if "Services" not in data:
+            return entries
```
5ila5: Please raise an exception as this will show up in the logs while this just fails silently

custom_components/waste_collection_schedule/waste_collection_schedule/source/cincinnati_gov.py
```diff
@@ -0,0 +1,91 @@
+from datetime import datetime, timedelta
+
+import requests
+from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+
+TITLE = "City of Cincinnati, OH"
+DESCRIPTION = "City of Cincinnati, OH, USA"
+URL = "https://www.cincinnati.gov/"
+COUNTRY = "us"
+TEST_CASES = {}
```
5ila5: please add test cases

custom_components/waste_collection_schedule/waste_collection_schedule/source/cincinnati_gov.py
```diff
@@ -0,0 +1,91 @@
...
+            for service_date in service_dates:
+                start_date_str = service_date.get("startDate")
+                if not start_date_str:
+                    continue
+
+                # Parse the date string (format: "Wednesday, September 17, 2025")
+                try:
+                    date_obj = datetime.strptime(start_date_str, "%A, %B %d, %Y").date()
+                    # Subtract one day from the collection date
+                    date_obj = date_obj - timedelta(days=1)
```
5ila5: Why do you subtract 1 is this some timezone thing or do they just report the wrong date?

5ila5 (changes requested):
Please add a md file and don't update the json files manually but run the update_docu_links.py to update all json and MD files


more infos: <https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md>

custom_components/waste_collection_schedule/waste_collection_schedule/source/cincinnati_gov.py
5ila5: Looks like I'm geo-blocked so I don't know how exactly this looks

bbr111:
@tylergets 

please update your PR

`commit merge upstream/master'
run 'python3 update_docu_links.py'

Check changes from 5ilas5

if this is not relevant anymore, please close PR.

markvp:
@tylergets I've pushed fixes addressing all of @5ila5's review feedback:

- Added doc file and test case
- Raised `SourceArgumentNotFound` instead of silent return when no services found
- Improved the HOW_TO_GET_ARGUMENTS_DESCRIPTION with a direct URL
- Removed the `-1` day subtraction (couldn't find justification for it — if dates are actually off by one, please let us know)
- Ran linting and update_docu_links.py

Could you please **validate the fix works for your address**? In particular, confirm that the collection dates are correct without the day subtraction. The API is geo-blocked outside the US so we can't test from here.

markvp:
Closing as no further activity since fixes were pushed, and the Cincinnati API is geo-blocked from outside the US so we can't self-validate. Happy to reopen if a US-based contributor can confirm the dates are correct for their address.

PR #4729, opened 2025-09-17
Add Czyste Miasto Gdańsk source

### LINKED ISSUE #4262: [Source Request]: Gdańsk, Poland
### Municipality / Region

Gdańsk, Poland

https://czystemiasto.gdansk.pl/harmonogram-odbioru-odpadow/

### Example Address

_No response_

### Collection Data Format

Dedicated IOS/Android APP, As PDF Download, As html on a webpage

### Additional Information

[kalendarz.pdf](https://github.com/user-attachments/files/20746274/kalendarz.pdf)

dulek:
Closes #4262


5ila5:
Gdańsk is already supported by the `ecoharmonogram_pl` source. You even call an ecoharmonogram API

dulek:
Ha, I just tested it and you're right. :) Though I see that 2 categories are missing from `ecoharmonogram_pl` source - "ODPADY ZIELONE" (green waste) and "TERMINY PŁATNOŚCI" (payment date) which are available in the Czyste Miasto Gdańsk app.

Anyway your call if you want to include this or assume it's redundant. Stuff returned from `ecoharmonogram_pl` is actually enough for my own purposes.

5ila5:
I don't think adding a additional source is not a good idea if we call the API of the same service provider. We could add extra info section to whe ecoharmogram_pl so the supported regions show up in the README and HA config flow. But there are a ton of regions and I think it would clutter the interface more than it would help.

If you want you could try to figure out why the https://ecoharmonogram.pl/api/api.php does not contain the green waste.
But I'll close this for now

dulek:
> I don't think adding a additional source is not a good idea if we call the API of the same service provider. We could add extra info section to whe ecoharmogram_pl so the supported regions show up in the README and HA config flow. But there are a ton of regions and I think it would clutter the interface more than it would help.
> 
> If you want you could try to figure out why the https://ecoharmonogram.pl/api/api.php does not contain the green waste. But I'll close this for now

Ouch, only now I realized this is actually calling exactly the same API. So yeah, good call to drop this, thanks!

PR #4742, opened 2025-09-22
Update bodenseekreis_de.md due to changes on bodenseekreis.de

snowstoked:
Changed the description and URL due to new implementation generating the ical file on bodenseekreis.de since 19.09.2025


5ila5:
Thanks for your contribution. The MD file is actually generated by the YAML file, so your changes would've been overwritten. I applied these changes in the YAML file (and modified the last line a bit as the old one does not make much sense when displayed in the HA GUI and I try to remove this line from all sources one by one)

PR #4743, opened 2025-09-22
Add street filter for Frankenberg DE

hnrmrl:
If a street is selected, then the filter shall be applied to the iCal generator


5ila5:
Thanks for your contribution

PR #4752, opened 2025-09-24
Fixing #4738

### LINKED ISSUE #4738: [Bug]: Wokingham recycling broken
### I Have A Problem With:

A specific source

### What's Your Problem

The council have added (week 1) or (week 2) after recycling, which has broken that bin source.

Changing line 131 in the wokingham_gov_uk.py source file to:

`            waste_type = card.find("h3").text.split("(")[0].strip()`

Fixes it, but I’m not going to be near a pc where I can submit a PR in the next week, so just putting this here for now, in case anyone needs it, and I can try and get a PR done next week, if no one else does it in the meantime.

### Source (if relevant)

wokingham_gov_uk

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

Rujith:
Fix #4738. 
Implementing the solution suggested in the bug


badguy99:
Thanks for picking this up @Rujith

PR #4766, opened 2025-09-29
Added new collection source for Woollahra NSW, Australia

sh00t2kill:
Added new collection source for Woollahra NSW, Australia


sh00t2kill:
Apologies for the multiple commits.

I missed a couple of files due to line endings being screwed up in VSCode. Looks to be sorted now!

custom_components/waste_collection_schedule/waste_collection_schedule/source/woollahra_nsw_gov_au.py
```diff
@@ -0,0 +1,268 @@
...
+        if june_date_match:
+            try:
+                day = june_date_match.group(1)
+                year = june_date_match.group(2)
+                date_str = f"{day} June {year}"
+                return datetime.datetime.strptime(date_str, DATE_FORMAT_LONG).date()
+            except ValueError:
+                pass
+                
+        return None
```
5ila5: You can replace the whole function with just
```py
        try:
            return dateutil.parser.parse(date_text, dayfirst=True, fuzzy=True).date()
        except (dateutil.parser.ParserError, OverflowError) as e:
            return None
```

and `import dateutil` at the top. That's probably even more robust

custom_components/waste_collection_schedule/waste_collection_schedule/source/woollahra_nsw_gov_au.py
```diff
@@ -0,0 +1,268 @@
...
+                else:
+                    raise Exception(f"Failed to fetch waste services: {r.status_code} (attempt {attempt + 1})")
+                    
+            except requests.RequestException as e:
+                if attempt == max_retries - 1:
+                    raise Exception(f"Network error during waste services fetch: {str(e)}")
+                time.sleep(2 ** attempt)
+
+        if r.status_code != 200:
+            raise Exception(f"Unable to access Woollahra waste services API (status: {r.status_code})")
```
5ila5: I think it would be good to move this to it's own function as you use this twice and is quite a lot of code for "Make this request with retries and some error handling"
maybe
```py

  @staticmethod
  def _request_with_retries(
      uri: str, headers: dict[str, str], session: requests.Session
  ) -> requests.Response:
    ...
```

bbr111:
@sh00t2kill

please update your PR

`commit merge upstream/master'
run 'python3 update_docu_links.py'

Check changes from 5ilas5

if this is not relevant anymore, please close PR.

sh00t2kill:
Fixed up as per comments. Things got a bit hairly with line endings on VSCode but I got there eventually1

custom_components/waste_collection_schedule/waste_collection_schedule/source/woollahra_nsw_gov_au.py
```diff
@@ -0,0 +1,214 @@
...
+                    continue
+                else:
+                    raise Exception(f"Failed to fetch: {r.status_code} (attempt {attempt + 1})")
+                    
+            except requests.RequestException as e:
+                if attempt == max_retries - 1:
+                    raise Exception(f"Network error: {str(e)}")
+                time.sleep(2 ** attempt)
+        
+        return None
```
bbr111: You should not return None.


```suggestion
        raise requests.exceptions.RequestException(
            "Max retries exceeded while fetching URL"
        )
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/woollahra_nsw_gov_au.py
```diff
@@ -0,0 +1,214 @@
...
+                    )
+                )
+
+        return entries
+
+    def _parse_date(self, date_text: str) -> datetime.date:
+        """Parse various date formats found in the response"""
+        try:
+            return dateutil.parser.parse(date_text, dayfirst=True, fuzzy=True).date()
+        except (dateutil.parser.ParserError, OverflowError) as e:
```
bbr111, line 214:
```suggestion
    def _parse_date(self, date_text: str) -> datetime.date|None:
```

bbr111, line 214: import re is unused

bbr111, line 214: If _make_request_with_retry() returns None, r will be None here and the following r.status_code access will fail.
The method should either always raise or we need to guard against None.

bbr111:
thanks for you contribution

PR #4774, opened 2025-10-04
Update eko_tom_pl.py

kielich89:
Now works for eko tom


5ila5:
Thanks for your contribution

PR #4798, opened 2025-10-11
[JC] Add Source Wandsworth Borough Council

JackCurtis96:
Adds waste collection schedule source for Wandsworth London Borough. I believe I have added everything I need. If anything maybe I need more test cases?
Cheers


custom_components/waste_collection_schedule/waste_collection_schedule/source/wandsworth_gov_uk.py
```diff
@@ -0,0 +1,74 @@
...
+    "Small electrical items": "mdi:battery",
+}
+
+
+class Source:
+    def __init__(self, uprn):
+        self._uprn = uprn
+
+    def fetch(self):
+        r = requests.get(API_URL.format(uprn=self._uprn))
```
5ila5, line 31: You don't format the URL at all.

I'd also prefer to use the `requests.get` `params` argument to pass the get arguments:
```py
API_URL = "https://www.wandsworth.gov.uk/my-property/"
...
r = requests.get(API_URL, params={"UPRN":self._uprn, "propertyidentified": "Select"})
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/wandsworth_gov_uk.py
```diff
@@ -0,0 +1,74 @@
...
+                        break
+
+            if next_div:
+                date_text = next_div.get_text(
+                    strip=True).replace("Next:", "").strip()
+
+                try:
+                    parsed_date = parser.parse(date_text).date()
+                except Exception:
+                    parsed_date = date_text
```
5ila5, line 65: Using the string as date WILL break stuff. Either get the date some other way or use `continue`

markvp:
Closing — Wandsworth was already added via #5471.

PR #4802, opened 2025-10-11
Added Komunala Kranj (Slovenia) to garbage collection sources

TKosir:
Added Komunala Kranj (Slovenia) to garbage collection sources.


5ila5:
Thanks for your contribution

I did some minor reformatting and run the update_docu_links script to update MD and JSON files.

I also "fixed" the timezone handling for python <=3.10 as our CI pipeline still uses an older python version (we should probably update as HA only supports the latest two python versions)

TKosir:
Thanks for awesome hass addon, and correcting the code!

PR #4824, opened 2025-10-16
Fix derby_gov_uk after website changes

Azelphur:
Hello,

Derby seems to have made some changes to their bin collection page, which broke this integration. This PR fixes it, so it now works again.

Derby no longer supports lookup by postcode / house number. I'd need the post code AND the exact address (including post code) this seemed a bit redundant, and address/postcode lookup was already not recommended. I decided to remove the ability to use search, now you must provide premises ID (which can be easily obtained from the URL when searching using a web browser.)

So, my changes:

- Corrected the URL for retrieval, they added a `s` on the end of `BinDay`.
- Removed the ability to search for address, user must now provide premises ID.
- The church seems to have gone, so I added a new test address and removed the church.
- I noticed I was initiating a requests.session but not using it, now I am.
- I raise_for_status to account for any errors the website might throw.

### DIFF
--- a/custom_components/waste_collection_schedule/translations/de.json
+++ b/custom_components/waste_collection_schedule/translations/de.json
@@ -15090,7 +15090,7 @@
       },
       "args_derby_gov_uk": {
         "title": "Quelle konfigurieren",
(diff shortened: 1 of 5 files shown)


dansollok:
Fixes #4828 #4829

5ila5:
Thanks for your contribution

I added more information, showing in the config flow

Issue #3696, opened 2025-01-27
[Bug]: Wrong region for service provider ZKRI


blurap:
### I Have A Problem With:

A specific source

### What's Your Problem

I had successfully set up the ‘Goldau’ region via the service provider ZKRI / Source sammelkalender.ch. The data was correct until a few weeks ago. However, a few weeks ago I suddenly started receiving data for the second region available for this service provider, ‘Arth (incl. Oberarth)’. I have not changed anything in the configuration. The configuration shown in the ‘Relevant Configuration’ section worked and still works in principle, but with the dates from the wrong region. I have already spent some time trying to find the problem, but with no success. What do I have to enter to get the dates from the ‘Goldau’ region again?
Thank you very much in advance for any support.

### Source (if relevant)

sammelkalender.ch / ZKRI

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
# Konfiguration Abfallkalender
waste_collection_schedule:
  sources:
  - name: sammelkalender_ch
    args:
      service_provider: zkri
      municipality: Arth
      street: Goldau
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
fixed by #4832 

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do "update" of WCS until the next release, as pressing the update button performs a "downgrade" to the latest release version

PR #4832, opened 2025-10-19
Fixed faulty usage of dummy streets response from API

JonasArnold:
The ZKRI API returns a dummy street (show below) if the parameter `fuerStr=1` is set. Since the plugin always first searches for streets, and the dummy street includes a default `SAGEid` this default is always used and the SAGE (Sammelgebiet) can effectively not be chosen.

Dummy street returned by ZKRI API for the following parameters:
https://daten.zkri.ch/web/Sammelkalender/kunden_WANN_auswahl_form.php?optGem=33&fuerStr=1&Jahr=2025

```
[
  {
    "STRid": null,
    "STRname": null,
    "STRhausnr": null,
    "SAGESTRid": null,
    "SAGEid": 189
  }
]
```

This fixes issue #3696.


5ila5:
Thanks for your contribution

PR #4840, opened 2025-10-19
Fix(valorlux_lu): Update API endpoint and parsing logic

koosoli:
- Updates the API endpoint to the new URL.
- Corrects the data parsing logic to handle the new data structure.
- Adds "Parc Hosingen" to the test cases.

PR #4845, opened 2025-10-20
Fix/valorlux scraper

koosoli:


5ila5:
What's that?

I already merged #4840

Don't add test scripts for any sources, just use the existing test script e.g.

`./custom_components/waste_collection_schedule/waste_collection_schedule/test/test_sources.py -s valorlux_lu -lit`

bbr111:
@koosoli 

please respond.
Should we go forward, or close PR?

koosoli:
Hi, sorry for replying this late. For m,e the last merge already fixed it. So this here is just double dipping and redunant.

PR #4855, opened 2025-10-22
muenchenstein_ch: Fixed date parsing

r3turnNull:
Source changed where they store the date for collection events. Code has been fixed to use the new source format and parse dates correctly.


5ila5:
Thanks for your contribution

PR #4243, opened 2025-06-12
fix: add new api source

puralpha:
This adds a new API source since original source now has CAPTCHA/browser integrity check page and fixes #4208

The [API ](https://whichb.in/) is a project by [Will Murphy](https://willmurphy.co.uk/) (not me)

Testing source eastleigh_gov_uk ...
  found 5 entries for 100060319000
    2025-06-16 : Recycling Bin [mdi:recycle]
    2025-06-16 : Food Waste Bin [mdi:food]
    2025-06-16 : Glass Box and Batteries [mdi:bottle-soda]
    2025-06-20 : Garden Waste Bin [mdi:leaf]
    2025-06-23 : Household Waste Bin [mdi:trash-can]
  found 3 entries for 100060300958
    2025-06-18 : Recycling Bin [mdi:recycle]
    2025-06-18 : Glass Box and Batteries [mdi:bottle-soda]
    2025-06-19 : Garden Waste Bin [mdi:leaf]
  found 3 entries for 100060306006
    2025-06-19 : Recycling Bin [mdi:recycle]
    2025-06-19 : Glass Box and Batteries [mdi:bottle-soda]
    2025-06-23 : Garden Waste Bin [mdi:leaf]
  found 1 entries for 10009587286
    2025-06-19 : Household Waste Bin [mdi:trash-can]

PR #4858, opened 2025-10-23
Fix: eastleigh_gov_uk

dansollok:
Fixes: #4853

Bypassed cloudflare bot check with browser headers and also it will only work if we add www. to the url (thanks ali-cato)

The response from the server is brotli compressed so we need that library to be installed on the client.
Is manifest.json the right place for that @5ila5  ?

Also... I see there are an old pull request #4243 still open with a completely different approach using an external source.


custom_components/waste_collection_schedule/waste_collection_schedule/source/eastleigh_gov_uk.py
```diff
@@ -23,18 +23,35 @@
...
 class Source:
     def __init__(self, uprn: str | int):
         self._uprn: str | int = uprn
 
     def fetch(self):
+        s = requests.Session()
+        s.headers.update({
+            "User-Agent": "Mozilla/5.0 (Linux; Android 6.0; Nexus 5 Build/MRA58N) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/141.0.0.0 Mobile Safari/537.36",
+            "Accept": "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7",
+            "Accept-Encoding": "gzip, deflate, br, zstd",
```
5ila5, line 38: I'd prefer to just remove `br` here
it works for me, and we don't need additional dependencies

bbr111:
@dansollok 

Please reply.

markvp:
This fix is already present in master — `eastleigh_gov_uk.py` was updated to use `curl_cffi` with `impersonate="chrome124"` which handles the Cloudflare bypass properly. The manual browser headers and `brotli` dependency in this PR are no longer needed. Closing as resolved.

PR #4860, opened 2025-10-23
Fix SSL certificate verification notificaiton for gardenbags_co_nz

PMKA:
This PR fixes the InsecureRequestWarning that appears in Home Assistant logs when using the GardenBags NZ source. The fix follows the same pattern used by other sources in this integration (awg_de, basingstoke_gov_uk, etc.) that handle SSL certificate verification issues in Home Assistant environments.

`Testing source gardenbags_co_nz ...
  found 0 entries for Test 1`


5ila5:
Thanks for your contribution

PR #4867, opened 2025-10-27
Fix Arun District Council

whi-tw:
Arun are inconsitent between postcodes about whether they show Rubbish collection dates on Recycling weeks. This change ensures that if Recycling is shown without Rubbish, Rubbish is also added for the same day.

I can't find any postcodes that don't follow this pattern. If there are, we could do a week by week check, then pull which day of the week the rubbish collection on the schedule normally is, then add in an event on that day. maybe.

Also, fix an annoying variable name


5ila5:
Thanks for your contribution

Issue #2619, opened 2024-09-05
West Oxfordshire Source seems to have broken?


mattsgreen:
Both myself and another user using the same (West Oxfordshire) source have started seeing issues this week.  I see the below in the logs: 

`fetch failed for source West Oxfordshire District Council: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 145, in fetch entries = self._source.fetch() ^^^^^^^^^^^^^^^^^^^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/westoxon_gov_uk.py", line 235, in fetch serialized_state = r.json()["actions"][0]["returnValue"]["response"][ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^ TypeError: 'NoneType' object is not subscriptable`

_Originally posted by @mattsgreen in https://github.com/mampfes/hacs_waste_collection_schedule/issues/818#issuecomment-2331118178_

domdomegg:
This seems to be failing at `westoxon_gov_uk.py:235`. This is because there is an error response back from the server for the request made on line 234:

<details><summary>Error response</summary>

```json
{
    "actions": [
        {
            "id": "123;a",
            "state": "ERROR",
            "returnValue": null,
            "error": [
                {
                    "exceptionEvent": true,
                    "useDefault": true,
                    "event": {
                        "descriptor": "markup://aura:serverActionError",
                        "attributes": {
                            "values": {
                                "error": {
                                    "message": "Interview state is not supplied",
                                    "stackTrace": "",
                                    "id": "-944885593"
                                }
                            }
                        }
                    }
                }
            ]
        }
    ],
    "context": {
        "mode": "PROD",
        "app": "siteforce:communityApp",
        "contextPath": "/s/sfsites",
        "pathPrefix": "",
        "fwuid": "eGx3MHlRT1lEMUpQaWVxbGRUM1h0Z2hZX25NdHFVdGpDN3BnWlROY1ZGT3cyNTAuOC40LTYuNC41",
        "mlr": 1,
        "coos": 1,
        "loaded": {
            "APPLICATION@markup://siteforce:communityApp": "vgD8vvaBHzgKYqb_JQjQdw",
            "COMPONENT@markup://flowruntime:flowRuntimeForFlexiPage": "6vLCX6RcjM4U8N2_kygrQw",
[...]

PR #4870, opened 2025-10-28
Fix westoxon

jamesweale:
Fixes: #2619

I reviewed the network log when looking for an address and noticed the flowDevName in one of the request headers had changed, updating it to 'WebFormAlloyWasteCollectionEnquiry' seems to fix the issue.

I also updated the icon map as I noticed the bin types didn't match for the garden bin and food caddy.

Here is the test run result:
image


5ila5:
Thanks for your contribution

PR #4878, opened 2025-10-30
Recyclecoach: Adds steps to retrieve district_id, project_id, and zone_id

maxfarrior:
The goal was to add the steps one needed to take in order to determine the user's district_id, project_id, and zone_id.

The current directions do mention one way to get the district_id and project_id, but do not specifically explain how to get the zone_id.


5ila5:
Thanks for improving the documentation

PR #4890, opened 2025-11-02
Fix Herefordshire script

greytuk:
Herefordshire Council changed the bin day page structure. The legacy HTML ids wasteCollectionDates, altnextWasteDay and altnextRecyclingDay no longer exist. The page now renders future dates inside the #binCollectionDetails block as <li> list items.

What this PR does

Updates selector to target #binCollectionDetails instead of the removed #wasteCollectionDates

Extracts the "next" collection dates from the first <li> in each matching section

Keeps all other logic intact (UPRN lookup unchanged, existing API URL unchanged)

Minimal code change — just fixes parsing to match new HTML

Impact

This restores working collection date retrieval for Herefordshire Council users.


5ila5:
Thanks for your contribution,

I made some minor changes to return all collection  not just one

PR #4903, opened 2025-11-05
buergerportal_de: adapt to API field change (Volumina/VolumenObj)

### LINKED ISSUE #4888: [Bug]: buergerportal_de doesn't work since today
### I Have A Problem With:

The integration in general, A specific source

### What's Your Problem

Since today buergerportal_de doesn't work any more. 

https://buerger-portal-albdonaukreisabfallwirtschaft.azurewebsites.net/calendar is working.

Does anyone els have a problem like this or maybe it is a general problem?

### Source (if relevant)

buergerportal_de

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: buergerportal_de
      args:
        operator: alb_donau
        district: XXXXX
        street: XXX
        number: X
      calendar_title: Müllkalender
```

### Checklist Source Error

### Checklist Sensor Error

### Required

jlai79:
The provider changed the field name.
Fixes #4888.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/buergerportal_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/buergerportal_de.py
@@ -150,6 +150,37 @@ class Source:
         self.street = street
         self.number = number
         self.show_volume = show_volume
+        self.new_params = True
+
+    def _do_fetch_request(
+        self,
+        session: requests.Session,
+        district_id: int,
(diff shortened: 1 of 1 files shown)


5ila5:
Thanks for your contribution

That's not completely correct, the first and third test case fail with these changes while the second one fails without them

I modified your code to try both and remember for the next fetch

PR #4905, opened 2025-11-05
Fix support for North Hertfordshire, UK

Zepheus:
North Herts schedule was previously scraped through the council website, which broke multiple times.
There was a recent introduction of an app which uses a proper API which this pull request introduces.

I have kept the old address formatting for backwards compatibility, but had to introduce a fuzzy matching approach to resolve addresses to internal API identifiers (UPRNs).

Tested through
```
test_sources.py -s northherts_gov_uk -i -l
```


grahamhar:
I've added this to my install and works perfectly ❤️

5ila5:
Thanks for your contribution

PR #4930, opened 2025-11-11
Update zva_wmk_de.py

Demel75:
Year String for 2026 again changed without any logic


5ila5:
Thanks for your contribution

Issue #3742, opened 2025-02-05
[Source Request]: North Yorkshire Council - Ryedale, UK


gw72:
### Municipality / Region

North Yorkshire Council - Ryedale, UK

### Collection Calendar Webpage

https://www.northyorks.gov.uk/bin-calendar/lookup

### Example Address

https://www.northyorks.gov.uk/bin-calendar/Ryedale/results/10007635405

### Collection Data Format

Something different (add to additional information)

### Additional Information

Identical to what you already have for other North Yorkshire Councils such as Scarborough, UK

5ila5:
added in #4937 

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

PR #4937, opened 2025-11-13
Add support North Yorkshire Ryedale

tapayne88:
Resolves https://github.com/mampfes/hacs_waste_collection_schedule/issues/3742


5ila5:
As a lot of these northyorks sources seem to be nearly identical, I would like to merge them into one (providing EXTRA_INFO, so they keep listed seperatly in the README.md and HA config GUI) Bit I think this is fine for now

Thanks for your contributeion

Issue #4881, opened 2025-10-31
[Bug]: Newark and Sherwood council showing blank calender


Hatton920:
### I Have A Problem With:

A specific source

### What's Your Problem

This has been working for a long time and not sure when it went funny but the Newark and Sherwood council source now shows nothing.  My calendar has today as the last bin collection and nothing else.

I have tried removing and re-entering the source details with the same issue.

### Source (if relevant)

newark_sherwooddc_gov_uk

### Logs

```Shell
no relevent logs, cant see any errors
```

### Relevant Configuration

```YAML
Latest home assistant OS on a n100 mini pc
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

Hatton920:
Actually I see the issue.....

The page https://app.newark-sherwooddc.gov.uk/bincollection/calendar?pid=100031451269 shows this year up until today and next years is on https://app.newark-sherwooddc.gov.uk/bincollection/Calendar?pid=100031451269&nc=1

Not sure how to proceed with this now

Hatton920:
ok temp work around

added 

`params={"pid": self._uprn, "nc": "1"},`

Too the source file for now, works as expected for now

Hatton920:
The website has reverted back to how it normally was, removed my work around and all is good

StreamingStar:
They flip between the 2 pages. nc1: is now showing next year from November 2026.  I have the intergation look at both pages.

StreamingStar:
```
import calendar
import datetime
import re

import requests
from bs4 import BeautifulSoup
from waste_collection_schedule import Collection  # type: ignore[attr-defined]

TITLE = "Newark & Sherwood District Council"
DESCRIPTION = "Source for Newark & Sherwood services."
[...]

PR #4940, opened 2025-11-13
Add second request with nc=1 to fetch full Newark & Sherwood bin data

Hatton920:
Newark & Sherwood District Council returns incomplete bin collection data unless the 'nc=1' parameter is added to the request.

This update performs a second fetch using:
  payload = {'pid': uprn, 'nc': '1'}

and parses both responses to capture all available months.

This ensures all bin types and dates are collected reliably.

Credit to StreamingStar #4881 for identifying the required request parameter.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/newark_sherwooddc_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/newark_sherwooddc_gov_uk.py
@@ -28,10 +28,16 @@ class Source:
         self._uprn = uprn
 
     def fetch(self):
-        # get json file
+        entries = self.get_data({"pid": self._uprn})
+        try:
+            entries += self.get_data({"pid": self._uprn, "nc": "1"})
+        except Exception:
+            pass
+        return entries
+
+    def get_data(self, payload):
         r = requests.get(
-            "http://app.newark-sherwooddc.gov.uk/bincollection/calendar",
-            params={"pid": self._uprn},
+            "http://app.newark-sherwooddc.gov.uk/bincollection/calendar", params=payload
         )
 
         entries = []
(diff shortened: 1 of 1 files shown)


5ila5:
Looks like you never used git / opened a PR

You cannot just put all your changes at the bottom of the file.

Do you want to figure out how to fix this PR or should I?

Hatton920:
> Looks like you never used git / opened a PR
> 
> You cannot just put all your changes at the bottom of the file.
> 
> Do you want to figure out how to fix this PR or should I?

Your right I have never used github or opened a PR before, however the changes to the file are added at the end as the source needs to look at two URLs to get the required information.  Any help would be appreciated. Thank you

5ila5:
> however the changes to the file are added at the end as the source needs to look at two URLs to get the required information

Ah, then your changes were technically correct. 

I really don't like this kind of code duplication, so I moved the requests into its own function (you can see the changes in the `Files Changed` tab)

I don't think this change would be necessary, as the first approach already retrieved a lot of collection dates, but maybe this helps someone, so I'll merge this.

Putting `#################### START OF EDIT credit to StreamingStar ######################` comments in your changed file is still not a good idea. One can see changes of a file from within git (git diff) or here on GitHub on the `Files Canged` section or by clicking on a commit.

Hatton920:
> > however the changes to the file are added at the end as the source needs to look at two URLs to get the required information
> 
> Ah, then your changes were technically correct.
> 
> I really don't like this kind of code duplication, so I moved the requests into its own function (you can see the changes in the `Files Changed` tab)
> 
> I don't think this change would be necessary, as the first approach already retrieved a lot of collection dates, but maybe this helps someone, so I'll merge this.
> 
> Putting `#################### START OF EDIT credit to StreamingStar ######################` comments in your changed file is still not a good idea. One can see changes of a file from within git (git diff) or here on GitHub on the `Files Canged` section or by clicking on a commit.

Thank you.  The source URL changes towards the end of the year to account for a new calendar year, without the changes for around a couple of weeks the calendar will appear empty.  With the changes it looks at both and keeps the data current and recent.

https://github.com/mampfes/hacs_waste_collection_schedule/issues/4881#issuecomment-3522203848

I wanted to give credit to the person who actually coded the fix, apologies if this was not to be put there.

PR #4944, opened 2025-11-14
Update alchenstorf_ch.py

1337hium:
Fix for the new Calendar and strange formatting of Dates

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/alchenstorf_ch.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/alchenstorf_ch.py
@@ -31,38 +31,37 @@ class Source:
 
         table = html.find("table", attrs={"id": "icmsTable-abfallsammlung"})
         if not table:
-            return []
+            raise ValueError("Could not find the waste collection table on the page")
 
         try:
             data = json.loads(table.attrs["data-entities"])
         except (json.JSONDecodeError, KeyError):
-            return []
+            raise ValueError("Could not parse the waste collection data from the page")
 
         entries = []
         for item in data.get("data", []):
             try:
                 waste_type = BeautifulSoup(item["name"], "html.parser").text.strip()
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/alchenstorf_ch.py
```diff
@@ -19,35 +18,69 @@
...
     def fetch(self):
         response = requests.get("https://www.alchenstorf.ch/abfalldaten")
+        response.raise_for_status()
 
         html = BeautifulSoup(response.text, "html.parser")
 
         table = html.find("table", attrs={"id": "icmsTable-abfallsammlung"})
-        data = json.loads(table.attrs["data-entities"])
+        if not table:
+            return []
```
5ila5: raise an exception here so something shows up in the logs and it does not just fail silently

custom_components/waste_collection_schedule/waste_collection_schedule/source/alchenstorf_ch.py
```diff
@@ -19,35 +18,69 @@
...
+
+        try:
+            data = json.loads(table.attrs["data-entities"])
+        except (json.JSONDecodeError, KeyError):
+            return []
```
5ila5: same

custom_components/waste_collection_schedule/waste_collection_schedule/source/alchenstorf_ch.py
```diff
@@ -19,35 +18,69 @@
...
+                    entries.append(
+                        Collection(
+                            date=pickup_date,
+                            t=waste_type,
+                            icon=icon,
+                        )
+                    )
+
+            except Exception:
+                pass
```
5ila5: Keeping the logger from before the change and putting a warning here might be helpful

5ila5:
Thanks for your contribution

Thanks for fixing the remaining issues @bbr111

PR #4957, opened 2025-11-19
Edited class CitiesAppsV2 to add subcalendars

tst-Me:
This change fixes handling of **CitiesApps v2** waste-management areas that include an addressNumbers array (used for per‑household variants such as "Einfamilienhäuser" and "Mehrfamilienhäuser"). Previously the integration only exposed the parent area (e.g. "Judendorf‑Straßengel") so users could not select the specific variant and ended up receiving combined data.

### What this PR does

Expands addressNumbers into child entries (each retains its original _id) and adds a readable displayName of the form "ParentName - AddressNumber" so variants are selectable.
Adds normalization and tolerant matching for calendar names (Unicode normalization, ß -> ss, remove punctuation, collapse whitespace) to accept inputs like "Judendorf-Straßengel Einfamilienhäuser" or "Judendorf‑Straßengel - Einfamilienhäuser".
Uses the child addressNumbers._id when calling GET /waste-management/areas/{area_id}/calendar so the correct calendar is retrieved.
Keeps autocomplete as first choice and falls back to full listing when needed (empty query or 400).
Why this is needed CitiesApps now models household-type variants as addressNumbers inside area objects (type area_with_subareas). Without exposing these, the integration cannot let users select the correct plan for single‑family vs multi‑family houses.

### Additionally

Added a normalization function for special characters, often used in german langugage. And the fix preserves existing behaviour for areas without addressNumbers and uses a conservative fallback strategy (autocomplete first, full list fallback). 

### Testing

I tested locally against Gratwein‑Straßengel: entries like "Judendorf‑Straßengel - Einfamilienhäuser" now appear and fetching the calendar returns the correct garbageCollectionDays.
A syntax check was performed with python -m py_compile on the modified file.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/CitiesAppsCom.py
@@ -1,6 +1,7 @@
 import json
 import re
 import time
+import unicodedata
 import urllib.parse
 
 import requests
@@ -1764,8 +1765,6 @@ class CitiesApps:
             return self.get_garbage_plans(specific_calendar)
 
         def _normalize(self, s: str) -> str:
-            import unicodedata
-            import re
             if not s:
                 return ""
             s = unicodedata.normalize("NFKD", s)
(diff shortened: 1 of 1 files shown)


5ila5:
LGTM

Thanks for your contribution

PR #4966, opened 2025-11-20
Add London Borough of Havering

SamirHafez:
Adding source for [London Borough of Havering](https://www.havering.gov.uk/) in London, UK.

- [x]  Added `custom_components/waste_collection_schedule/waste_collection_schedule/source/havering_gov_uk.py`
- [x]  Added `doc/source/havering_gov_uk.md`
- [x]  Updated `README.md` via `update_docu_links.py`
- [x]  Updated `info.md` via `update_docu_links.py`
- [x]  Tested with `test_sources.py`
```bash
$ ./test_sources.py -s havering_gov_uk
Testing source havering_gov_uk ...
  found 14 entries for 53 Argyle Gardens
```


5ila5:
Thanks for your contribution

I did some reformatting and used get with a default on `COLLECTION_MAP` so the source does not compleatly fail if they rename the collection types for some reason

PR #4969, opened 2025-11-21
Fix month check and update year string formats for 2026

respecreload:
deleted years 2021 to 2024. changed the link for year 2026 to new url and also did same thing for 2025 so that the current year also works. also changed the month, where the next year is visable to 11 so that the added dates can be seen now.


respecreload:
this is already been approved by another user, so I closed it.

PR #4990, opened 2025-11-28
Fix and update Renfrewshire Council Provider

JamiePhonic:
This provider has been broken for a while Renfrewhire Council has been updating their website and, unusually for a governmental agency, has actually made things better in the process.

You now only need to perform a GET to https://www.renfrewshire.gov.uk/bins-and-recycling/bin-collection/bin-collection-calendar/check-your-bin-collection-day/view/{uprn} to get your bin schedules, so no more faffing with forms!

Conveniently, the nice graphics on the page are dynamically generated from a JSON object embedded in the page, which makes scraping and parsing it a breeze!

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/renfrewshire_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/renfrewshire_gov_uk.py
@@ -1,9 +1,8 @@
-from urllib.parse import parse_qs, urlparse
-
-import requests, json
+import json
 from datetime import datetime
+
+import requests
 from bs4 import BeautifulSoup
-from dateutil import parser
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
 TITLE = "Renfrewshire Council"
@@ -24,6 +23,7 @@ ICON_MAP = {
     "Blue": "mdi:note",
 }
 
+
 class Source:
     def __init__(self, postcode, uprn):
         self._postcode = postcode
@@ -39,7 +39,9 @@ class Source:
         r.raise_for_status()
 
         soup = BeautifulSoup(r.text, features="html.parser")
-        collections_data = soup.find("script", {"type": "application/json", "id": "collections-data"})
+        collections_data = soup.find(
+            "script", {"type": "application/json", "id": "collections-data"}
+        )
 
         entries = []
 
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/renfrewshire_gov_uk.py
```diff
@@ -23,89 +24,48 @@
...
-        for sched_entry in schedule:
-            entries.append(
-                Collection(
-                    date=sched_entry[0],
-                    t=sched_entry[1],
-                    icon=ICON_MAP.get(sched_entry[1]),
-                )
-            )
+
+        if collections_data:
```
5ila5: If this does not exist raise an exception so it does not fail silently

custom_components/waste_collection_schedule/waste_collection_schedule/source/renfrewshire_gov_uk.py
```diff
@@ -23,89 +24,48 @@
...
+            try:
+                # Get the text content and parse JSON
+                binData = json.loads(collections_data.string.strip())
+            except json.JSONDecodeError as e:
+                print("JSON Decode Failed with: " + e)
+            finally:
```
5ila5: The finally does not make sens if the JSON decode fails continuing does not make any sense, but you do anyway.

custom_components/waste_collection_schedule/waste_collection_schedule/source/renfrewshire_gov_uk.py
```diff
@@ -23,89 +24,48 @@
...
-                    icon=ICON_MAP.get(sched_entry[1]),
-                )
-            )
+
+        if collections_data:
+            try:
+                # Get the text content and parse JSON
+                binData = json.loads(collections_data.string.strip())
+            except json.JSONDecodeError as e:
+                print("JSON Decode Failed with: " + e)
```
5ila5: if this happens further processing does not make sense so just raise and exception here:

```py
raise Exception("JSON Decode Failed with: " + str(e)) from e
```

Your `+ e` would produce an error anyway

JamiePhonic:
Hi @5ila5 

Thnks for your feedback. I have made some changes that i believe address your concerns.

5ila5:
Thanks for your contribution

PR #5022, opened 2025-12-03
Fix Peterborough UK and use ICS

### LINKED ISSUE #4731: [Bug]: Peterborough City Council Error
### I Have A Problem With:

A specific source

### What's Your Problem

When I am trying to set up the configuation I am getting the error.

The source returned an invalid response: "'NoneType' object has no attribute 'text'". Please check the provided arguments and try again.

I have tried the multiple addess but its not returning a result

### Source (if relevant)

peterborough_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

skipishere:
Fixes #4731

Peterborough Council have changed the way you can access the bin collection information, the previous code accessed a hidden API which has now been removed. There is now a dedicated ICS file people can access which this change now uses.

The new way requires users to set the Postcode and the UPRN, given this has been broken for a while now I think users should be ok with having to update their YAML config (if they were previously only relying on address look up, which is also broken on the Peterborough website)

### DIFF
--- a/custom_components/waste_collection_schedule/init_ui.py
+++ b/custom_components/waste_collection_schedule/init_ui.py
@@ -270,6 +270,20 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry) ->
                 # postcode arg no longer needed, so delete it
                 del new_data["args"]["postcode"]
 
+        if config_entry.version < 2 or (
+            config_entry.version == 2 and config_entry.minor_version < 10
+        ):
+            # Migrate from birmingham_gov_uk to roundlookup_uk
+            if new_data.get("name", "") == "peterborough_gov_uk":
+                _LOGGER.debug("Migrating from peterborough_gov_uk to ics configuration")
+
+                new_data["args"]["post_code"] = new_data["args"].get("post_code")
+                new_data["args"]["post_code"] = new_data["args"].get("uprn")
+                if "number" in new_data["args"]:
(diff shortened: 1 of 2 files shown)


5ila5:
Thanks for your contribution

I wrote a migration for those who had UPRN and Postcode configured on the old source. I doubt that are a lot of people but nonetheless.

custom_components/waste_collection_schedule/init_ui.py
```diff
@@ -270,6 +270,20 @@ async def async_migrate_entry(hass: HomeAssistant, config_entry: ConfigEntry) ->
...
 
+        if config_entry.version < 2 or (
+            config_entry.version == 2 and config_entry.minor_version < 10
+        ):
+            # Migrate from birmingham_gov_uk to roundlookup_uk
+            if new_data.get("name", "") == "peterborough_gov_uk":
+                _LOGGER.debug("Migrating from peterborough_gov_uk to ics configuration")
+
+                new_data["args"]["post_code"] = new_data["args"].get("post_code")
+                new_data["args"]["post_code"] = new_data["args"].get("uprn")
```
skipishere, line 281: Should this line have been ```new_data["args"]["uprn"] = new_data["args"].get("uprn")``` ?

5ila5:
Oh, sorry, you're right I missed that

I fixed this in #5143

PR #5027, opened 2025-12-03
Changed user agent of awb_es_de to make downloads of ics work again

meilon:
The download of the ICS files stopped working today, it seems the previous user agent got blocked. Changing it to something different worked for me

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py
@@ -13,7 +13,7 @@ TEST_CASES = {
     "Kohlberg": {"city": "Kohlberg", "street": "alle Straßen"},
 }
 
-HEADERS = {"user-agent": "Mozilla/5.0 (xxxx Windows NT 10.0; Win64; x64)"}
+HEADERS = {"user-agent": "Mozilla/5.0 (X11; Linux x86_64; rv:60.0) Gecko/20100101 Firefox/81.0"}
 
 
 class Source:


Paddy0174:
This PR doesn't do good, as it only changes _one_ source. What about all the others, where the user agent needs to be changed as well?

meilon:
I would welcome a centrally managed UA that sources could use if needed. But I don't know if that would make things easier or harder. Maybe use a library like `user_agent` to generate a fresh one on the fly?

Paddy0174:
I don't think, the user agent is the problem. Unfortunately.

If I read the code right, the problem is, that practically all users try to get a new download at the same time, with the same user agent. To a normal sys admin that might look suspicious, more like a DDoS attack. It's 03:00h in the morning, and suddenly a flood of connections pops up, all at the same time, all the same user agents, but different IP addresses. If I where abfall.io, I'd scan for exactly the same thing and block it. 

That's why I don't think, changing the user agent alone will solve this problem. I'd even go so far, to implement a new user agent, so abfall.io does see what is going on. And we would need some kind of timing window, so not all users come together at the same time of the night.

meilon:
That's why a library generating a random UA on the fly would solve that issue. I have `fetch_time` and `random_fetch_time_offset` configured, if this would be enforced for everyone, that would also help. Best would be to use the time of the last HA boot, that should be fairly distributed. A UA would have to be updated regularly, still.

5ila5:
We have cloudscraper as dependency now. You can use cloudscraper as a drop in for requests or just use the fresh user agent (`user_agent = cloudscraper.create_scraper().headers['User-Agent']`)

I don't think people change, `random_fetch_time_offset` so the request should be spread out through an hour.

5ila5:
this seems to work:

```diff
diff --git a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py
index c6082dd9..1b676260 100644
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/awb_es_de.py
@@ -1,8 +1,8 @@
-import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
 from waste_collection_schedule.service.ICS import ICS
+import cloudscraper
 
 TITLE = "Abfallwirtschaftsbetrieb Esslingen"
 DESCRIPTION = "Source for AWB Esslingen, Germany"
@@ -13,8 +13,6 @@ TEST_CASES = {
     "Kohlberg": {"city": "Kohlberg", "street": "alle Straßen"},
 }
 
-HEADERS = {"user-agent": "Mozilla/5.0 (xxxx Windows NT 10.0; Win64; x64)"}
-
 
 class Source:
     def __init__(self, city, street=None):
@@ -28,7 +26,7 @@ class Source:
             "parent": "",
             "kind": "removaldate.city",
         }
-        r = requests.post(
+        r = self._session.post(
             "https://www.awb-es.de/statics/abfallplus/search.json.php", data=data
         )
         r.raise_for_status()
@@ -49,7 +47,7 @@ class Source:
             "parent": self._city,
             "kind": "removaldate.street",
         }
-        r = requests.post(
+        r = self._session.post(
             "https://www.awb-es.de/statics/abfallplus/search.json.php", data=data
         )
         r.raise_for_status()
@@ -64,14 +62,14 @@ class Source:
         )
 
     def fetch(self):
-        session = requests.Session()
+        self._session = cloudscraper.create_scraper()
 
         params = {
             "city": self._city,
             "street": self._street,
             "direct": "true",
         }
-        r = session.get(
+        r = self._session.get(
             "https://www.awb-es.de/abfuhr/abfuhrtermine/__Abfuhrtermine.html",
             params=params,
         )
@@ -95,7 +93,7 @@ class Source:
         entries = []
         for ics_url in ics_urls:
             # get ics file
-            r = session.get(ics_url, headers=HEADERS)
+            r = self._session.get(ics_url)
             r.raise_for_status()
 
             # parse ics file
```

Issue #5023, opened 2025-12-03
[Bug]: gfa_lueneburg_de does not work correctly at the end of the year due to the "new season"


CommSter:
### I Have A Problem With:

A specific source

### What's Your Problem

At the end of the year they have two instances running, e.g. one for 2025 and one for 2026.
On their website there is a radio group during tht period for selecting the correct year. The default is always the upcoming year (2026) which is why the next scheduled date for a specifc type of waste is in e.g. 35 days instead of 2.
It would be helpful, if the integration should check if the radiogroup is present and then selecting the "old" year until new years eve.

https://gfa-lueneburg.de/service/abfuhrkalender.html

<img width="889" height="636" alt="Image" src="https://github.com/user-attachments/assets/ff30fe15-83e5-4abb-96ba-41bb04455075" />

### Source (if relevant)

gfa_lueneburg_de

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

bsourdeau:
I was running into the same issue which always leads to the situation that at the end of the year once they publish the calendar for the upcoming year, only the dates for the next year are properly shown in HA.
The data for the last dates in the current year is missing though as explained by @CommSter above.

5ila5:
fixed in #5031

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

Marcuss28:
I installed the latest version.

[...]

PR #5031, opened 2025-12-04
Fixes concurrent years in gfa_lueneburg_de

CommSter:
Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/5023

At the end of each year, GFA Lüneburg (https://gfa-lueneburg.de/service/abfuhrkalender.html
) runs two calendar instances in parallel, e.g., one for the current year and one for the upcoming year.
During this period, their website displays a radio group allowing users to select which year’s waste collection schedule should be returned. The default selection is always the upcoming year (e.g., 2026).

Because of this, the integration incorrectly retrieves only dates from the next year — resulting in the next scheduled pickup being shown as e.g. 35 days away instead of 2 days.

This fix adds the missing payload parameter Zeitraum, filling it dynamically with the current year.
The expected format is:
"Jahresübersicht {current_year}" (e.g., "Jahresübersicht 2025").

This ensures the correct year is always selected and the integration returns the proper upcoming pickup dates.


5ila5:
Thanks for your contribution

PR #5036, opened 2025-12-05
Add Oak Bay (BC) to waste collection sources

MartyTremblay:


5ila5:
That's not how this works, you should not modify the JSON files by hand but add your changes to the recyclecoach_com.py EXTRA_INFO section and run the update_docu_links.py script, otherwise your changes get just overwritten.

https://github.com/mampfes/hacs_waste_collection_schedule/blob/721bab4d892e87064869ac5648f12dabb3bb3854/custom_components/waste_collection_schedule/waste_collection_schedule/source/recyclecoach_com.py#L72-L159

PR #5047, opened 2025-12-07
Add eva_abfallentsorgung_de

bbr111:
Add Support for EVA Abfallentsorgung.
Remove ICS md file


5ila5:
What do you mean by "Remove ICS md file"?

There already is the ICS source which is a lot simpler and does not require as many web requests.

The documentation currently states `Remove the year argument from the url` if this does cause issues, it might be better to advise replacing the year with `{%Y}`

PR #5052, opened 2025-12-08
Fixed greenwich borough holidays parsing

timocov:
https://www.royalgreenwich.gov.uk/recycling-and-rubbish/bins-and-collections/bank-holiday-collection-dates now has `Collection as usual` value meaning “to is the same as from”.


5ila5:
Thanks for your contribution

PR #5058, opened 2025-12-08
Added glass collection to rushcliffe_gov_uk

ribz:
Rushcliffe have now added a 4th bin (glass).

I've added an entry to ICON_MAP and adjusted the parsing code to only return "glass" rather than "glass (purple lid)", which is how the council currently refer to the bin on the website.


5ila5:
Thanks for your contribution

PR #5076, opened 2025-12-13
fix(hobartcity_com_au): Skip entries with empty dates

wmcleanutas:
## Problem
The `hobartcity_com_au` source fails with error:
```
time data '' does not match format '%a %d/%m/%Y'
```

## Cause
The Hobart City Council API now returns a 4th article for "Collection zone: Wednesday (Week 1)" which contains a calendar download link but has an **empty date** in the `div.next-service` element:

```html
<article>
  <h3>Collection zone: Wednesday (Week 1)</h3>
  <div class="service-details">
    <div class="note">...</div>
    <div class="next-service">
      
    </div>
  </div>
</article>
```

## Fix
- Skip articles with empty dates in `get_collections()`
- Update `TEST_CASES` addresses to match current API format:
  - Removed "Tasmania" (API no longer accepts it)
  - Fixed double spaces in addresses


5ila5:
Thanks for your contribution

PR #5078, opened 2025-12-14
Update logan_qld_gov_au API

KuzonCode:
Logan has moved to a [new portal](https://www.logan.qld.gov.au/home#myproperty). The [old one](https://az-prod-wa02.azurewebsites.net/) no longer works. They are also moving to a new backend URL.

I have updated Logan's code to accept the new address format while still allowing back-compat with the old address format. 

Unfortunately, the new API doesn't have the Formatted Address property hence the need for the change. Updated the documentation to reflect this change.


5ila5:
Thanks for your contribution

I reformatted the whole file (.pre-commit-config.yaml) even though it was already wrongly formatted before your changes.

I also raised an issue instead of returning `[]` when the backup failed. As the old approach would just fail silently while the Exception will show up in the logs

PR #5083, opened 2025-12-14
feat(midsussex_gov_uk): Add replacement source script and documentation

Gdkkauowj33:
### Test Output
Testing source midsussex_gov_uk ...
  found 10 entries for Test Case 1: 23 High Street
  found 10 entries for Test Case 2: Hapstead Hall
  found 10 entries for Test Case 3: The Gardeners Arms

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py
@@ -1,154 +1,162 @@
-import re
-from datetime import datetime
-
+import datetime
 import requests
 from bs4 import BeautifulSoup
-from waste_collection_schedule import Collection  # type: ignore[attr-defined]
-from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
+from waste_collection_schedule import Collection
+from waste_collection_schedule.exceptions import SourceArgumentException
 
-TITLE = "Mid-Sussex District Council"
-DESCRIPTION = (
-    "Source for midsussex.gov.uk services for Mid-Sussex District Council, UK."
-)
-URL = "https://midsussex.gov.uk"
+# --- METADATA AND ICONS ---
+TITLE = "Mid Sussex District Council (Whitespace WRP)"
+DESCRIPTION = "Source script for Mid Sussex District Council. Fetches collection dates using a four-step web scraping process on the Whitespace Waste & Recycling Portal."
+URL = "https://www.midsussex.gov.uk/waste-and-recycling/"
 
+# Date format for the final page: DD/MM/YYYY
+DATE_FORMAT = "%d/%m/%Y"
+
+# *** TEST CASES ***
 TEST_CASES = {
-    "Test_001": {"house_number": "6", "street": "Withypitts", "postcode": "RH10 4PJ"},
-    "Test_002": {
-        "house_name": "Oaklands",
-        "street": "Oaklands Road",
-        "postcode": "RH16 1SS",
-    },
-    "Test_003": {"house_number": 9, "street": "Bolnore Road", "postcode": "RH16 4AB"},
-    "Test_004": {"address": "HAZELMERE REST HOME, 21, BOLNORE ROAD, RH16 4AB"},
+    # 1. Simple Numerical Address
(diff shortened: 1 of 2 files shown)


Gdkkauowj33:
Mid Sussex changed to a new service provider, which presented a completely new web site and sequencing, hence fix required.

KarlocWow:
Appears to work fine, thanks

KaliPete:
For the less capable amongst us, what do we need to do to get Mid Sussex working again?
(I tried replacing the code in /homeassistant/custom_components/waste_collection_schedule/source/midsussex_gov_uk.py with the code I found on https://github.com/mampfes/hacs_waste_collection_schedule/pull/5083/commits/7b9bf2d72a5594706ee0cf77d55cc906be193516 and restarted HA, but it still errored out the same way)

Gdkkauowj33:
You can either wait until the changes are merged and then update as you would normally. Or you can copy the file from the link below. It’s possible that you might need to update your address settings in the GUI.   https://github.com/mampfes/hacs_waste_collection_schedule/blob/7b9bf2d72a5594706ee0cf77d55cc906be193516/custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py

KarlocWow:
I think I had to delete the existing entity then add the integration as if new (after copying the updated py file and restarting HA) either because the source fields changed or because I was incompetent.

custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py
```diff
@@ -1,154 +1,162 @@
...
+                class_='app-subnav__link',
+                string=lambda t: t and search_text in t.upper()
+            )
+
+            if not address_link:
+                if soup2.find(string=lambda t: t and "address not listed" in t.lower()):
+                    raise SourceArgumentException(
+                        "Address not found. Check house name/number and postcode.",
+                        argument="number"
                     )
```
5ila5, line 120: Please look at the classes before you use them

```py
class SourceArgumentException(Exception):
  def __init__(self, argument: str, message: str) -> None:
```

so this code generates a `SourceArgumentException.__init__() got multiple values for argument 'argument'`

I'd suggest raising a `SourceArgumentNotFound` here


```suggestion
                    raise SourceArgumentNotFound(
                        "number",
                        self._number
                    )
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py
```diff
@@ -1,154 +1,162 @@
...
+            r3 = session.get(final_schedule_url)
+            r3.raise_for_status()
+
+            soup3 = BeautifulSoup(r3.text, 'html.parser')
+
+            collection_entries = soup3.find_all('ul', class_='displayinlineblock')
+
+            if not collection_entries:
+                if soup3.find(string=lambda t: t and "not available" in t.lower()):
+                    return [] # Return empty list if schedule is unavailable
```
5ila5, line 136: Do not return an empty list as this will just fail silently. Raise an exception here so something shows up in the logs

custom_components/waste_collection_schedule/waste_collection_schedule/source/midsussex_gov_uk.py
```diff
@@ -1,154 +1,162 @@
...
+                if len(list_items) < 3:
+                    continue
+
+                date_str = list_items[1].p.text.strip()
+                waste_type = list_items[2].p.text.strip().upper()
+
+                try:
+                    collection_date = datetime.datetime.strptime(date_str, DATE_FORMAT).date()
+                except ValueError:
+                    print(f"Skipping entry with unparsable date format: {date_str}")
```
5ila5, line 151: If you want this to show up in the HA logs, you need to use the logging library

bbr111:
@Gdkkauowj33 
Can we close this?

Gdkkauowj33:
> @Gdkkauowj33 Can we close this?

Yes, trying fix under [5529](https://github.com/mampfes/hacs_waste_collection_schedule/pull/5529 )

Gdkkauowj33:
trying to fix under 5529.

PR #5102, opened 2025-12-17
Update awido_de.md Add Königstein

AlexJacu:
Working from today on.


5ila5:
Thanks for your contribution

I added Köngstein to the awido service map so it shows up at all the correct places

PR #5103, opened 2025-12-17
Update awido_de.py

AlexJacu:
Königstein uses AWIDO from now on instead of Mein Abfallkalender


5ila5:
I already did this to your last PR. Please don't spread changes across multiple PRs if they address the same issue.

Issue #3462, opened 2025-01-07
[Source Request]: PreZero Bad Oeynhausen, Germany


deankannenberg:
### Municipality / Region

Bad Oeynhausen, Germany

### Collection Calendar Webpage

https://abfallkalender.prezero.network/bad-oeynhausen

### Example Address

street: Eidingsen
houseNo: 2

### Collection Data Format

As ICS/ICAL file

deankannenberg:
https://github.com/mampfes/hacs_waste_collection_schedule/pull/3449

florian-asche:
PR is https://github.com/mampfes/hacs_waste_collection_schedule/pull/3449 what is the process to get it into release?

deankannenberg:
Is there any possibility to implement this source? Do I need to adjust anyting from coding perspective? I am getting asked from several people if we can merge this code.

5ila5:
added in  #5107 

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

Issue #3635, opened 2025-01-19
[Source Request]: Add Bad Oeynhausen, Germany


florian-asche:
### Municipality / Region

Bad Oeynhausen, Germany

### Collection Calendar Webpage

https://abfallkalender.prezero.network/bad-oeynhausen/calendar/791/39?year=2025&type=month&no=01

### Example Address

Gohfelder Poststr. 1, 32549 Bad Oeynhausen

### Collection Data Format

As ICS/ICAL file

### Additional Information

Stadtwerke / Rathaus: https://www.badoeynhausen.de/rathaus-service-politik/stadtverwaltung/was-erledige-ich-wo/dienstleistungen-a-z/abfallkalender
Directs to Service Provider: https://abfallkalender.prezero.network/bad-oeynhausen/calendar/791/39?year=2025&type=month&no=01
ICS: https://abfallkalender.prezero.network/bad-oeynhausen/download/ical/791/39/2025

If you like, please also add the other city, that is managed by prezero, that is the City Willich.

If you need help, please leave a feedback, i am happy to help!

florian-asche:
PR is https://github.com/mampfes/hacs_waste_collection_schedule/pull/3449 what is the process to get it into release?

florian-asche:
This is a duplicate of 3462

5ila5:
added in  #5107 

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

PR #3449, opened 2025-01-06
Adding PreZero Bad Oeynhausen as new source

deankannenberg:
Adding the wast collection provider for Bad Oeynhausen (PreZero)

PR #5107, opened 2025-12-17
New Source: PreZero Bad Oeynhausen

deankannenberg:
This PR adds PreZero Bad Oeynhausen as source.
It is using the iCal implementation

This closes: 
https://github.com/mampfes/hacs_waste_collection_schedule/issues/3462
https://github.com/mampfes/hacs_waste_collection_schedule/issues/3635
https://github.com/mampfes/hacs_waste_collection_schedule/pull/3449

If further adjustments are needed, let me know.


5ila5 (changes requested):
This could also have been done with an ICS source: https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_ics.md

But this is fine as well


I don't think putting bad-oyenhausen in the source name is a good idea if the source supports more regions (even though I noticed for willich I always get "Leider sind für diese Adresse noch keine Abfuhrtermine enthalten. Bitte versuchen Sie es zu einem späteren Zeitpunkt erneut.")

You source file name should probably follow the standard naming of url with `.` replaced with `_` (abfallkalender_prezero_network).

Your URL parameter should not contain a path.

As your source does not end in a country code (e.g. *_de.py) you need to prvide `COUNTRY = "de"` by the other "magic" variables at the top of the file

deankannenberg:
Thank you for the hint. I renamed the source according to the URL (abfallkalender.prezero.network) and I added the "COUNTRY" variable.

5ila5:
I prefer this, You could put a EXTRA_INFO section there so Bad Oeyenhausen still appears in the README and the HA config GUI as a link to this source (Maybe stadt Willich as well, but I'm not sure if this is even properly supported

deankannenberg:
The EXTRA_INFO has been added and the Source is now supporting Bad Oeynhausen (default) and Willich. I adjusted the documentation as well.

5ila5 (approved):

Issue #5077, opened 2025-12-13
[Bug]: South Holland stopped working. Client error. Forbidden for URL


dustyviper:
### I Have A Problem With:

A specific source

### What's Your Problem

Hello hope you are all well. I noticed that the south holland pull stopped working. 

I did try the update suggested In previous closed issues (https://github.com/mampfes/hacs_waste_collection_schedule/pull/4788 and https://github.com/mampfes/hacs_waste_collection_schedule/issues/4702) but my self along with a couple up of other people in the previous closed issue receive the error HTTPError: 403 Client Error: Forbidden for url: https://www.sholland.gov.uk/article/7156/Check-your-collection-days

I tried to install the very latest update manually (copying the file over and renaming the old one) as well as via the home assistant GUI as per other suggestions in those previous issues but it still kept erroring with the same.

Thanks in advance for looking into this for us.

### Source (if relevant)

sholland_gov_uk

### Logs

```Shell
HTTPError: 403 Client Error: Forbidden for url: https://www.sholland.gov.uk/article/7156/Check-your-collection-days
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

pcaouolte:
The fix from Lewis1352 works for me at two different properties.

dustyviper:
I can confirm that the fix from @Lewis1352 does work and fixes the issue! Thanks so much for sorting this out.

My sensors did need amending slightly to make it co-operate with my BIndicator but Lewis1352 has done the hard part so thanks once again.

5ila5:
fixed by #5111 

This will be part of the next release or is available now as master (git) version 

[...]

Issue #4804, opened 2025-10-11
[Bug]:


JDoades:
### I Have A Problem With:

A specific source

### What's Your Problem

Sholland (South Holland) source returns an error. Page URL has changed from 

https://www.sholland.gov.uk/article/7156/Check-your-collection-days

to

https://www.sholland.gov.uk/mycollections

Changing this in the python script file and uploading it manually to my home assistant integration has fixed it for me. I would upload as a pull request, but have absolutely no idea how to do that lol.

### Source (if relevant)

sholland_gov_uk

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

pcaouolte:
I agree that the Page URL has changed to https://www.sholland.gov.uk/mycollections

Marvman100:
Hi @JDoades & @pcaouolte 

I know this is a bit old but....

Can I ask if your collection integration is working, could you share your code from sholland_gov_uk.py please?
I'm guessing this is the only update you needed to make?

I've updated the "API_URL" but I'm getting "Exception: Unable to find form" in the logs

Thanks for any help

pcaouolte:
Hi @Marvman100 

My collection integration is **not** working. It doesn't matter if I use the old URL or the new one, it doesn't work.

It fails with:
[...]

PR #5111, opened 2025-12-17
Fixes sholland_gov_uk switch to JSON backend

Lewis1352:
South Holland District Council has switched to using a JSON backend.
This replaces the HTML scraping with a direct call to the council's backend JSON-RPC API (`SouthHolland.Waste.getCollectionDaysAjax`).

Possible fix for #5077 and #4804

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sholland_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sholland_gov_uk.py
@@ -1,7 +1,7 @@
 import datetime
 import re
-import requests
 
+import requests
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
 TITLE = "South Holland District Council"
@@ -19,7 +19,7 @@ ICON_MAP = {
     "recycling": "mdi:recycle",
 }
 
-TYPE_LABELS = { 
+TYPE_LABELS = {
     "refuse": "Refuse",
     "recycling": "Recycling",
     "garden": "Garden",
@@ -34,6 +34,7 @@ HEADERS = {
     "Referer": "https://www.sholland.gov.uk/mycollections",
 }
 
+
 class Source:
     def __init__(self, uprn: str | int, postcode: str):
         self._uprn = str(uprn)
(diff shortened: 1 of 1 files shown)


pcaouolte:
I've tested this and it works for me at two different properties. Thank you for your efforts. 🥇

Marvman100:
Hi @Lewis1352 

Thanks so much for your efforts on this

I'm hoping you can assist please

I've replaced my sholland_gov_uk.py file with the one you have uploaded but I'm getting "API call unsuccessful"

`fetch failed for source South Holland District Council: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/sholland_gov_uk.py", line 57, in fetch raise Exception("API call unsuccessful") Exception: API call unsuccessful`

any help appreciated

Lewis1352:
Hi @Marvman100 

Thanks for letting me know. Believe I've just created a fix, when dropping this into my HA it seems to work. 

If it's still not working for you, let me know. First though, please double-check that your UPRN is correct. From my testing, the backend no longer requires a postcode and just uses the UPRN now.

dustyviper:
@Lewis1352 - Not all hero's wear capes. Thanks so much for sorting this out. I can confirm it does work!

My sensors did need amending slightly to make it co-operate with my BIndicator but you've done the hard part so thanks once again.

Marvman100:
Hi @Lewis1352 

Thanks so much for this.. all sorted.

I was in the process of typing a reply... after your first fix and a couple of HA restarts with no luck I removed the integration and re-added and I got the events showing in the calendar and was just typing about sensors not working and boom you already had a fix!!

As per dustyviper comment I know I wont miss you around SH with your cape on with a big HA on the back!!!!

I'm sure we will all be back here in a few months when we have six wheely bins to figure out with this integration 

thanks again

dustyviper:
AHH the 2nd one you release meant I didn't have to change my sensors. Massive thanks!

5ila5:
I did some reformatting

But running the script shows a `403 Client Error: Forbidden`

Maybe my german IP getting geoblocked? But I can lookup collections here: https://www.sholland.gov.uk/mycollections

Does this still work for you? Then I'll merge this.

Lewis1352:
@5ila5 Tested and works for me, thanks.

Yeah looks like it's geo-blocking on the council side. The endpoint is an internal AJAX API that their website uses. Because so few headers are required it's likely the WAF just stops access based on IP location.

5ila5:
Thanks for your contribution

Issue #5100, opened 2025-12-17
[Bug]: umweltverbaende_at stopped working


mindestens:
### I Have A Problem With:

A specific source

### What's Your Problem

After HA restart the attached error occurs. Therefore the integration is broken.

Thank you!

### Source (if relevant)

umweltverbaende_at

### Logs

```Shell
Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Quelle: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
Integration: Waste Collection Schedule (Dokumentation)

fetch failed for source Die NÖ Umweltverbände: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py", line 490, in fetch return self.fetch_new() ~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py", line 697, in fetch_new bin_type = zone["fraktion"] ~~~~^^^^^^^^^^^^ TypeError: string indices must be integers, not 'str'
```

### Relevant Configuration

```YAML
https://stpoeltenland.umweltverbaende.at/fuer-die-bevoelkerung/abholtermine/#

waste_collection_schedule:
  sources:
    - name: umweltverbaende_at
      args:
        district: "stpoeltenland"
        municipal: "-REMOVED-"
        town: "-REMOVED-"
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

mindestens:
[...]

PR #5135, opened 2025-12-22
Fix: TypeError and add multi-year support for umweltverbaende.at

mindestens:
- Fix TypeError: "string indices must be integers, not 'str'"  Added isinstance() check before iterating zones list, as API can return "error" string   #5100 
  
- Add multi-year support for December queries. When the current month is December, query both next year and current year. Prioritizes next year data (more relevant for end-of-year queries)
  
- Handle optional dropdowns correctly. API returns "error" or empty string for dropdowns that don't exist. Added proper validation to skip non-existent dropdowns gracefully
  
- Add year parameter to API requests. API requires "jahr" parameter with current/next year as a string
  
- Improve error handling and fallback chain fetch_new() → fetch_old() if new method fails or returns no data ICS calendar detection and automatic usage when available
  
- Add complex address parameter support. Some districts (e.g., St. Pölten/Neulengbach) require a full address: gemeinde + town + plz + street for correct waste collection data

- Fix **Scheibbs** region

- 2 Test cases removed (Lilienfeld and Tulln); No ICAL or API available anymore - only redirects to the local municipal websites

image

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
@@ -6,7 +6,6 @@ import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import (
-    SourceArgumentException,
     SourceArgumentNotFoundWithSuggestions,
     SourceArgumentRequiredWithSuggestions,
 )
@@ -87,7 +86,7 @@ EXTRA_INFO: list[E_I_TYPE] = [
     #     "default_params": {
     #         "district": "klosterneuburg",
     #     },
-    # }, The site still exists but does not offer any shedules anymore they now have a own page not yet supported by this integration: https://www.klosterneuburg.at/Natur_Umwelt/Recycling/Muellabfuhr/Muellabfuhrkalender
+    # }, The site still exists but does not offer any schedules anymore they now have a own page not yet supported by this integration: https://www.klosterneuburg.at/Natur_Umwelt/Recycling/Muellabfuhr/Muellabfuhrkalender
     {
         "title": "Abfallverband Korneuburg",
         "url": "https://korneuburg.umweltverbaende.at/",
@@ -329,10 +328,10 @@ TEST_CASES = {
     },
     # "Krems": {"district": "krems", "municipal": "Aggsbach"}, # 0 results with this config
     # "Stadt Krems Old Version": {"district": "kremsstadt", "municipal": "Rehberg"}, # Does not offer same schedule but is supported via generic ICS kremsstadt_umweltverbaende_at
-    #"Lilienfeld": {
+    # "Lilienfeld": {
     #    "district": "lilienfeld",
     #    "municipal": "Annaberg",
-    #},  # No ICAL or API available anymore - only redirects to the local municipal websites
+    # },  # No ICAL or API available anymore - only redirects to the local municipal websites
     "Laa/Thaya": {
         "district": "laa",
         "municipal": "Staatz",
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/zaw_de.py
5ila5, line 1: ZAW is already supported throug the jumomind_de source. It is already listed in the README.md and the HA config GUI


https://github.com/mampfes/hacs_waste_collection_schedule/blob/05a7ebd6641dbfe4938fa4ea463b381f53429569/README.md?plain=1#L990

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
```diff
@@ -709,9 +743,7 @@ def fetch_new(self) -> list[Collection]:
             or not (nonce := nonce_match.group(1))
             or not isinstance(nonce, str)
         ):
-            raise Exception(
-                f"Could not find nonce for page {self._district_url}fuer-die-bevoelkerung/abholtermine/"
-            )
+            return []
```
5ila5: Why? Silent failure is a bad thing.

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
```diff
@@ -486,16 +493,40 @@ def append_entry(self, ent: list, txt: list):
...
+                try:
+                    return self.fetch_old()
+                except Exception:
+                    # If both methods fail, raise the original error from new method
+                    raise e
         return self.fetch_old()
 
+    def fetch_new_year(self, year: int) -> list[Collection]:
+        """Fetch data for a specific year using the new API method."""
+        return self.fetch_new(year=year)
```
5ila5: Why does this method exists? you can just call fetch_new?

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
```diff
@@ -486,16 +493,40 @@ def append_entry(self, ent: list, txt: list):
...
-            return self.fetch_new()
+            try:
+                # In December, try both current and next year
+                now = datetime.now()
+                if now.month == 12:
+                    entries_current = self.fetch_new_year(now.year)
+                    entries_next = self.fetch_new_year(now.year + 1)
+                    entries = entries_current + entries_next
+                else:
+                    entries = self.fetch_new_year(now.year)
```
5ila5: You should probably put the second request into a try except block if only next year fails it should still use the collection from this year.


```suggestion
                now = datetime.now()
                entries = self.fetch_new_year(now.year)
                if now.month == 12:
                    try:
                        entries += self.fetch_new_year(now.year + 1)
                    except Exception:
                        pass
```

5ila5 (changes requested):
Please remove the ZAW source

(even if it were new: your MD file has the wrong name and I would have liked to have these separate issues in separate PRs)

custom_components/waste_collection_schedule/waste_collection_schedule/source/zaw_de.py
mindestens: Thank you for the info!

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
mindestens: Indeed, thank you!

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
mindestens: Thanks, fixed!

custom_components/waste_collection_schedule/waste_collection_schedule/source/umweltverbaende_at.py
mindestens: Indeed, fixed as well

5ila5:
Thanks for your contribution

PR #5157, opened 2025-12-26
Fix date parsing for wyreforestdc

matty:
Looks like some of the dates that are returned don't match the expected, so added additional try for parsing those dates. Tested on my HA and it works.


5ila5:
Thanks for your contribution.

PR #5159, opened 2025-12-26
awb_es_de: Replace requests with cloudscraper for session handling

meilon:
Closes #5027 
Thanks to @5ila5 for the cloudscraper idea, works for me as well!


5ila5:
Thanks for your contribution

PR #5169, opened 2025-12-27
feat(heinz_entsorgung_de): Add new source for redesigned website

tsgoff:
The Heinz-Entsorgung website has been completely redesigned from a WordPress-based ICS download to a Blazor WebAssembly application with encrypted API parameters.

Changes:
- Add new heinz_entsorgung_de source using the JSON API
- Update ICS documentation to mark old method as deprecated
- Add source documentation with instructions for parameter extraction
- Add test file for the new source

The new source requires users to extract an encrypted 'param' value from the website using browser DevTools, as the encryption cannot be reproduced programmatically.


doc/ics/heinz_entsorgung_de.md
5ila5, line 1: This does not work this file is automatically generated from `doc/ics/yaml/heinz_entsorgung_de.yaml` when running update_docu_links.py

5ila5:
Thanks for your contribution

Issue #5173, opened 2025-12-27
[Bug]: Source ALW_WF_DE does not work anymore


knumskull:
### I Have A Problem With:

A specific source

### What's Your Problem

I only see calendar entries up to 31st 2025, although more are available.

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: alw_wf_de
      args:
        ort: "Sickte"
        strasse: "Am Mühlengraben"
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

sniffulus:
I have the same problem. The website mentions a new app for Android and iOS. I assume that the calendar has also been redesigned.
See: [https://www.alw-wf.de/index.php/abfallkalender](https://www.alw-wf.de/index.php/abfallkalender)

knumskull:
@sniffulus - I've figured it out and creating a pull request right now.
With the new app, ALW switched to use Jumomind provider. So it's possible to configure the service like this.

```
waste_collection_schedule:
  sources:
    - name: jumomind_de
      args:
        service_id: "wol"
        city: "Linden"
        street: "Am Buschkopf"
```

chrkoe:
thanks @knumskull 
I just realized today that they had changed their systems. Glad you checked and worked on it already.

Just for the links -> PR #5174 (for some reasons it doesn't show up correct.)

chrkoe:
@knumskull I applied your changes locally in my custom_components directory but still got errors in the GUI configuration. I figured out that there was still a change missing in sources.json. I forked your changes and also added the last fix. With that I made it working locally.

5ila5:
This is fixed in #5174 

[...]

PR #5174, opened 2025-12-27
fix: switch source to use jumomind (#5173)

knumskull:
ALW Wolfenbüttel changed their backend and using jumomind from 2026 on. They also provided a new app.
I've changed the documentation to reflect all the necessary changes.

It's easy to configure the service with what already exist in the hacs extension.

```
waste_collection_schedule:
  sources:
    - name: jumomind_de
      args:
        service_id: "wol"
        city: "Linden"
        street: "Am Buschkopf"
```


Daedalica:
Thank you!

chrkoe:
fixes #5173

5ila5:
Thanks for your contribution

I added a migration so users (who configured using the GUI, not YAML) should be automatically migrated. This may not work with all cities

Issue #5137, opened 2025-12-23
Provider Städteservice Raunheim-Rüsselsheim (staedteservice_de), Fehler 405 seit ~20. Dezember 2025, URL die du im Log siehst, und dass die Webseite manuell noch funktioniert (ICS-Download möglich).


MLight007:
Provider Städteservice Raunheim-Rüsselsheim (staedteservice_de), Fehler 405 seit ~20. Dezember 2025, URL die du im Log siehst, und dass die Webseite manuell noch funktioniert (ICS-Download möglich).

franky1808:
Unfortunately, I'm having the same problem. I've been using it via HACS until now. But since this morning, it's no longer working.

Unfortunately, it's not possible to set it up as a regular HA device:

“The source returned an invalid response: ”405 Client Error: Method Not Allowed for url: https://portal.staedteservice.de/api/ZeigeAbfallkalender?orteId=2&strassenId=59&fixedYear=2025&hausNr=%278b%27&unixZeitOption=0&dateiName=%27Abfallkalender2025.ics%27.“ Please check the arguments and try again.”

"Die Quelle hat eine ungültige Antwort zurückgegeben: "405 Client Error: Method Not Allowed for url: https://portal.staedteservice.de/api/ZeigeAbfallkalender?orteId=2&strassenId=59&fixedYear=2025&hausNr=%278%27&unixZeitOption=0&dateiName=%27Abfallkalender2025.ics%27". Bitte überprüfe die Argumente und versuche es erneut."

franky1808:
Duplicate #5124

cgomm:
I have added an PR to address this issue: #5176

5ila5:
fixed in #5176 (I know I kind of marked the wrong issues as duplicate)

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

5ila5:
fixed in #5176 (I know I kind of marked the wrong issues as duplicate)

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
[...]

Issue #5124, opened 2025-12-20
[Bug]: Staedteservice Raunheim Rüsselsheim 405 Client Error: Method Not Allowed for url:


jheisel01:
### I Have A Problem With:

The integration in general

### What's Your Problem

Steps I made:
add Hub - Germany - Städteservice Raunheim Rüsselsheim
Error : Die Quelle hat eine ungültige Antwort zurückgegeben: "405 Client Error: Method Not Allowed for url: https://portal.staedteservice.de/api/ZeigeAbfallkalender?orteId=1&strassenId=20&fixedYear=2025&hausNr=%2720%27&unixZeitOption=0&dateiName=%27Abfallkalender2025.ics%27". Bitte überprüfe die Argumente und versuche es erneut.
No further information available, as the source cannot be configured.

![Image](https://github.com/user-attachments/assets/d4560013-e5de-46fd-a3f6-4eb543e2a5a4)

### Source (if relevant)

https://portal.staedteservice.de

### Logs

```Shell
Die Quelle hat eine ungültige Antwort zurückgegeben: "405 Client Error: Method Not Allowed for url: https://portal.staedteservice.de/api/ZeigeAbfallkalender?orteId=1&strassenId=20&fixedYear=2025&hausNr=%2720%27&unixZeitOption=0&dateiName=%27Abfallkalender2025.ics%27". Bitte überprüfe die Argumente und versuche es erneut.
```

### Relevant Configuration

```YAML
Waste Collection Schedule - Version 2.10.0

Home Assistant Core 2025.12.4
Supervisor 2025.12.3
Operating System 16.3
Frontend 20251203.3
```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

franky1808:
Duplicate #5137

[...]

PR #5176, opened 2025-12-27
fix: Change API call from GET to POST method for Staedteservice

cgomm:
The Staedteservice changed the API spec to require the API to get called via POST on the endpoint ZeigeAbfallkalender.

Testresult via POST:
image

Testresult via GET:
image

The tickets #5124 and #5137 will be addressed by this PR.

---

Changed
* Changed the GET request for `ZeigeAbfallkalender` to POST.

Tested with HA `2025.12.4`


5ila5:
Thanks for your contribution

PR #5195, opened 2025-12-29
Added City of Subiaco, Western Australia as a source.

elSpike:
Created resources for the City of Subiaco in Western Australia based of the information located at https://www.subiaco.wa.gov.au/residents/in-your-home/bins,-recycling-and-waste/waste-area-guide

Files created are:

subiaco_wa_gov_au.py
subiaco_wa_gov_au.md

Updated info and readme using the update_docu_links.py process.

(venv) patrick@neliteist test % ./test_sources.py -s subiaco_wa_gov_au
Testing source subiaco_wa_gov_au ...
  found 114 entries for Area 1 Monday
  found 114 entries for Area 2 Wednesday

--- Excerpt from:
(venv) patrick@neliteist test % ./test_sources.py -s subiaco_wa_gov_au -i -l
Testing source subiaco_wa_gov_au ...
  found 114 entries for Area 1 Monday
    2025-12-01 : FOGO [mdi:leaf]
    2025-12-01 : General Waste [mdi:trash-can]
    2025-12-08 : FOGO [mdi:leaf]
    2025-12-08 : Recycling [mdi:recycle]
    2025-12-15 : FOGO [mdi:leaf]
    2025-12-15 : General Waste [mdi:trash-can]
    2025-12-22 : FOGO [mdi:leaf]
    2025-12-22 : Recycling [mdi:recycle]


5ila5:
I really don't like these "hardcoded" dates sources. I advise using the static source instead. If the reference date seems to stay the same (even across years) I'm fine with adding this.

You should then check for holidays with variable year. 

You could do `rrule` stuff here (`from dateutil import rrule`) but I doubt it helps a lot (especially for the holiday exceptions)

elSpike:
Thanks @5ila5. Unfortunately the council only provides a JPEG image that looks like a fridge magnet with the dates of collection and holidays.

Area-2-Calendar

Until this changes we will have to update the source each time that file is published. Alternatively I can create and host an iCS or static source but it would not be on an official domain, just my own and pretty much the same problem of needing manual editing as the image is published.

Which would you prefer?
Patrick

5ila5:
The problem with these sources, is they need maintenance or report wrong dates.
Do you know if the reference dates stayed the same?

I think this is fine, but I still think you should replace the holidays with a check for day + month (without year).
For Easter, you could use `from dateutil.easter import easter` but I don't know which "Easter holiday" is treated like a holiday.

elSpike:
I've updated based on your recommendation to use dateutil to generate Good Friday (Easter Sunday -2) and removed the years for the fixed New Year and Christmas holidays. I've also added some logic for future years in case the PH lands on a Saturday or Sunday and places it on the following Monday as is the norm here.

Thanks again for the advice!

5ila5:
LGTM,
thanks for your contribution

Issue #4546, opened 2025-08-06
[Bug]: Fetch failed for source Gateshead Council


rathmannm:
### I Have A Problem With:

A specific source

### What's Your Problem

The integration has stopped working for Gateshead council with the following error message:

Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/gateshead_gov_uk.py", line 39, in fetch r.raise_for_status() ~~~~~~~~~~~~~~~~~~^^ File "/usr/local/lib/python3.13/site-packages/requests/models.py", line 1026, in raise_for_status raise HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.gateshead.gov.uk/article/3150/Bin-collection-day-checker

Any chance this could be looked at please?

Thanks!

### Source (if relevant)

gateshead_gov_uk

### Logs

```Shell
Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 158, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/gateshead_gov_uk.py", line 39, in fetch r.raise_for_status() ~~~~~~~~~~~~~~~~~~^^ File "/usr/local/lib/python3.13/site-packages/requests/models.py", line 1026, in raise_for_status raise HTTPError(http_error_msg, response=self) requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://www.gateshead.gov.uk/article/3150/Bin-collection-day-checker
```

### Relevant Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: gateshead_gov_uk
      args:
        uprn: 100000070113
      customize:
        - type: Household
          show: true
      calendar_title: Bin Collection
```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
[...]

PR #5197, opened 2025-12-29
Fix UK, Gateshead Council #4546

06benste:
fix for issue #4546

Summary
This PR fixes Cloudflare protection blocking requests for the Gateshead Council waste collection source (`gateshead_gov_uk.py`).

Problem
The Gateshead Council website is protected by Cloudflare, which blocks automated requests and returns 403 Forbidden errors, preventing the integration from fetching bin collection schedules.

Solution
- Added `cloudscraper` support to bypass Cloudflare protection
- Falls back to regular `requests` with browser headers if `cloudscraper` is not available
- Added realistic browser headers (User-Agent, Accept, Accept-Language, etc.) to mimic browser requests
- Added 30-second timeout to HTTP requests

Changes
- Import `cloudscraper` with fallback to `requests`
- Use `cloudscraper.create_scraper()` when available to handle Cloudflare challenges
- Add browser headers as fallback when using regular `requests.Session()`

Testing
- Tested with real Gateshead Council UPRN - successfully retrieves collection dates
- Confirmed Cloudflare bypass works correctly

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/gateshead_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/gateshead_gov_uk.py
@@ -142,7 +142,14 @@ class Source:
             elif dt < now.date() and (now.date() - dt).days > 180:
                 dt = dt.replace(year=now.year + 1)
 
+            day_change_regex = re.compile(
+                r"(.*?)\s*-?\s*DAY CHANGE\s*$", re.IGNORECASE | re.DOTALL
+            )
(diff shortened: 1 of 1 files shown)


06benste:
I've been testing for a few days and all is working as expected so I have removed the fallback to requests and have migrated to pure cloudscraper use. Commit made to update.

5ila5:
Thanks for your contribution

I noticed the last test case may generate a entry containing `DAY CHANGE` I removed that `DAY CHANGE` in 0cbd6b6

PR #5203, opened 2025-12-29
repair was_wolfsburg_de

bbr111:
Adds a new source for Wolfsburger Abfallwirtschaft und Straßenreinigung (WAS) using the current public JSON API.

The implementation replaces the deprecated PHP endpoint and uses the same API and token as the official WAS frontend.

Details

Uses https://apiabfuhrtermine.waswob.de/api/download-json

Public, static API token (from frontend JS)

Supports Bioabfall, Restabfall, Altpapier, Wertstofftonne

street and house_number are required

Dates parsed from ISO format (YYYY-MM-DD)


bbr111:
fixes #5196

PR #5214, opened 2025-12-30
fix: redbridge council when no collections present for type

ryanwr:
This PR fixes redbridge council source when collections stop for a waste type. Have noticed missing data for a few weeks now.

Issues:
- Waste types are sometimes suspended e.g. garden waste over winter and hence extracting a date fails
- Similar to above, new waste types being introduced soon e.g. food waste causing parsing to break
- Year seems to have not previously been included but is now entered with the month text in format 'February 2026'

Improvements:
- Skip waste type if no dates available
- Try to use year included in month field and fallback to previous logic if not present

Future improvements:
Once available we should add the new waste types e.g. food


5ila5:
Thanks for your contribution

Issue #4352, opened 2025-07-05
[Bug]: Coventry City Council, UK


IliasEagle:
### I Have A Problem With:

A specific source

### What's Your Problem

I checked my configuration multiple times before raising this ticket. 

At least for me the "Rubbish" and "Recycling" dates are inverted.
This is the schedule from the Coventry City council:

<img width="555" height="724" alt="Image" src="https://github.com/user-attachments/assets/1b62e81e-cb46-4053-bda1-9899bca8e00b" />

and here is the output from the integration:
```
Attributes
Types
Recycling, Rubbish
Upcoming
- date: '2025-07-11'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-07-18'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-07-25'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-08-01'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-08-08'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-08-15'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-08-22'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-08-29'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-09-05'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-09-12'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-09-19'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-09-26'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-10-03'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-10-10'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-10-17'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-10-24'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-10-31'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-11-07'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-11-14'
icon: mdi:recycle
picture: null
type: Recycling
- date: '2025-11-21'
icon: mdi:trash-can
picture: null
type: Rubbish
- date: '2025-11-28'
[...]

PR #5215, opened 2025-12-30
Coventry fix to choose first matching search result

JustinWingChungHui:
Fixes https://github.com/mampfes/hacs_waste_collection_schedule/issues/4352


5ila5:
This is probably fine if we assume the “shortest match” is always the first entry. Alternatively, we could enforce the exact spelling, but this might break existing configurations

5ila5:
Thanks for your contribution

PR #5220, opened 2025-12-31
Add support for OZO Ostrava (Czech Republic)

viki-vavrik:
hi! i used gemini to create this script to add schedules for OZO Ostrava. https://ozoostrava.cz/svoz


custom_components/waste_collection_schedule/waste_collection_schedule/source/ozoostrava_cz.py
```diff
@@ -0,0 +1,83 @@
...
+URL = "https://ozoostrava.cz"
+
+EXTRA_INFO = [
+    {
+        "title": "OZO Ostrava",
+        "url": "https://ozoostrava.cz",
+        "country": "cz",
+        "region": "Moravskoslezský kraj",
+    }
+]
```
5ila5: Remove this

This will create duplicate entries. It is used for additional regions not the same again, plus `region` is not valid here anyway

5ila5 (changes requested):
please follow the guide here https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/contributing_source.md

and provide a md file

5ila5:
Just so you know, this is still missing, the MD documentation file. (and running `python update_docu_links.py` but I can do this part as well before merging)

bbr111:
Please Update your PR

PR #5222, opened 2025-12-31
fix(Maidstone): Fix holiday duplicates and filter inactive services

ArronTay:
## Description
Updated the `maidstone_gov_uk` source to address two issues observed during the holiday period:

1.  **Duplicate/Incorrect Dates:** The previous logic grabbed all keys containing `NextCollectionDateMM`, which included `Default` (standard schedule) and `Original` (pre-reschedule) dates. During holidays, this resulted in the integration showing both the cancelled date and the rescheduled date. Logic updated to strict matching on the actual next collection date.
2.  **Inactive Bins:** The API returns data for all possible services (Bulky, Clinical, etc.) even if the user hasn't subscribed. Added a check for the `_Active` flag in the JSON response to filter out services set to "N".

## Verification
- Tested during the Christmas rescheduling period.
- Confirmed that "Default" dates (e.g., Jan 7th) are ignored in favour of rescheduled dates (e.g., Jan 8th).
- Confirmed that inactive services (Clinical/Bulky) no longer appear in the schedule.


5ila5:
Thanks for your contribution

Issue #3608, opened 2025-01-16
[Bug]: France Saint Quentin en Yvelines (instance id : 701) doesn't work


cyhoarau:
### I Have A Problem With:

A specific source

### What's Your Problem

Hi,
I'm new to the community and development in general, so sorry if I'm not using the right terms or process to ask for bug fix / evolution.
Here is the issue : The source for France Instance id 701 has added a type of waste named "sapin" (pine tree). As it is not declared in following structures in the following file:
/homeassistant/custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_fr.py

ICON_MAP = {
    "omr": "mdi:trash-can",
    "emb": "mdi:recycle",
    "enc": "mdi:truck-remove",
    "dv": "mdi:leaf",
    "verre": "mdi:bottle-wine",
}

LABEL_MAP = {
    "omr": "Ordures ménagères",
    "emb": "Emballages",
    "enc": "Encombrants",
    "dv": "Déchets verts",
    "verre": "Verres",
}

This leads to an error message (exception appears in the dialog box) quite difficult to interpret after having submitted the adress, instance id, INSEE code

This can be fixed by including the "sapin" (pine tree, collected after christmas) entry in those structures. Example below:

ICON_MAP = {
    "omr": "mdi:trash-can",
    "emb": "mdi:recycle",
    "enc": "mdi:truck-remove",
    "dv": "mdi:leaf",
    "verre": "mdi:bottle-wine",
    **"sapin": "mdi:tree",**
}

LABEL_MAP = {
    "omr": "Ordures ménagères",
    "emb": "Emballages",
    "enc": "Encombrants",
    "dv": "Déchets verts",
    "verre": "Verres",
    **"sapin": "Sapin",**
}

It works in my environment after this fix. 

Following test case included in the  /homeassistant/custom_components/waste_collection_schedule/waste_collection_schedule/source/publidata_fr.py file doesn't work today. Will be fixed with the proposal:

    "Saint Quentin en Yvelines, Coignières": {
        "address": "1 rue du four à chaux",
        "insee_code": "78168",
        "instance_id": 701,
    },

Regards

### Source (if relevant)

publidata_fr.py

### Logs

```Shell
no relavant loggs: error message appears in the dialog box
```

### Relevant Configuration

```YAML
At the confiuguration of the sensor:

Country: France
Source: Saint Quentin en Yvelines (publidata_fr)
Address": 1 rue du four à chaux
[...]

PR #5224, opened 2025-12-31
fix: (publidata_fr) add missing 'bio' waste type mapping

klsx0:
## Description
This PR addresses an issue where the `publidata_fr` source returns a waste type `bio` (Biodéchets) that was missing from the internal `LABEL_MAP` and `ICON_MAP`. 
Previously, receiving this unmapped type caused the integration to fail with a `NoneType object has no attribute 'strip'` error, as the system attempted to process a `None` waste type.
## Changes
- Added `bio` mapped to `"Biodéchets"` in `LABEL_MAP`.
- Added `bio` mapped to `"mdi:food-apple"` in `ICON_MAP`.


5ila5:
Thanks for your contribution. This should not have happened. I added a fallback lookup value

I also added a icon for `sapin` like suggested in #3608

Issue #4927, opened 2025-11-11
[Bug] Bolton Council provider broken - carehomes.bolton.gov.uk no longer exists


bolarinwa01-byte:
**Summary**
The Bolton Council provider appears to be broken — it references an old URL (carehomes.bolton.gov.uk/bins.aspx) that no longer resolves.

**Error message**
Error fetching data: HTTPSConnectionPool(host='carehomes.bolton.gov.uk', port=443): Max retries exceeded with url: /bins.aspx 
(Caused by NameResolutionError("Failed to resolve 'carehomes.bolton.gov.uk ([Errno -5] Name has no usable address)")).

**Likely cause**
The subdomain `carehomes.bolton.gov.uk` no longer exists. The current bin collection page is now hosted under:
https://www.bolton.gov.uk/next-bin-collection

**Environment**
- HA version: 2025.11.x
- Integration version: latest (HACS)
- Config used:
```yaml
waste_collection_schedule:
  sources:
    - name: bolton_council
      args:
        postcode: "BL1 4TG"
        house_number: "1"

melhiore:
Me too. Tried to update today. Getting:

`The source returned an invalid response: "Error fetching data: HTTPSConnectionPool(host='carehomes.bolton.gov.uk', port=443): Max retries exceeded with url: /bins.aspx (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f9aa476c190>: Failed to resolve 'carehomes.bolton.gov.uk' ([Errno -5] Name has no usable address)"))". Please check the provided arguments and try again.`

bolarinwa01-byte:
> Me too. Tried to update today. Getting:
> 
> `The source returned an invalid response: "Error fetching data: HTTPSConnectionPool(host='carehomes.bolton.gov.uk', port=443): Max retries exceeded with url: /bins.aspx (Caused by NameResolutionError("<urllib3.connection.HTTPSConnection object at 0x7f9aa476c190>: Failed to resolve 'carehomes.bolton.gov.uk' ([Errno -5] Name has no usable address)"))". Please check the provided arguments and try again.`

Any luck with a fix? It goes to Bolton care homes instead of the bin collection website

darrencuk:
Seems to be working for Bolton after the latest update. 26/12/25

bolarinwa01-byte:
> Seems to be working for Bolton after the latest update. 26/12/25

[...]

PR #5262, opened 2026-01-03
Use the verintcloudservices.com API for Bolton UK

chartionic:
The [Gov.UK website](https://www.gov.uk/rubbish-collection-day/bolton) now links out directly to https://bolton.portal.uk.empro.verintcloudservices.com/ rather than the previous service: https://web.bolton.gov.uk/bins.aspx.

While the latter is still accessible, it doesn't contain data for all bins (for me at least).

I've made a little tweak to the first test data item as the existing version of this source is matching on any house number that starts with the user supplied house number, and in this case the postcode started at a higher value.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
@@ -1,10 +1,13 @@
 import logging
-import requests
-
 from datetime import datetime, timedelta
+
+import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection
-from waste_collection_schedule.exceptions import SourceArgumentNotFound, SourceArgumentNotFoundWithSuggestions
+from waste_collection_schedule.exceptions import (
+    SourceArgumentNotFound,
+    SourceArgumentNotFoundWithSuggestions,
+)
 
 TITLE = "Bolton Council"
 DESCRIPTION = "Source for Bolton Council, UK."
@@ -18,7 +21,7 @@ API_URLS = {
     "authentication": "api/citizen?archived=Y&preview=false&locale=en",
     "postcode_lookup": "api/widget?action=propertysearch&actionedby=ps_address&loadform=true&access=citizen&locale=en",
     "set_object": "api/setobjectid?objecttype=property&objectid={uprn}&loaddata=true",
-    "collection_dates": "api/custom?action=es_get_bin_collection_dates&actionedby=uprn_changed&loadform=true&access=citizen&locale=en"
+    "collection_dates": "api/custom?action=es_get_bin_collection_dates&actionedby=uprn_changed&loadform=true&access=citizen&locale=en",
 }
 
 ICON_MAP = {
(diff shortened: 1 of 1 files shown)


chartionic:
This should hopefully correct the second issue mentioned in #4927 where not all bins were being retrieved.

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
```diff
@@ -31,148 +43,132 @@
...
-        kwargs["ssl_context"] = context
-        return super().init_poolmanager(*args, **kwargs)
+_LOGGER = logging.getLogger(__name__)
 
 
 class Source:
     def __init__(self, postcode: str, house_number: str):
         self._postcode = postcode
         self._house_number = str(house_number)
+        self._session = requests.session()
```
5ila5: I don't think this is a good idea. The object may live for months. you should create a new session for every fetch

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
```diff
@@ -31,148 +43,132 @@
...
-                                    )
-                                )
-                            except ValueError:
-                                continue
-
-        except requests.RequestException as e:
-            raise Exception(f"Error fetching data: {e}")
+        uprn, auth_token = self._find_uprn(auth_token=auth_token)
+        if not uprn:
+            return []
```
5ila5: don't `return []` this fails silently, raise an exception instead

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
```diff
@@ -31,148 +43,132 @@
...
+        for address in data["data"]:
+            label = address["label"]
+            # Try exact match first (e.g., "3 STREET" or "3, STREET")
+            if label.startswith(f"{self._house_number} ") or label.startswith(f"{self._house_number},"):
+                return address["value"], auth_token
+        
+        # If no exact match, try matching within the label (for cases like "WOODSLEIGH COPPICE 2")
+        for address in data["data"]:
+            label = address["label"]
+            import re
```
5ila5: move the import up

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
chartionic: I've moved that to the beginning of the fetch() method.

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
chartionic: Apologies for not reading the documentation on Exceptions first. I've altered the _find_upn method so it will always return a value, or throw an exception.

custom_components/waste_collection_schedule/waste_collection_schedule/source/bolton_gov_uk.py
chartionic: I've removed that matching logic altogether, it probably doesn't make sense to just look for the supplied number anywhere in the label. And the user will now have the available values in the UI or the logs if they entered a valid postcode.

5ila5:
Thanks for your contribution

cunners:
Fixes #5553

Issue #5153, opened 2025-12-26
[Source Request]: Kranenburg 47559 Germany


Lookinator:
### Municipality / Region

Kranenburg, 47559, Kreis Kleve, Northrine Westphalia, Germany

### Collection Calendar Webpage

https://www.kranenburg.de/rat-verwaltung/serviceportal/abfallnavi-abfallkalender

### Collection Data Format

As ICS/ICAL file

bbr111:
Add Gemeinde Kranenburg to AbfallNavi

Lookinator:
Heck yeahhh, thanks for picking this up!

PR #5277, opened 2026-01-04
Add Gemeinde Kranenburg AbfallNavi

bbr111:
Add Gemeinde Kranenburg AbfallNavi

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/AbfallnaviDe.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/AbfallnaviDe.py
@@ -230,8 +230,7 @@ class AbfallnaviDe:
             return list(streets.keys())
         if street is None:
             raise SourceArgumentRequiredWithSuggestions(
-                "street", "street is required of this city", list(
-                    streets.values())
+                "street", "street is required of this city", list(streets.values())
             )
         matches = [id for id, name in streets.items() if name == street]
         if len(matches) == 0:


bbr111:
fixes #5153

doc/source/abfallnavi_de.md
```diff
@@ -17,17 +17,17 @@ waste_collection_schedule:
...
+**ort**
 *(string) (required)*
 
-**strasse**  
-*(string) (optional)*  
+**strasse**
+*(string) (optional)*
 *required if the service provider website requests it*
 
-**hausnummer**
```
5ila5, line 30: Is there any reason you removed the line break? This is the format used by most of the sources

custom_components/waste_collection_schedule/waste_collection_schedule/service/AbfallnaviDe.py
```diff
@@ -225,7 +230,8 @@ def get_street_ids(self, city_id, street):
             return list(streets.keys())
         if street is None:
             raise SourceArgumentRequiredWithSuggestions(
-                "street", "street is required of this city", list(streets.values())
+                "street", "street is required of this city", list(
+                    streets.values())
```
5ila5: I use black for formatting (ruff should be a drop-in replacement). If you could start formatting the same way, this would make it easier to collaborate on this.

You can run automatic formatting on commit using [pre-commit](https://pre-commit.com/): .pre-commit-config.yaml:
setup `pre-commit install`. There is a a spell checker running that may flag correct values (especially if the file contains non english words) you then need to commit with `-n`. They don't integrate very well in IDEs but it still works if you use this for Source Controll management

bbr111: maybe that is the problem.
I use vscode + docker for dev.
this are the settings:
```
				"python.formatting.provider": "black",
				"editor.formatOnPaste": false,
				"editor.formatOnSave": true,
				"editor.formatOnType": true,
```
isn't black format break lines after 80 chars?

if you use different settings please let me know.

doc/source/abfallnavi_de.md
bbr111: No.

i think this came from

`				"editor.formatOnSave": true,`

from the includes devcontainer.json.
I didn't change anything in this config.

PR #5290, opened 2026-01-05
Update centralbedfordshire_gov_uk.py

royhuntley:
Fixed the UserAgent header to match what is sent by a browser - this avoids the 403 forbidden response.


royhuntley:
The site is protected by bot / WAF rules (not just headers)

Central Bedfordshire uses a web application firewall (WAF) layered on top of their CMS (Granicus / GovDelivery–style stack). These systems look at:

Request sequence

Form tokens / state

Cookies

JavaScript execution

POST payload validity

Timing + behaviour

PR #5291, opened 2026-01-05
Adding support for Kungälvs kommun as a source.

### LINKED ISSUE #4953: [Feature]: Kungälv Avfall/Energi trash collection schedule
### I propose a feature for:

The integration in general

Would be an appreciated addition

MattiasC:
I have added support for Kungälvs kommun as a source. 

The single source test pass and the pytest pass to 100%. I have also updated the documentation and ran the update_docu_links.py script.

Should fix #4953


custom_components/waste_collection_schedule/waste_collection_schedule/source/kungalv_se.py
```diff
@@ -0,0 +1,91 @@
...
+        response = requests.post(
+            f"{API_URL}/SearchAdress",
+            data={"searchText": self._street_address},
+            timeout=30,
+        )
+        response.raise_for_status()
+
+        address_data = response.json()
+        if not address_data.get("Succeeded") or not address_data.get("Buildings"):
+            raise SourceArgumentException("Failed to get schedule for ", self._street_address)
```
bbr111: Bugfix: SourceArgumentException previously received multiple arguments (tuple). Changed to a single formatted string to ensure proper error messages.
```suggestion
            raise SourceArgumentException(
                f"Failed to get schedule for {self._street_address}"
        )
```

custom_components/waste_collection_schedule/waste_collection_schedule/source/kungalv_se.py
```diff
@@ -0,0 +1,91 @@
...
+
+        # Extract building ID from first result (format: "Address, CITY (ID)")
+        building = address_data["Buildings"][0]
+        match = re.search(r"\((\d+)\)", building)
+        if not match:
+            raise SourceArgumentException("Could not extract building ID from result.", self._street_address)
+
```
5ila5: You're using `SourceArgumentException` wrong. Please look at the docstring.

> ## Args
> 
> *argument* : str
> The source argument that caused the exception (written exactly like in the source __init__).
> 
> *message* : str
> Message to be displayed for the provided argument.




I think you should use `SourceArgumentNotFound` here

markvp: done

Issue #4783, opened 2025-10-06
[Bug]: Unable to configure Waste collection


Sherman-c:
### I Have A Problem With:

The integration in general

### What's Your Problem

After select country / provider, and enter the property id, it display an error message
"The source returned an invalid response: "Expecting value: line 1 column 1 (char 0)". Please check the provided arguments and try again."
Country : Australia
Provider: Hornsby Shire Council
Property ID: ID provided by Hornsby with Google map search
https://hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySearch.aspx?EnquiryListId=2&ERB=91055

### Source (if relevant)

hornsby_nsw_gov_au

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

dt215git:
See discussion in #4029

5ila5:
fixed in  #5317

This will be part of the next release or is available now as master (git) version 

install master/git: Developer tools -> actions -> select `update: Install update` -> click `Coose entity` and select WCS -> select `version` and type `master` in the input field -> run the action -> restart HA
Do not  "update" WCS through HACS until the next release, as pressing the update button performs a "downgrade" to the latest release version

Issue #3529, opened 2025-01-10
[Bug]: Hornsby, NSW, Australia - source broken


mkstalen:
### I Have A Problem With:

A specific source

### What's Your Problem

The Hornsby source has recently changed so this no longer works.
The search URL https://hscenquiry.hornsby.nsw.gov.au/Pages/XC.Track/SearchPropertyWaste.aspx now returns a "502 Bad Gateway" error.
The councils new interactive search is available at:
https://experience.arcgis.com/experience/41cd9f3895f64892b3699d02f60f980a
But the property ID value is no longer a valid data point.

### Source (if relevant)

hornsby_nsw_gov_au

### Logs

```Shell
no relevant logs
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

dw1562:
I just decided to take a look at this integration and have come up against the same issue.  I tried configuring manually in configuration.yaml based on the documentation and getting my Property ID from the council website.  But I still get stuck adding the integration as it still requests the config data.

revixisse:
You can easily get your property ID from this Hornsby Council property lookup (working as of April 17 2025):

![Image](https://github.com/user-attachments/assets/85425937-eba0-47b1-9ef0-d15473701d64)

https://[hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySearch.aspx?EnquiryListId=2&ERB=25961](https://hornsbyprd-pwy-epw.cloud.infor.com/ePathway/Production/Web/GeneralEnquiry/EnquirySearch.aspx?EnquiryListId=2&ERB=25961)

5ila5:
fixed in  #5317

This will be part of the next release or is available now as master (git) version 

[...]

PR #5317, opened 2026-01-08
Fixed: Hornsby Shire, NSW, Australia

jourdant:
The existing integration for Hornsby Shire relied on a custom API that had dependencies on council services that are no longer exposed externally. The provider has now been updated to use new API's published by the council directly. The following steps are followed:

1. Take a plain text address and search for an ID
2. Take the ID and search for the waste schedule
3. Download the waste and bulky waste PDFs and process to retrieve the schedules

I did have to bring in an additional PDF lib, pymupdf because it supports extracting shapes with their metadata (which I need to identify which days has which waste on the calendar. By processing the PDFs we can handle changes that may be published out of band (seems unlikely, but is possible).

Also generated some test cases to make sure the provider is working.

Happy for any feedback.

Update 09/01/2026: Corrected some details as per thread below.

### DIFF
--- a/custom_components/waste_collection_schedule/translations/fr.json
+++ b/custom_components/waste_collection_schedule/translations/fr.json
@@ -114,7 +114,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {
           "calendar_title": "Un nom plus lisible ou convivial pour le calendrier des déchets. Si aucun n'est fourni, le nom renvoyé par la source sera utilisé."
@@ -125,7 +125,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {}
       },
@@ -156,7 +156,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {
           "calendar_title": "Un nom plus lisible ou convivial pour le calendrier des déchets. Si aucun n'est fourni, le nom renvoyé par la source sera utilisé."
@@ -167,7 +167,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {}
       },
@@ -248,7 +248,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {
           "calendar_title": "Un nom plus lisible ou convivial pour le calendrier des déchets. Si aucun n'est fourni, le nom renvoyé par la source sera utilisé."
@@ -259,7 +259,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {}
       },
@@ -344,7 +344,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {
           "calendar_title": "Un nom plus lisible ou convivial pour le calendrier des déchets. Si aucun n'est fourni, le nom renvoyé par la source sera utilisé."
@@ -355,7 +355,7 @@
         "description": "Configurez votre fournisseur de services. {howto}Plus de détails : {docs_url}.",
         "data": {
           "calendar_title": "Titre du Calendrier",
-          "address": "Address"
+          "address": "Adresse"
         },
         "data_description": {}
       },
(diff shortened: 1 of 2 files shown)


jourdant:
Fixes #3529
Fixes #4029
Fixes #4783

jourdant:
@sh00t2kill @5ila5 would appreciate your review as I replaced the external API it was originally relying on. 

It's a pretty fundamental rewrite that uses the council's address lookup API, grabs the relevant waste schedule and parses the calendar PDFs for both waste and bulky waste.

We essentially are looking for the coloured circles on the calendar inside the PDF to identify which days of which type of waste, then turning those back into a proper array of clean dates.

sh00t2kill:
Just a couple of notes; the underlying issue is because council changed the waste collection website.

Parsing the PDF it gives much better output which means a better outcome. I never went down this road initially, as I wrote the custom API for my specific use case before I found this component.

FWIW The bulk collection dates are based on a week offset  The custom API is wrote used the pdf file name, which has the week of the year as part of the file name, and used that to determine the bulk waste date.

The PDF parsing method is more correct though.

jourdant:
Apologies @sh00t2kill, you're right. I've updated my original description to reflect that.

jourdant:
Just checking up on the status on this one. @5ila5 not wanting to rush you, but do you have an ETA when you might have a chance to check this one out?

custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
```diff
@@ -1,78 +1,476 @@
...
+    """Classify fill color into 'green' or 'yellow'."""
+    r, g, b = rgb
+    if g > 0.45 and r < 0.45 and b < 0.45:
+        return "green"
+    if r > 0.70 and g > 0.55 and b < 0.45:
+        return "yellow"
+    return "unknown"
+
+
+def _extract_events_from_weekly_pdf(pdf_bytes: bytes) -> list[Collection]:
```
5ila5, line 208: It sounds like using pymupdf instead of fitz is the better version here


https://pymupdf.readthedocs.io/en/latest/tutorial.html#note-on-the-name-fitz

custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
```diff
@@ -1,78 +1,476 @@
...
+                )
+
+    doc.close()
+    return entries
+
+
+def _extract_bulky_events_from_pdf(pdf_bytes: bytes) -> list[Collection]:
+    """Extract bulky waste dates from the bulky waste flyer PDF."""
+    try:
+        import fitz  # PyMuPDF
```
5ila5: same

tests/test_hornsby_nsw_gov_au.py
5ila5, line 1: I don't mind this but I doubt this serves much purpose over the test script

5ila5 (changes requested):
I don't like pdf parsing and think it should be avoided if possible. But if there is no other way I think it's fine. 

We had some problems in the past with introducing additional dependencies, so I'm hesitant to introducing new dependencies. But I would be willing to give this a try.

You need to add the dependency in custom_components/waste_collection_schedule/manifest.json so HA knows it needs to pull it

tests/test_hornsby_nsw_gov_au.py
jourdant: Removed the file as per comment

custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
jourdant: Resolved

custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py
jourdant: Good call - updated.

jourdant:
@5ila5 thank you for the review, you're right PDF parsing is not ideal, but it's the only way that Hornsby Shire offers right now with a full calendar view. I'll keep an eye on it when they publish the new schedule in July - I've set a reminder to look a couple of weeks out.

I've made the other changes as per your comments, plus removed the need for numpy too. So its just the one extra pdf dependency we're adding in this case (that will give us access to the vector shapes in the document).

5ila5:
This looks good for me

can you remove the mypy errors? The hassfest CI fail is not related to this PR but needs to be "fixed" for the whole project so you can ignore this.


```bash
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py: note: In function "_extract_events_from_weekly_pdf":
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:340: error:
Incompatible types in assignment (expression has type "dict[Any, Any]", variable
has type "Match[str] | None")  [assignment]
                for m in markers:
                ^
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:341: error:
Value of type "Match[str] | None" is not indexable  [index]
                    marker_rect = m["rect"]
                                  ^~~~~~~~~
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:342: error:
Item "str" of "str | Any" has no attribute "x0"  [union-attr]
                    cx = (marker_rect.x0 + marker_rect.x1) / 2.0
                          ^~~~~~~~~~~~~~
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:342: error:
Item "str" of "str | Any" has no attribute "x1"  [union-attr]
                    cx = (marker_rect.x0 + marker_rect.x1) / 2.0
                                           ^~~~~~~~~~~~~~
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:343: error:
Item "str" of "str | Any" has no attribute "y0"  [union-attr]
                    cy = (marker_rect.y0 + marker_rect.y1) / 2.0
                          ^~~~~~~~~~~~~~
custom_components/waste_collection_schedule/waste_collection_schedule/source/hornsby_nsw_gov_au.py:343: error:
Item "str" of "str | Any" has no attribute "y1"  [union-attr]
                    cy = (marker_rect.y0 + marker_rect.y1) / 2.0
                                           ^~~~~~~~~~~~~~
Found 6 errors in 1 file (checked 1 source file)
```

jourdant:
@5ila5 - apologies, I have run the mypy tests on my local env and have cleaned up the specific items that were highlighted in the hornsby provider (ignored the HA related ones).

5ila5:
Thanks for your contribution

PR #5352, opened 2026-01-13
Fix OLO 4-week cycle pattern calculation

vaind:
olo.sk started showing new schedules after new year and turns out some assumptions that worked in 2025 don't work in 2026 so I had to update the implementation. Too bad there're no docs and the actual implementation on the site is just obfuscated javascript...

## Summary
- Fixes 4-element frequency patterns like `[-,-,2,-]` which were incorrectly interpreted as "Nth weekday of month"
- The actual OLO behavior is a rolling 4-week cycle based on ISO week numbers: `(ISO_week - 1) % 4 == position`
- Unifies 2-element and 4-element pattern handling into a single code path

## Test plan
- [x] Verified pattern `[-,-,2,-]` returns Jan 13, 2026 (not Jan 20)
- [x] Verified pattern `[-,-,-,3]` returns Dec 24, 2025
- [x] All 27 unit tests pass
- [x] Existing biweekly patterns `[4,-]`, `[-,4]`, `[4,4]` work unchanged

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/olo_sk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/olo_sk.py
@@ -240,8 +240,7 @@ class Source:
                     byweekday=weekday(day - 1),
                 )
                 dates.extend(
-                    dt.date() for dt in rule
-                    if (dt.isocalendar()[1] - 1) % 4 == i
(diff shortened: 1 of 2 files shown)


5ila5:
Thanks for your contribution

PR #5362, opened 2026-01-16
Update description and ics link

starwarsfan:
New year, new ics link but a more stable one this year!


bbr111:
thanks for your contribution

Issue #5344, opened 2026-01-12
[Bug]: Silea (Italy) not working


mariellamarco:
### I Have A Problem With:

A specific source

### What's Your Problem

<img width="610" height="600" alt="Image" src="https://github.com/user-attachments/assets/dd4bee5c-d09e-429e-a688-d03ddcae847f" />

 **'list' object has no attribute 'values'**

### Source (if relevant)

silea_it

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

bbr111:
Testing source silea_it ...
  found 419 entries for Test_001
    2026-01-13 : INDIFFERENZIATO [None]
    2026-01-15 : UMIDO [None]
    2026-01-17 : CARTA E CARTONE [None]
    2026-01-19 : PLASTICA, LATTINE E TETRAPAK [None]
    2026-01-19 : UMIDO [None]
    2026-01-20 : INDIFFERENZIATO [None]
    2026-01-21 : VETRO [None]
    2026-01-22 : UMIDO [None]
    2026-01-24 : CARTA E CARTONE [None]
    2026-01-26 : PLASTICA, LATTINE E TETRAPAK [None]
    2026-01-26 : UMIDO [None]

returns normal results in test

mariellamarco:
How can i fix it? As you can see it doesn’t work.
What data do you use for testing? 

I tried also with example data in the .md but same

Thanks 

<img width="634" height="613" alt="Image" src="https://github.com/user-attachments/assets/fbd0ba5c-a968-4c74-b868-2b015e101534" />

Pilly81:
Hello, i have the same problem , i've noticed that my sensors are in the state "unknown" from 2 weeks...

[...]

PR #5417, opened 2026-01-30
silea_it use .json() instead of json.loads()

bbr111:
silea_it use .json() instead of json.loads()

to avoid errors in config_flow


bbr111:
fixex #5344

PR #5422, opened 2026-02-01
Fix: update oslokommune_no to use new API endpoint

### LINKED ISSUE #5421: [Bug]: oslokommune_no source returns null - API endpoint appears defunct
## Source

oslokommune_no

## Description

The `oslokommune_no` source is no longer returning any data. The underlying API endpoint at `https://www.oslo.kommune.no/xmlhttprequest.php?service=ren.search` returns `{"data": {"result": null}}` for all addresses, including the test addresses from the source code (Villa Paradiso at Olaf Ryes Plass 8, and Nåkkves vei 5).

## Steps to reproduce

1. Configure `oslokommune_no` via GUI config flow with valid parameters
2. The integration sets up without errors but shows no data
3. Manually querying the API confirms null response:

```bash
# Test address from source code (Villa Paradiso):
curl -s "https://www.oslo.kommune.no/xmlhttprequest.php?service=ren.search&street_name=Olaf%20Ryes%20Plass&street_code=24283&house_number=8"
# Returns: {"data": {"result": null}}

# Second test address from source code (Nåkkves vei):
curl -s "https://www.oslo.kommune.no/xmlhttprequest.php?service=ren.search&street_name=N%C3%A5kkves%20vei&street_code=22LO&house_number=5"
# Returns: {"data": {"result": null}}
```

## Root cause

Oslo Kommune appears to have migrated their waste collection service to a new platform. The old waste lookup page at `oslo.kommune.no/avfall-og-gjenvinning/avfallshenting/` now redirects to `oslo.kommune.no/avfall-og-gjenvinning/nar-hentes-avfallet/`, which references a new application at `renovasjonstjenester.app.oslo.kommune.no`.

The old `xmlhttprequest.php?service=ren.search` endpoint still responds with HTTP 200 but returns null data for all queries.

## Environment

- waste_collection_schedule version: 2.11.0
- Home Assistant version: 2026.1.3

## Additional context

This was previously working with YAML configuration. The issue is not related to GUI vs YAML setup — the upstream API itself no longer returns data.

Olen:
## Summary

Oslo Kommune has migrated their waste collection lookup service. The old API endpoint (`xmlhttprequest.php?service=ren.search`) now returns `{"data": {"result": null}}` for all addresses.

This PR updates the `oslokommune_no` source to use the new endpoint at `/actions/snap-lib-waste-complaint/search-by-address`.

The new endpoint returns the same data structure (`HentePunkts`, `Tjenester`, `Fraksjon`, `TommeDato`) — the only differences are:
- New URL path with `&`-separated parameters
- Response uses `response["result"]` instead of `response["data"]["result"]`
- Requires `Accept: application/json` header

Verified working with both test addresses (Villa Paradiso and Nåkkves vei).

Fixes #5421

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/oslokommune_no.py
@@ -80,21 +80,19 @@ class Source:
         self._point_id = point_id
 
     def fetch(self):
-        # The new API uses &-separated parameters in the path
-        url = (
-            f"{API_URL}"
-            f"&street={self._street_name}"
-            f"&number={self._house_number}"
-            f"&street_id={self._street_id}"
-        )
+        params = {
+            "street": self._street_name,
+            "number": self._house_number,
+            "street_id": self._street_id,
+        }
         if self._house_letter:
-            url += f"&letter={self._house_letter}"
+            params["letter"] = self._house_letter
 
         headers = {
             "Accept": "application/json",
         }
 
-        r = requests.get(url, headers=headers)
+        r = requests.get(API_URL, params=params, headers=headers, timeout=30)
         r.raise_for_status()
 
         today = datetime.date.today()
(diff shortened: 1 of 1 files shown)


bbr111 (commented):
thanks for your contribution

PR #5433, opened 2026-02-01
Update toronto_ca.py

constvariable:
Added recycle pickup from separate source URLs now that recycling is handled by a different company

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/toronto_ca.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/toronto_ca.py
@@ -15,10 +15,16 @@ TEST_CASES = {
     "324 Weston Rd": {"street_address": "324 Weston Rd"},
 }
 
-CSV_URL = "https://www.toronto.ca/ext/swms/collection_calendar.csv"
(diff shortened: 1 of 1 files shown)


bbr111:
the recently updated toronta_ca.py works for me.
please check if this PR is needed.

bbr111:
please reply.

constvariable:
My PR fixes specifically the "recycle" integration. I've tested that and it works. The Copilot integration might add some extra error checking, but I haven't had the time to validate all the suggested changes.

bbr111:
please check if the PR is needed. there is an new version of this file.

markvp:
@constvariable Can you confirm this is still needed against the current master? The Toronto source was updated recently and I want to make sure recycling is still broken before we merge this significant rewrite.

Specifically — does the current master `toronto_ca.py` fail to return Recycling entries for your address? If so, what address are you testing with so we can verify?

Also, the hardcoded `RECYCLE_PROPERTY_LOOKUP_CIRCMATONT` list of municipalities is fragile — if RecycleCoach adds/removes municipalities, this breaks. Is there an API endpoint that returns this list dynamically?

constvariable:
Seems that the original API has had Recycling added back. This PR is no longer necessary. Thanks for reaching out to check.

PR #5448, opened 2026-02-04
Add multiple date collection for bins (rushcliffe.gov.uk)

ribz:
Rushcliffe Borough Council now provides the next 2 dates each bin will be collected. This codes pulls both dates for the bins, rather than the first date in the list.


bbr111:
@ribz thanks for your contribution.

PR #5449, opened 2026-02-05
Fix hume_vic_gov_au

seedzero:
Fix API Access Denied error


seedzero:
Something else I had not considered with the refactor for the new API is that the bin names have changed (this also happened on Hume's website). "Rubbish" becomes "Garbage", and "Green Waste" becomes "Food and garden". "Recycling" remains the same. How should this typically be handled, as this would break sensors people have set up until they rename them?

bbr111:
@seedzero thanks for your contribution

PR #5487, opened 2026-02-15
Fix Coventry UK bin collection

tymscar:
The Coventry website showing the collection times changed and is not showing the dates inside `<li>`s so I had to do modify it to read the new format. I have tested my fork and it seems to work fine.


bbr111:
thanks for your contribution

tymscar:
Thank you! Do you know by any change @bbr111 when a new update will come out with this included?

PR #5029, opened 2025-12-04
Remove completely broken source - South Kesteven

MatStace:
[South Kesteven broke their bin lookup](https://github.com/mampfes/hacs_waste_collection_schedule/issues/4751), replacing it with a jpeg of the calendar. Until this is resolved, I propose removing South Kesteven entirely
This fixes #4751

Issue #4751, opened 2025-09-23
[Bug]: South Kesteven UK changed and broke everything


MatStace:
### I Have A Problem With:

A specific source

### What's Your Problem

South Kesteven have completely updated their bin collections, and have replaced their old checker with something which takes a postcode and returns only which day of the week the bins are collected. To find out which bin is collected on which Tuesday, the site links to an image of the annual weekly calendar.

I'd (sadly) suggest removing South Kesteven until they update their website with a functioning bin checker.

### Source (if relevant)

southkesteven_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML
No relevant configuration - new bin checker is at https://pre.southkesteven.gov.uk/skdcNext/tempforms/checkmybin.aspx
```

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

andymk3:
Agreed it needs to be removed, very annoying what they have done with the site update for the new collections.

I've fed Gemini the calendar image from the site, and eventually managed to get it to spit out an accurate .ics file of the collection dates which I have imported into Home Assistant if that helps anyone wanting an alternative.

SierraDee:
I'm in the same boat, very annoyed that their recent changes have broken the API fully. 
I would be very excited to have an accurate alternative, if available?

thanks in advance, Andymk3

PR #5505, opened 2026-02-17
Fix southkesteven_gov_uk after SKDC endpoint change

CraigBell:
### Summary
This updates the South Kesteven source from the legacy `pre.southkesteven.gov.uk/BinSearch.aspx` workflow to the current self-service form workflow.

- Old endpoint now fails (`500` / broken behavior).
- New implementation uses SKDC selfservice flow and parses returned collection rows.

### What changed
- Replaced old POST-to-`BinSearch.aspx` scraping logic.
- New flow:
  1. GET form metadata from `selfservice.southkesteven.gov.uk/renderform?...t=213`
  2. POST selected address ID to `RenderForm`
  3. Parse collection rows from `table.Alloy-table`
- Added fallback:
  - If no rows and input is not a `U...` address ID, call `core/addresslookup` with the provided value (e.g. postcode), pick first address ID, retry fetch.
- Kept duplicate-event dedupe (`date`, `type`).
- Updated `TEST_CASES` to values compatible with the new flow (postcode input).

### Why this is safe
- Uses the current SKDC-hosted flow that is actively returning collection dates and bin types.
- No changes to unrelated sources.
- Parser is conservative: only rows with parseable date + type are used.

### Reproduction (before)
- Source `southkesteven_gov_uk` fails due to the old `BinSearch.aspx` endpoint no longer providing usable data.

### Validation (after)
- Address lookup endpoint returns structured data:
  - `POST https://selfservice.southkesteven.gov.uk/core/addresslookup`
- RenderForm flow returns collection table rows with concrete dates and bin types.
- Local compile check passes for updated source file.

### Notes
- Existing issue context:
  - https://github.com/mampfes/hacs_waste_collection_schedule/issues/4751
  - https://github.com/mampfes/hacs_waste_collection_schedule/pull/5029

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/southkesteven_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/southkesteven_gov_uk.py
@@ -1,16 +1,21 @@
 import logging
-import re
-from datetime import datetime, timedelta
+from datetime import datetime
 
 import bs4
 import requests
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
 URL = "https://southkesteven.gov.uk"
+FORM_URL = (
+    "https://selfservice.southkesteven.gov.uk/"
+    "renderform?k=2074C945A63DDC0D18F1EB74DA230AC3122958B1&t=213"
+)
+RENDER_URL = "https://selfservice.southkesteven.gov.uk/RenderForm"
+ADDRESS_LOOKUP_URL = "https://selfservice.southkesteven.gov.uk/core/addresslookup"
 TEST_CASES = {
-    "Long Bennington": {"address_id": 33399},
-    "Bourne": {"address_id": "7351"},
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/southkesteven_gov_uk.py
```diff
@@ -25,87 +30,81 @@
...
+            if icon is None:
+                lower = raw_type.lower()
+                if "refuse" in lower:
+                    icon = ICON_MAP["black"]
+                elif "recycling" in lower or "paper" in lower:
+                    icon = ICON_MAP["gray"]
+                elif "green" in lower:
+                    icon = ICON_MAP["green"]
+                elif "purple" in lower:
+                    icon = ICON_MAP["purple"]
```
bbr111, line 87: Consider to add default icon

Issue #5485, opened 2026-02-13
[Bug]: sutton council 503 server error


rddc1975:
### I Have A Problem With:

A specific source

### What's Your Problem

Sutton council ran some maintenance a couple days ago and since the integration no longer works.

### Source (if relevant)

Sutton_gov_uk

### Logs

```Shell
The source returned an invalid response: "503 Server Error: Service Unavailable for url: https://waste-services.sutton.gov.uk/waste/444####". Please check the provided arguments and try again.
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

AceTallPaul:
I also have an issue with Sutton.  As of a few days ago, it is not connecting.  The url is operational, and I can see my waste schedule manually on Sutton's webpage.  It does not seem to be an availability issue. Some kind of issue where they are blocking automated requests?

My Log is 
This error originated from a custom integration.

Logger: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Source: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:160
integration: Waste Collection Schedule (documentation)
First occurred: 11:39:19 AM (1 occurrence)
Last logged: 11:39:19 AM

[...]

PR #5510, opened 2026-02-18
Update User-Agent header for requests

cunners:
Fixes 503 error by changing user-agent in header.

#5485


cunners:
Ok. It was failing when the PR was submitted, for multiple users. See #5485.

bbr111:
but now it is working, with the old header.
I think we can close this.

PR #5514, opened 2026-02-20
Changed MidCoast Council schedule to not reset General Waste prematurely

AkimboDadz:
Removed reset on general waste cycle when days_ahead reaches 0. This was causing the next general waste pickup to jump to the next week once the next applicable day was reached.


bbr111:
@AkimboDadz thanks for your contribution

PR #5534, opened 2026-02-23
Avoid blocking platform import warning during entry setup

CraigBell:
## Summary
Pre-import calendar and sensor platform modules in the executor before calling async_forward_entry_setups.

## Why
Home Assistant warns about a blocking import_module call from init_ui.py during startup:
- Detected blocking call to import_module ... custom_components/waste_collection_schedule/calendar

Preloading platform modules via hass.async_add_executor_job(importlib.import_module, ...) keeps import work off the event loop.

## Change
- Add importlib import
- In async_setup_entry, import each platform in PLATFORMS via executor
- Keep the existing async_forward_entry_setups flow unchanged

### DIFF
--- a/custom_components/waste_collection_schedule/init_ui.py
+++ b/custom_components/waste_collection_schedule/init_ui.py
@@ -82,7 +82,8 @@ async def async_setup_entry(hass: HomeAssistant, entry: ConfigEntry) -> bool:
     hass.data.setdefault(const.DOMAIN, {})[entry.entry_id] = coordinator
 
     # Pre-import platform modules in executor to avoid blocking import warnings
-    # when forwarding setups from the event loop.
+    # during async_forward_entry_setups, e.g.:
+    # "Detected blocking call to import_module ... waste_collection_schedule/calendar"
     for platform in PLATFORMS:
         await hass.async_add_executor_job(
             importlib.import_module, f"{__package__}.{platform}"


bbr111:
Thanks for the PR and for addressing the blocking import warning.

After reviewing the platform modules (calendar.py and sensor.py), both are import-safe and do not perform any IO or heavy work at module import time. They only define classes, schemas, and functions, which makes the import effectively lightweight.

The warning:

Detected blocking call to import_module ...

is triggered because Home Assistant performs import_module() inside the event loop when forwarding entry setups. This is part of HA core behavior and does not necessarily indicate actual blocking work inside the module.

Since the platform modules themselves are IO-free, pre-importing them via hass.async_add_executor_job(importlib.import_module, ...) does not address a real blocking issue — it only suppresses the warning by warming the Python import cache.

From an architectural perspective, platform loading should remain the responsibility of the Home Assistant loader, and integrations generally should not manually manage platform imports.

Given that the modules are import-safe, I would suggest removing the manual pre-import and relying on the standard async_forward_entry_setups() behavior.

If there were actual IO during import, refactoring that out of the module top-level would be the preferred solution instead.

Let me know what you think.

5ila5:
I generally think we would address warnings shown by HA to the user (even if it is just to keep warnings out of the logs [and our issues])

But I could not reproduce this warning (HA 2026.4.0.dev0) I only see the error from (https://github.com/mampfes/hacs_waste_collection_schedule/pull/5437#issuecomment-3977053801)

I could not find any online examples that preload the platforms.

If you have any other projects / examples / docs that mention this approach could you please share them?

bbr111:
@5ila5 I tried deep dived into the problem.
If I see this right, the warning messages should only come once.
At the first start with the integration.
if some modules needs to be loaded and are not in module cache from HA.

i also didn't see the warning in any version.

markvp:
Can anyone confirm if the recent release has resolved this? I think there was a duplicate ticket that was merged.

markvp:
This fix is already present in master (parallel pre-import using `asyncio.gather`). Closing as resolved.

PR #5548, opened 2026-02-25
Update Arun to use Cloud9 API

whi-tw:
Arun, Rugby and North Herts all have Cloud9 Apps, and therefore can use a standard backend.

UPRN is preferred, but address search is possible, using the address lookup functionality.

Paging @LukeT and @Zepheus who've previously contributed to rugby and northherts.


bbr111:
headers=dict(BASE_HEADERS) everywhere

You rebuild dicts multiple times. Not expensive, but a bit noisy.

Suggestion: store self._headers = {…} in __init__ and copy only when needed.

Issue #3464, opened 2025-01-07
[Source Request]: Valcobreizh [FR]


mattpeufpeuf:
### Municipality / Region

France, Bretagne

### Collection Calendar Webpage

https://dechets.valcobreizh.fr/mQLmrtBQVM/calendar/

### Example Address

1 Rue de Rennes  35850 Irodouër

### Collection Data Format

As html on a webpage

vinzd:
Hey @mattpeufpeuf, as you are providing the publidata ID may I assume that you already tried the Publidata generic provider? Is it working for you?

mattpeufpeuf:
![image](https://github.com/user-attachments/assets/0bbe0771-6ab7-42a5-9890-74caf2dbf0c4)

It works (but not perfectly) by modify  the file `custom_components\waste_collection_schedule\waste_collection_schedule\source\publidata_fr.py`

The blue calendar is with the integration.
The red calendar is manual, do by me.

Here you can see that recycle bin task doesn't work well ( once every two weeks)
When there are bankday the trash schedule is report by one day.

vinzd:
Hey, I have found the issues and created a fix for your use case => https://github.com/vinzd/hacs_waste_collection_schedule/commit/a37193f0a28847a5c9619d47244c536cdaf35fd4

However, it now breaks other sources so I need to figure those out before submitting a MR

PR #5555, opened 2026-02-28
improve publidata_fr to accomodate Valcobreizh

vinzd:
Notable change in the source: handle date ranges with year, such as `2024 Jan 01-2024 May 12`.

Should close https://github.com/mampfes/hacs_waste_collection_schedule/issues/3464 (and probably a few other issues)


bbr111:
thanks for your contribution

PR #5569, opened 2026-03-03
aw_harburg_de: normalize yellow waste type and harden district matching

strausmann:
## Summary
- normalize provider type naming in aw_harburg_de (Gelber Sack -> Gelbe Tonne)
- harden district matching in parse_level by normalizing whitespace/NBSP differences
- fix district value in parse error message
- document the naming change and backward-compatible alias example

## Why
Landkreis Harburg changed naming to Gelbe Tonne, which can break existing filters expecting Gelber Sack.
Additionally, strict string matching can fail for some area names (e.g. Neu Wulmstorf variants) due to whitespace/encoding differences.

## Validation
- live fetch for Seevetal / Maschen (nördlich der Autobahn A 39) returns expected types including Gelbe Tonne
- live fetch for issue case Neu Wulmstorf (nur südlich der Hauptstraße einschließlich Hauptstraße gesamt, ohne Daerstorf, Wulmstorf) succeeds

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_harburg_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_harburg_de.py
@@ -137,7 +137,9 @@ class Source:
                 option_text = option.text
                 option_value = option.get("value")
                 level_ids[option_text] = option_value
-                level_ids_normalized[normalize_district_name(option_text)] = option_value
+                level_ids_normalized[normalize_district_name(option_text)] = (
+                    option_value
+                )
 
         if level_ids == {}:
             raise Exception(f"Error: Level {level} Dictionary empty")


strausmann:
> ## Pull request overview
> This PR updates the `aw_harburg_de` source to keep collection types stable after Landkreis Harburg renamed “Gelber Sack” to “Gelbe Tonne”, and improves robustness of district matching by normalizing whitespace differences (including NBSP).
> 
> **Changes:**
> 
> * Normalize waste type names returned from the provider (`Gelber Sack` → `Gelbe Tonne`).
> * Harden `parse_level` district matching by normalizing whitespace / NBSP in option text and user input.
> * Update documentation with the naming change and a backward-compatible alias example.
> 
> ### Reviewed changes
> Copilot reviewed 2 out of 2 changed files in this pull request and generated no comments.
> 
> File	Description
> doc/source/aw_harburg_de.md	Documents the new `Gelbe Tonne` type and shows how to alias it back to `Gelber Sack`.
> custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_harburg_de.py	Adds type normalization and whitespace-normalized district lookup; fixes the district value used in the error message.
> Comments suppressed due to low confidence (1)
> **custom_components/waste_collection_schedule/waste_collection_schedule/source/aw_harburg_de.py:156**
> 
> * The final `if district not in level_ids:` check is redundant here: at this point the function has already returned for the only successful cases, so this condition will always be true and the exception will always be raised. Consider simplifying to an unconditional `raise` (or `else: raise`) after the normalized lookup to make the control flow clearer.
> 
> ```
>         if district not in level_ids:
>             raise Exception(
>                 f"Error: District {district} is not in the dictionary: {level_ids}"
>             )
> ```
> 
> 💡 [Add Copilot custom instructions](/mampfes/hacs_waste_collection_schedule/new/master?filename=.github/instructions/*.instructions.md) for smarter, more guided reviews. [Learn how to get started](https://docs.github.com/en/copilot/customizing-copilot/adding-repository-custom-instructions-for-github-copilot).

Thanks for the review.
I addressed the Copilot suggestion by simplifying the redundant conditional in parse_level to a direct raise after the successful return paths.

Validation was rerun after the change (no file errors, and live fetch for Harburg still succeeds).

bbr111:
I'm not sure if we should do this.
Another possible solution would be to fix the filters in the config.

If we go this route, we would need to update the source every time the provider changes something.

However, your idea is interesting. Maybe we could introduce some kind of general normalization dictionary per country.

There was also an idea to normalize the icon_map. Maybe we could combine both approaches (see PR #2814).

@5ila5 I’d appreciate your thoughts on this.

5ila5:
> Landkreis Harburg changed naming to Gelbe Tonne, which can break existing filters expecting Gelber Sack.
I don't understand why changing `Gelber Sack` to `Gelbe Tonne` should fix this.

Generally I'm against modifying collection type names on a per-source basis, as this is an additional point where we would need to track potential changed names of reported collection types.

I think normalizing on a global view could make sense. My thoughts on this:

- \+ 
  - may prevent breaking sensors if the service provider change collection type name
- \-
  - may break sensors for existing users once (if we do the change) (I think this is acceptable if properly announced)
  - may not be possible to know what the normalized type should be everywhere (e.g. Germany "Grüne Tonne" may be organic or paper waste depending on region)
  - may cause confusion????? Maybe there are some regions that have two different collections that may be treated the same by our normalization (not sure though)

PR #5577, opened 2026-03-05
Generate source owner mapping from source module metadata

bbr111:
This pull request introduces automated source ownership mapping and notification for bug reports, improving issue routing and maintainership transparency. The main changes include generating a `.github/source_owners.json` file from source metadata, updating documentation to explain the workflow, and adding a GitHub Action that pings and assigns owners when a bug issue is filed.

**Automated source ownership mapping and notification:**

* Added generation of `.github/source_owners.json` in `update_docu_links.py`, mapping source modules to their owners using the new optional `SOURCE_CODEOWNERS` variable in source files.
* Introduced `SOURCE_CODEOWNERS` as a documented variable for source files, allowing maintainers to specify GitHub usernames for ownership.

**Issue routing workflow and documentation:**

* Added a new GitHub Action workflow (`.github/workflows/notify-source-owners.yaml`) that reads the `Source (if relevant)` field from bug reports, pings the mapped owners, and assigns them to the issue.
* Updated `doc/contributing.md` to explain the source ownership mapping and notification workflow, including how ownership is determined and which files are involved.

**Supporting code changes:**

* Extended `update_docu_links.py` to extract, store, and propagate `SOURCE_CODEOWNERS` from source modules throughout the metadata generation process.


5ila5:
The code looks good. But I have some general suggestions:

- I think ICS sources should have an owner too
- I think you should check somewhere if the `SOURCE_CODEOWNERS` is a sensible value. 
  - You currently just get it and convert it to a set. If the value is for some reason a string and not a list it should probably at least warn the user and not create a set like `{'l', 'a', 'i', '5'}` (for `set("5ila5")`)
  - If the update_docu_links.py does not right out fail if the value is bullshit you should probably add a test to the tests/test_source_components.py, so CI pipeline warns us before merging.

Some more other things I thought about (let me know what you think)

- Users seem to rarely fill out that value with the correct value. Do you think it would be a good idea to make it a drop-down menu also updated by the update_docu_links.py script? It would be an additional file new sources need to touch, but it could make this easier
- Is there any way to trigger this manually? If the bug report does not specify the source, so that we can specify the source and if a code owner exists he gets pinged?




I just noticed the whole script fails if it is not called from the project root (so `cd custom_components && python ../update_docu_links.py` fails) I changed some stuff to use the `pathlib.Path` instead of these relative paths in #5583.
I know it was probably not the best Idea to work on this simultaneous to yours, but mine can wait until this is thought, and I can rebase doing the same to your code. (If you really want you obviously can merge mine and rebase yours)

bbr111:
@markvp should we close this here and implement it your new BaseSource Concept?

markvp:
> @markvp should we close this here and implement it your new BaseSource Concept?

if we put this in draft, after we merge the other branch we could re-base this and re-apply it

doc/contributing_source.md
```diff
@@ -200,6 +200,7 @@ The script iterates through all source files and extracts some meta information
 | HOW_TO_GET_ARGUMENTS_DESCRIPTION | Dict | [Optional] Description of how to get the arguments, will be shown in the GUI configuration form above the input fields, does not need to be translated in all languages. |
 | PARAM_DESCRIPTIONS | Dict | [Optional] Description of the arguments, will be shown in the GUI configuration below the respective input field. |
 | PARAM_TRANSLATIONS | Dict | [Optional] Translate the arguments, will be shown in the GUI configuration form as placeholder text. Some common parameters will be automatically translated if you do not provide this |
+| SOURCE_CODEOWNERS | List of Strings | [Optional] List of GitHub users (for example `"@your-user"`) who maintain this source. Used by `update_docu_links.py` to generate `.github/source_owners.json` for issue routing. |
```
markvp: we should add this to the source development agents as well

markvp:
alternatively we could merge this one first and then i will merge/rebase it into the other branch. i think if we push the other PR that's probably a "major version" change since we are significantly changing the interface (even though we are maintaining backwards compatibility, we would want new sources implemented in the new structure).

markvp (approved):

markvp:
@bbr111 - shall we merge this in?

PR #5583, opened 2026-03-06
make update_docu_links independet on the location it is called from

5ila5:
This can wait until #5577 is through

### DIFF
--- a/update_docu_links.py
+++ b/update_docu_links.py
@@ -8,6 +8,7 @@ import json
 import re
 import site
 from functools import lru_cache
+from os import PathLike
 from pathlib import Path
 from types import ModuleType
 from typing import Any, Callable, Tuple, TypedDict, TypeVar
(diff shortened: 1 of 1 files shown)


bbr111:
Thanks for making the script independent of the current working directory. The general approach of resolving paths relative to __file__ looks good.

There is one cross-platform regression compared with the previous implementation:

@property
def relative_file_path_name(self):
    return "/" + str(self._file.relative_to(PROJECT_ROOT))

The previous implementation constructed documentation links explicitly with /, for example:

filename = f"/doc/source/{file}.md"

With the new implementation, str(Path(...)) uses the platform-specific path separator. On Windows this can generate links such as:

/doc\source\example.md

Since this value is used for Markdown links and GitHub documentation URLs, it should always use POSIX separators:

@property
def relative_file_path_name(self) -> str:
    return "/" + self._file.relative_to(PROJECT_ROOT).as_posix()

Apart from this cross-platform issue, the path refactoring looks good.

bbr111:
but maybe we should wait until the BaseSource PR is merged.

PR #5608, opened 2026-03-13
Fixed redbridge_gov_uk parser for new PDF format

bbodien:
As mentioned in #5393, Redbridge council now renders a PDF rather than HTML. 

I've updated the parser in `redbridge_gov_uk.py` to use pypdf in a similar vein to `mpo_krakow_pl.py`. 

It's a little ugly as the PDF calendar format lists multiple collections types per date, and the structure of the PDF text isn't very helpful.

I've also added the Food type since the council has rolled out food waste collections as of the start of 2026.


bbr111:
thanks for your contribution

PR #5651, opened 2026-03-24
Fixed timestamp for Calgary source

D-Jeffrey:
Corrected the timestamp query match for the source which only has 3 decimal seconds digits for resulting time query.


D-Jeffrey:
Correction, it should have beening with the %f and without the Z, as the source timestamp does not have the Z in the timestamp.

PR #5689, opened 2026-04-01
Fix gänsendorf not filtering for street and more

zaubara:


markvp:
Please resolvev the copilot comments

zaubara:
Done.

PR #5707, opened 2026-04-01
mijnblink: Move to Ximmio service

xesxen:
MijnBlink has moved over to Ximmio as their service provider. The old API integration doesn't return data anymore.

The UUID was retrieved from the "Mijn Blink Afvalkalender" link on https://mijnblink.nl/wanneer.


markvp:
Please review and address copilot feedback

PR #5724, opened 2026-04-03
Fix tonnenleerung.de parser for new JSON format

Boerni89:
## Summary
Fixes the tonnenleerung.de source, which stopped working after a website update.

## Problem
The website tonnenleerung.de has changed its data structure:
- **Old**: JavaScript arrays in HTML (`blueArray = [...]`)
- **New**: JSON data in the `<script type="application/json">` tag

## Changes
- The parser now supports the new JSON format
- Fallback to the old format for compatibility
- Consistent mapping: `braun` → `Bio` (as in the old format)

## Tests
- Tested with `nd-sob/neuburg-donau/karlsplatz/`
- All 5 bin types are recognized correctly

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/tonnenleerung_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/tonnenleerung_de.py
@@ -1,8 +1,8 @@
 import json
-import re
 from datetime import datetime
 
 import requests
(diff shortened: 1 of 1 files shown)


markvp:
I've approved... But... What is the value of retaining the old method?

Boerni89:
Maybe it's nostalgia? No, it's better to remove the outdated method entirely.

PR #5831, opened 2026-04-06
EVV Völklingen: Fixing duplicated paper collection dates

Synthenses:
Due to different collection dates for paper containers in various sizes, there were incorrect entries in the calendar for containers you don’t need. An optional field “behaelter” has been added so you can filter by the container size you have.


Synthenses:
> Thank @Synthenses I'll clean up the docs and merge shortly



> Thank @Synthenses I'll clean up the docs and merge shortly

Sorry tried already to fix the translation, hopefully it does not interrupter you.

PR #5934, opened 2026-04-09
Fix ecoharmonogram_pl crash when street has mixed sides and no matchers

czeslavo:
## Summary

- When the API returns streets with both `sides=""` (universal) and non-empty `sides` values, the source incorrectly raised `SourceArgumentRequiredWithSuggestions` mid-loop, even though a valid universal street was already collected.
- The raise is now deferred until after the loop and only fires when truly no street could be matched.
- Also fixes the return value to use the filtered list instead of the original unfiltered dict.
- Adds a `TEST_CASES` entry reproducing the bug: a street where the API returns a mix of empty and non-empty `sides` values with no `additional_sides_matcher` configured.

## Testing

- Verified the new test case crashes on the old code with the exact error I got in HA.
- Ran `test_sources.py -s ecoharmonogram_pl` — all existing cases passed, new case returned entries successfully.

_Debugging and fix assisted by Claude Code._

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ecoharmonogram_pl.py
@@ -188,6 +188,12 @@ TEST_CASES = {
         "additional_sides_matcher": "Заміська забудова",
         "language": "uk",
     },
+    "Sławków (mixed sides: empty + non-empty, no matcher)": {
+        "town": "Sławków",
+        "street": "Jagiellońska",
+        "house_number": "32",
+        "additional_sides_matcher": "", # Available sides are "Zabudowa wysoka" and "" (empty string)
+    },
 }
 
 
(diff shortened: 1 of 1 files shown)


markvp (commented):
Good fix — the deferred raise and the `return {**streets, "streets": to_return}` fix are both clearly needed.

I've rebased on latest master and pushed one small refinement: when `to_return` is empty **and** no `additional_sides_matcher` was provided, we now raise `SourceArgumentRequiredWithSuggestions` instead of `SourceArgumentNotFoundWithSuggestions`.

**Why:** The "Required" variant produces "Argument 'additional_sides_matcher' must be provided" which is accurate — the user hasn't provided a value. The "NotFound" variant produces "We could not find values for ... with the value ''" which is confusing since the user never supplied a value to search for. `SourceArgumentNotFoundWithSuggestions` is still used when a matcher *was* provided but didn't match any street, which is the correct semantic there.

Both exceptions are handled identically by the config flow (same `SourceArgumentSuggestionsExceptionBase` handler), so the suggestions dropdown still works either way — this is purely about the user-facing error message clarity.

Issue #5946, opened 2026-04-10
[Bug]: Wyre Forest District Council - Cannot add if collection is today


MeltonCG:
### I Have A Problem With:

A specific source

### What's Your Problem

If you add this source and the collection day is today, You get the error that is seen in the attached 

<img width="1161" height="593" alt="Image" src="https://github.com/user-attachments/assets/a2f08075-45cc-4123-a82c-615a65467c71" />

### Source (if relevant)

wyreforestdc_gov_uk

### Logs

```Shell
no relevant logs
```

### Relevant Configuration

```YAML

```

### Checklist Source Error

- [x] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [x] Checked that the website of your service provider is still working
- [x] Tested my attributes on the service provider website (if possible)
- [x] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [x] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [x] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [x] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

PR #5947, opened 2026-04-10
Fix for wyreforestdc_gov_uk value returned if adding source on curren…

MeltonCG:
This should now cope when trying to add the source on a collection day


MeltonCG:
This is in relation to the issue i raised #5946

markvp (approved):

PR #5971, opened 2026-04-11
Update hvcgroep_nl.py

kay1010100:
Line 296: "self.house_letter" pointed to "postal_code" instead of "house_letter"

Noticed this after my house number got a letter in adition to my house number.


kay1010100:
This bug has been in here for ~3 years, it can only be triggered if 1. your address has a letter AND 2. the first fetch returns 2 options.

PR #6000, opened 2026-04-13
Replace requests with curl_cffi for HTTP requests

davewins:
Teignbridge server uses TLS fingerprinting which blocks Python's default TLS client hello, so curl_cffi is required

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/teignbridge_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/teignbridge_gov_uk.py
@@ -1,6 +1,6 @@
 from datetime import datetime
 
-import requests
+from curl_cffi import requests
 from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


markvp (changes requested):
Thanks for the fix. The import swap is the right approach, but to match the standard pattern used by all other `curl_cffi` sources in this project, the code should use a `Session` with `impersonate`. For example:

```python
session = requests.Session(impersonate="chrome124")
r = session.get(BIN_URL, params={"uprn": self._uprn})
```

Every other source using `curl_cffi` in this codebase follows this pattern (see `blaby_gov_uk.py`, `stockton_gov_uk.py`, `gateshead_gov_uk.py`, etc.). Without `impersonate`, `curl_cffi` may not use a browser-like TLS fingerprint, which defeats the purpose of switching from `requests`.

davewins (commented):
Updated to include impersonation on the request

markvp (approved):

Issue #4871, opened 2025-10-28
[Source Request]: North Yorkshire Council (Richmondshire)


wozza999:
### Municipality / Region

UK

### Collection Calendar Webpage

https://www.northyorks.gov.uk/bin-calendar/lookup

### Example Address

Silver Street, DL10 6LT, UK

### Collection Data Format

As html on a webpage

### Additional Information

Please note Richmondshire District Council has been taken over by North Yorkshire and the current integration no longer works (old system migrated to North Yorkshire). I'm assuming you will be able to clone an existing integration e.g. Scarborough and tweak to work for Richmondshire addresses?

arl007:
Any news on this? 
My errors are attached
Just updated to latest version and it’s still not working.
Many Thanks
Andrew

<img width="1170" height="2532" alt="Image" src="https://github.com/user-attachments/assets/22ded6b7-c855-4266-b9d2-63b34c9bb95c" />

wozza999:
Hi I've cloned Scarborough's and tweaked to the correct endpoint and  managed to get it working partially but I'm not a coder so if someone else can fix it it would be much appreciated.  Here is the code to copy over the current Richmond_gov_uk.py in the source folder. Better than nothing I guess! You may have to reset the config in the GUI to get it going again. Just seams to get the Recycling data only though.

```
from datetime import datetime

import requests
from bs4 import BeautifulSoup
from waste_collection_schedule import Collection  # type: ignore[attr-defined]

TITLE = "North Yorkshire Council - Richmondshire"
DESCRIPTION = "Source for North Yorkshire Council - Richmondshire."
URL = "https://northyorks.gov.uk"
TEST_CASES = {
    "Test 1": {"uprn": 100050441864},
    "Test 2": {"uprn": "100052206243"},
}


ICON_MAP = {
    "Household waste": "mdi:trash-can",
    "Garden": "mdi:leaf",
    "Recycling": "mdi:recycle",
}


API_URL = "https://www.northyorks.gov.uk/bin-calendar/Richmondshire/results/{uprn}/ajax?_wrapper_format=drupal_ajax"


class Source:
    def __init__(self, uprn: str | int):
        self._uprn: str = str(uprn)

    def fetch(self):
        r = requests.post(API_URL.format(uprn=self._uprn))
        r.raise_for_status()

        html = None
        for res in r.json():
            if "data" in res and isinstance(res["data"], str):
                html = res["data"]
                break
[...]

PR #6003, opened 2026-04-13
Update richmondshire_gov_uk.py

wozza999:
Changed to new Council #4871


markvp (approved):
Nice work updating Richmondshire to the new North Yorkshire Council endpoint — welcome to the project!

I've rebased on master and made a couple of small tweaks:
- Switched the icon lookup to use prefix matching so it handles bin types with trailing text (e.g. "Garden waste" still matches "Garden")
- Added `timeout=30` to the POST request
- Regenerated derived files (sources.json, translations, etc.)

Looks good to merge.

PR #6004, opened 2026-04-13
Fix RegioEntsorgung street matching for whitespace variants

taker218:
This PR fixes street validation for regioentsorgung_de when provider data contains non-breaking spaces (\u00A0) but users enter normal spaces.

What changed:

- Added whitespace-normalized matching in regioentsorgung_de for city/street/house number option resolution.
- Kept request flow unchanged by still sending the provider’s original option value after matching.
- Added a regression test for the street matching case (Krefelder Straße vs Krefelder\u00A0Straße).
- Updated source documentation with a short note about whitespace normalization behavior.

Why:
RegioEntsorgung currently returns some street names with non-breaking spaces.
Exact string matching caused valid user input to fail with “street not found”.
This change makes matching robust without weakening validation.


tests/test_regioentsorgung_de.py
markvp, line 1: please use TEST_CASES in the module for testing instead of custom test files unless absolutely necessary.

markvp (changes requested):

taker218:
Thanks for the review. I removed the custom test file and switched to module-level TEST_CASES as requested.
I added a "Krefelder Straße" testcase in regioentsorgung_de.py to cover the whitespace normalization scenario directly via source test cases.

markvp (approved):

Issue #4818, opened 2025-10-15
[Source Request]: mijnafvalzaken The Netherlands


dennuzz:
### Municipality / Region

Netherlands - Uitgeest

### Collection Calendar Webpage

https://www.mijnafvalzaken.nl

### Example Address

1911LB 14

### Collection Data Format

As html on a webpage

kay1010100:
Note: works almost exacly the same as hcvgroup, it's just a diffrent webaddress and ID's

markvp:
Thanks for the tip about the similarity to hcvgroup. I've investigated the mijnafvalzaken.nl API and found the endpoints:

- Address lookup: `/adressen/{postcode}:{huisnummer}`
- Collection dates: `/rest/adressen/{bagid}/ophaaldata`
- Calendar: `/rest/adressen/{bagid}/kalender/{year}`

However, the address lookup returns empty results for Uitgeest postcodes (e.g. `1911LB:14`), despite the postcode being validated as valid. This suggests the address database for Uitgeest may not be fully populated yet on mijnafvalzaken.nl.

Could you confirm:
1. Does the website at mijnafvalzaken.nl actually return collection dates when you enter an address?
2. If so, could you share an example address (postcode + house number) that works, so we can test the API?

dennuzz:
HI,

Thanks for looking into this.

Yes, the website mijnafvalzaken.nl returns information when I enter the
postcode and number:

[image: image.png]

1911LB, number 14 is a valid address to be used for testing.

Thanks again.

Op ma 13 apr 2026 om 11:23 schreef markvp ***@***.***>:

kay1010100:
It worked yesterday, I'll check when I'm home. I'm nearly done creating a new source for it.

kay1010100:
I'll push it later this week together with another bug fix for hvcgroep_nl

dennuzz:
Awesome, thanks for the update!

Op ma 13 apr 2026 om 13:09 schreef kay1010100 ***@***.***>:

kay1010100:
@markvp maybe I'm making a silly mistake, but within my dev container in VSCode I keep running into ModuelNotFoundError: No module named 'yaml'.
Would you have  any idea what I'm doing wrong?
I have a clean instalation of python3.
I have docker already installed because of other projects.

I used to do JavaScript a lot, but this is one of the first time's I'm doing python.

P.S. the API seems to be working just fine for me.

kay1010100:
[...]

PR #6011, opened 2026-04-13
Mijnafvalzaken_nl [New Source]

kay1010100:
https://github.com/mampfes/hacs_waste_collection_schedule/issues/4818
New source for [Mijn Afval Zaken](https://www.mijnafvalzaken.nl/).
For municipalities: Bergen, Uitgeest, Castricum & Heilo (BUCH municipalities)

Note: ./update_docu_links.py has not been run, but updated manualy.
Failed to run the script, but it should be fine.

Tested using the "test_sources.py" & tested my branch on my personal Home Assistant without issue.

------------------

This also contains a small fix to "hvcgroep_nl" where an if statement tries to compair lower and higher case letters. (Added ".lower()" to the check.)


kay1010100:
I don't know why, but I can't seem to run "./update_docu_links.py" I've updated the docs manualy and tested in my personal Home Assistant.

markvp (changes requested):
Thanks for the contribution! A couple of things:

1. **The new `mijnafvalzaken_nl` source appears to use the same backend platform as `hvcgroep_nl`** — the API endpoints, data structures, and Source class are essentially identical. Could you clarify why this couldn't be added as another entry in `hvcgroep_nl`'s `SERVICE_MAP` instead of a new source file? If there's a difference I'm missing, happy to hear about it.

2. **Please keep PRs atomic** — the `.lower()` fix for `hvcgroep_nl` is a good standalone bugfix and would be straightforward to merge on its own. Mixing unrelated fixes with new source additions makes review harder. Could you split that into a separate PR?

kay1010100:
Because when people of this region are looking for their services they will be looking for "Mijn Afval Zaken" they waste service is done by the local authority's them selfs and "HVC" is a public company totaly not connected.

Afaik. They just bought their systems.

kay1010100:
I myself work for these local authorities.

kay1010100:
Or can it still be found if placed under the hvc group?

markvp:
it can still be found - the source can "provide" data for multiple locations.

See `SERVICE_MAP` in the hvcgroep_nl.py

kay1010100:
I'll make those changes later today then, np.

kay1010100:
@markvp Everything should be in order now.

- Moved everything to hvcgroep_nl.md, hvcgroep_nl.py, updated README.md, sources.json & info.md.

- Also this still includes a fix for house_letter check. Sometimes it tried compairing lowercase to uppercase so I addad a ".lower()" to the check.


i've tested my branch on my home assistant.

markvp (approved):

PR #6037, opened 2026-04-15
Fix avfallsor_no source to use Avfall Sør collection JSON API

Wuerger:
### Summary
This PR updates avfallsor_no.py to use Avfall Sør’s JSON pickup calendar API instead of scraping HTML from the address page.

### What changed
- Replaced HTML scraping logic with direct API use
- Looked up the address via `https://avfallsor.no/wp-json/addresses/v1/address?lookup_term=...`
- Extracted `propertyId` from the returned `href`
- Fetched collection data from `https://avfallsor.no/wp-json/pickup-calendar/v1/collections/property-id/{property_id}`
- Converted JSON collection items into `Collection` objects
- Removed obsolete BeautifulSoup/date parsing code

### Why
The Avfall Sør site renders pickup details dynamically, making HTML scraping fragile and unreliable. Using the official JSON endpoint is much more robust and matches the working approach used by the referenced external project.

### Credit
This implementation was inspired by the workaround used in the `oal/avfallsor-mqtt-py` project, which also uses the same Avfall Sør JSON API flow.

### Testing
- Verified locally that the branch builds and the source produces collection entries
- Branch pushed to `Wuerger/hacs_waste_collection_schedule` as `fix/avfallsor_no-api-source`

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/avfallsor_no.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/avfallsor_no.py
@@ -1,8 +1,7 @@
 import datetime
-import logging
+import re
 
 import requests
-from bs4 import BeautifulSoup, Tag
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
 
@@ -10,7 +9,6 @@ TITLE = "Avfall Sør, Kristiansand"
 DESCRIPTION = "Source for Avfall Sør, Kristiansand."
 URL = "https://avfallsor.no/"
 TEST_CASES = {"Auglandslia 1, Kristiansand": {"address": "Auglandslia 1, Kristiansand"}}
-_LOGGER = logging.getLogger(__name__)
 
 ICON_MAP = {
     "Restavfall": "mdi:trash-can",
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/avfallsor_no.py
```diff
@@ -1,79 +1,39 @@
...
 TEST_CASES = {"Auglandslia 1, Kristiansand": {"address": "Auglandslia 1, Kristiansand"}}
-_LOGGER = logging.getLogger(__name__)
 
 ICON_MAP = {
     "Restavfall": "mdi:trash-can",
     "Bioavfall": "mdi:leaf",
     "Papp og papir": "mdi:package-variant",
     "Plastemballasje": "mdi:recycle",
     "Glass- og metallemballasje": "mdi:bottle-soda",
+    "residual": "mdi:trash-can",
```
markvp: is there a reason for keeping both sets of naming?

markvp (changes requested):

custom_components/waste_collection_schedule/waste_collection_schedule/source/avfallsor_no.py
Wuerger: No not really :-) your comment made me realize I should use something other than the icons to deduce waste type. My latest commit switches to another, more stable waste type identifier and is free of redundant code I believe.

markvp (approved):

PR #6070, opened 2026-04-19
Add support for vanersborg (token auth)

zer0coo:
Adds support for vanersborg which requires:

- Updated API endpoint (/api/nova/v1/)
- Bearer token authentication
- X-App-Identifier header

Also skips device/address registration since vanersborg uses an existing device ID from the mobile app.

Tested and verified working.


markvp (changes requested):
Thanks for adding Vänerborg support and for the general code quality improvements — `_safe_json`, `_fetch_list`, better exception handling, and cleaner variable names are all genuine improvements.

The main concern is the pattern of checking the provider name to branch logic — `_is_vanersborg()`, and several `if self._is_vanersborg():` blocks in `_headers()`, `_register_device()`, `_register_address()`, and `fetch()`. This is an "is this specific provider, then do this" approach. It works for one provider, but it doesn't scale: if a second token-auth provider or a second registration-free provider appears, you'd need to add another method and another set of checks.

The more durable pattern is to express provider *capabilities* as configuration in `SERVICE_PROVIDERS`, and then branch on those capabilities. For example:

```python
SERVICE_PROVIDERS = {
    "soderkoping": {
        ...
        "supports_registration": True,
        "requires_token": False,
    },
    "vanersborg": {
        ...
        "supports_registration": False,
        "requires_token": True,
        "app_version": "3.0.0.0",   # if X-App-Version varies by provider
    },
}
```

Then in `__init__`, read those flags once:
```python
cfg = SERVICE_PROVIDERS[self._service_provider]
self._requires_token = cfg.get("requires_token", False)
self._supports_registration = cfg.get("supports_registration", True)
self._app_version = cfg.get("app_version")
```

And everywhere you currently have `_is_vanersborg()`, use the flag instead:

```python
def _headers(self) -> dict[str, str]:
    headers = {"Accept": "application/json"}
    if self._app_version:
        headers["X-App-Version"] = self._app_version
    if self._api_key:
        headers["X-App-Identifier"] = self._api_key
    if self._requires_token and self._token:
        headers["Authorization"] = f"Bearer {self._token}"
    return headers

def _register_device(self):
    if not self._supports_registration:
        raise SourceArgumentRequired(
            "api_key",
            "This provider requires a pre-existing device ID. "
            "Find it in the mobile app under 'Om appen'.",
        )
    ...
```

This way the `_is_vanersborg()` method and all name-checks disappear entirely, and adding a third token-auth provider in the future is just a config change in `SERVICE_PROVIDERS` — no code logic to touch.

A couple of smaller things:

- **`HOW_TO_GET_ARGUMENTS_DESCRIPTION`**: The Vänerborg-specific instruction (`"For Vänerborg, use the Enhets-ID…"`) shouldn't be mixed into the shared description. Each provider entry in `EXTRA_INFO` could carry its own `howto` key, or the general description should stay generic.
- **`token` parameter**: It's undocumented — it should appear in `PARAM_DESCRIPTIONS` / `PARAM_TRANSLATIONS`.
- **`__init__` validation**: If `requires_token` is True but no `token` is provided alongside `api_key`, the error should be raised at init time, not deferred to `fetch()`.

zer0coo:
Thanks, that makes sense.

I refactored the implementation to use provider capability flags in SERVICE_PROVIDERS instead of checking the provider name. I also:

- removed _is_vanersborg() and related conditionals
- added requires_token and supports_registration flags
- documented the token parameter
- moved token validation to __init__
- restored a generic shared how-to description

Tested locally and confirmed working for Vänerborg.

Let me know if you'd like any further adjustments.

markvp:
Thanks for the updated implementation — the capability flags approach is exactly right. I've pushed a few finishing touches:

- Tightened the `__init__` validation so token-required providers raise immediately if `api_key` or `token` is missing, rather than failing later in `fetch()`
- Updated `doc/source/avfallsapp_se.md` to move Vänerborg into the supported section and document the `token` parameter with a config example
- Ran `update_docu_links.py` to add Vänerborg to `sources.json`, `README.md`, and translations

markvp (approved):

markvp (approved):

PR #6094, opened 2026-04-20
Refactor waste collection data handling

davewins:
API Response has changed and no longer returns a status field


davewins (commented):
Updated to improve exception message handling

PR #6099, opened 2026-04-21
arun.gov.uk: Change ICON_MAP order

whi-tw:
`mdi:trash-can` was applying to too many things. Moved 'specific' collection types to the start of the list, allowing 'waste' to act as a fallback.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/arun_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/arun_gov_uk.py
@@ -29,12 +29,12 @@ PARAM_DESCRIPTIONS = {
     }
 }
 ICON_MAP = {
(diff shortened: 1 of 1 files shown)


markvp (approved):
Nice catch — `_resolve_icon` iterates in insertion order and returns on first match, so having `"waste"` at the top meant `"Garden Waste"` and `"Food Waste"` were silently getting `mdi:trash-can`. Moving the specific keywords first and letting the generic ones act as fallbacks is exactly the right fix.

whi-tw:
That's why I made the change I made! Thanks for the kind words, Claude!

markvp:
> That's why I made the change I made! Thanks for the kind words, Claude!

Sorry. Trying to make a dent on 1.2k open issues means some of the messages come out a bit too Claude-y 🤣

PR #6134, opened 2026-04-25
fix(potsdam_de): exception chain order, add type 7, JSON safety

johannesvedder:
## Summary

Fixes two critical bugs and several minor issues in `potsdam_de.py` to match actual Potsdam website behavior.

### 🔴 Fix 1: Exception chain order (Critical)

**Problem:** `__date_is_collection_1_to_4` checked `FROM` dates before `TO` dates in holiday exception processing. When a date is both a `FROM` (its own collection was shifted away) and a `TO` (receiving another day's shifted collection), the method returned `False` immediately on the `FROM` match, never reaching the `TO` check.

**Fix:** Single-loop approach with `ausnahme_from` flag, matching the website's `index.js` logic:

```python
if check_exception:
    ausnahme_from = False
    for e_from, e_to in exceptions.items():
        if day == e_to:           # check TO first, return immediately
            return self.__date_is_collection_1_to_4(...)
        if day == e_from:         # flag FROM, keep looping
            ausnahme_from = True
    if ausnahme_from:             # after loop, skip original date
        return False
```

### 🔴 Fix 2: Type 7 (Grünabfallsammlung) completely missing (Critical)

**Problem:** `page-data.json` contains 1813 type 7 entries across Potsdam streets. The website (`index.js`, `daymatch7`) renders them. The Python integration had no `ICON_MAP[7]`, no `TYPE_MAP[7]`, no check method — all Grünabfallsammlung events silently dropped.

**Fix:** Added type 7 support matching the website's `daymatch7` logic (checks `termin1` against date):

```python
# Added to maps
ICON_MAP[7] = "mdi:leaf"
TYPE_MAP[7] = "Grünabfallsammlung"

# New method
def __date_is_collection7(self, node, day: datetime) -> bool:
    if node["typ"] != 7:
        return False
    termin1 = datetime.fromisoformat(node["termin1"].replace("Z", "+00:00")).date()
    return termin1 == day.date()
```

Disabled by `__date_is_collection()` to include it.

### 🟡 Additional Fixes

- **JSON error handling:** Wrap `json.loads(entry["ausnahmen"])` in try/except, fall back to `{}` (prevents crash from malformed data)
- **Typo fix:** `"culd not"` → `"Could not"` in error messages
- **Date range:** Changed `datetime(year+1, 1, 1)` to `datetime(year, 12, 31)` (unnecessary extra day)

## Testing

- Verified against live Potsdam data (`page-data.json`).
- Exception days now correctly shows as a collection day
- Type 7 entries pass through the new check method correctly

## Files Changed

- `custom_components/waste_collection_schedule/waste_collection_schedule/source/potsdam_de.py`

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/potsdam_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/potsdam_de.py
@@ -39,6 +39,7 @@ ICON_MAP = {
     4: "mdi:package-variant",
     5: "mdi:shower-head",
     6: "mdi:pine-tree",
+    7: "mdi:leaf",
 }
 
 TYPE_MAP = {
@@ -48,6 +49,7 @@ TYPE_MAP = {
     4: "Altpapier",
     5: "Biotonnenreinigung",
     6: "Weihnachtsbaumabholung",
+    7: "Grünabfallsammlung",
 }
 
 RHYTHM_MAP = {
@@ -109,13 +111,13 @@ class Source:
             r.raise_for_status()
         except Exception:
             if year == today.year:
-                raise Exception("culd not get data from url exit")
+                raise Exception("Could not get data from URL")
             return None
 
         try:
             data = r.json()["result"]["data"]["dbapi"]
         except Exception as e:
-            raise Exception("culd not decode server response") from e
+            raise Exception("Could not decode server response") from e
 
         for ortsteil in data["allRegionen"][0]["regionen"][0]["ortsteile"]:
             if ortsteil["name"].lower().strip() != self._ortsteil.lower().strip():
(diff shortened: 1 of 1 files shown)


markvp (approved):
Verified locally:
- Golm Akazienweg test case goes from 79 → 85 entries; the 6 extra are the previously-missed type 7 (Grünabfallsammlung) collections.
- Exception-ordering fix matches the site's `index.js` behaviour for dates that are both `e_from` and `e_to`.
- Date-range change from `(year+1, 1, 1)` to `(year, 12, 31)` is safe — when today is December the loop iterates the next year separately, so nothing is missed.
- black/isort clean; `update_docu_links.py` produces no diff.

Note: the other two TEST_CASES (Teltower Vorstadt, Uetz) return 0 entries on master too — pre-existing issue, out of scope.

LGTM.

PR #6149, opened 2026-04-26
fix Enfield address matching

armaneshaghi:
## Summary

Fix Enfield Council address matching so a candidate address only matches as a whole normalized phrase, instead of matching arbitrary substrings.

## Root Cause

The Enfield source accepted substring matches between normalized address strings. This allowed addresses such as house number `1` to match an input for house number `51`, because `1 ...` was found inside `51 ...`. The lookup then became ambiguous and no collection schedule was returned for otherwise valid addresses.

## Impact

Address-based configuration for Enfield Council is less likely to fail with false ambiguous matches. UPRN-based configuration is unchanged.

## Validation

- `git pull --ff-only origin master`
- `python test_sources.py -s enfield_gov_uk -i -l`
- `pytest tests/test_source_components.py`
- `python3 -m py_compile custom_components/waste_collection_schedule/waste_collection_schedule/source/enfield_gov_uk.py tests/test_source_components.py`
- `git diff --check`

Only the source file and a focused synthetic regression test are included; generated docs, README, and info files are left for CI as requested by the contribution guide.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/enfield_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/enfield_gov_uk.py
@@ -223,16 +223,28 @@ class Source:
         return (
             candidate_full == normalized_input
             or candidate_key == normalized_input
-            or candidate_key in normalized_input
-            or normalized_input in candidate_full
+            or cls._contains_normalized_phrase(normalized_input, candidate_key)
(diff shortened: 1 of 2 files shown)


markvp (approved):
Great fix — the word-boundary regex approach correctly prevents house number `1` from matching `51`, `14`, etc. The regression test in `test_source_components.py` is a nice addition that will guard against this class of bug going forward.

No changes needed from our side — code is clean and all tests pass.

PR #6164, opened 2026-04-28
Fix vafab miljo when missing bintype info

gjohansson-ST:
They have inserted some "dummy" details in the json response for our area (at least) which does not contain all the bin type details resulting in string concatenation issues.
PR simply replace a `None` response with an empty string.


markvp:
Thanks for this fix — the defensive `or ""` guard on the `BinType` fields is exactly the right approach for handling malformed API responses gracefully.

Minor update applied: reformatted a few split lines to satisfy `black`.

markvp (approved):
Thanks for this fix — the defensive `or ""` guard on the `BinType` fields is exactly the right approach for handling malformed API responses gracefully.

Minor update applied: reformatted a few split lines to satisfy `black`.

PR #6167, opened 2026-04-28
Fix kirklees_gov_uk: rewrite for new my.kirklees.gov.uk API

### LINKED ISSUE #2988: [Bug]: Kirklees completely new website?
### I Have A Problem With:

A specific source

### What's Your Problem

The bin lookup UI (and their implementation) seems completely different now and I can't see a link to a calendar (which is what the current implementation looks for). The UI just now shows the next 2 collection dates, that's all you get.

Is it worth trying to support this new data source with just 2 calendar days? I could look in to this if you believe it's worth it

If not, should the source be removed as it seems totally incompatible?

### Source (if relevant)

kirklees_gov_uk

### Logs

_No response_

### Relevant Configuration

_No response_

### Checklist Source Error

### Checklist Sensor Error

### Required

elyobelyob:
Kirklees migrated from the old beta/HTML site to `my.kirklees.gov.uk`'s JSON form-builder API. The current source has been broken since then (see #2988).

Rewrites the source to use the new API end-to-end:
1. `domain/my.kirklees.gov.uk` → cookies
2. `authapi/isauthenticated` → auth-session SID
3. `apibroker/runLookup` → postcode lookup, property type, `validatedUPRN`, then collection dates.

## ⚠️ Breaking change for `kirklees_gov_uk` users

| | Before | After |
|---|---|---|
| Required | `door_num + postcode` | `uprn + postcode` |
| Optional | `uprn` | — |

The new API is keyed on UPRN, so `door_num` is no longer accepted. Users can find their UPRN at <https://www.findmyaddress.co.uk/>. Doc page updated.

## Tested live (2026-04-28)

- `HD9 7HA` (UPRN `83074265`) → Grey + Green wheelie bin dates ✅
- `HD8 8NA` (UPRN `83194785`) → Grey + Green wheelie bin dates ✅

## New optional arg: `predict`

Adds `predict: bool = False`. When `True`, the source extrapolates a year of fortnightly dates from the next known date. Default `False` — no behaviour change for HA users (who get nightly refreshes anyway). Useful for consumers exporting to static `.ics` files.

Closes #2988

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/kirklees_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/kirklees_gov_uk.py
@@ -1,139 +1,242 @@
-import logging
-import re
-from datetime import datetime
+from datetime import date, datetime, timedelta
+from time import time_ns
 from typing import Any
 
 import requests
-from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 
-_LOGGER = logging.getLogger(__name__)
 
 TITLE = "Kirklees Council"
-DESCRIPTION = "Source for waste collections for Kirklees Council"
+DESCRIPTION = "Source for waste collections for Kirklees Council (my.kirklees.gov.uk)"
 URL = "https://www.kirklees.gov.uk"
 TEST_CASES = {
-    "Test_001": {"door_num": 20, "postcode": "HD9 6LW"},
-    "test_002": {"door_num": "6", "postcode": "hd9 1js"},
-    "HD8 8NA, 1": {"door_num": "1", "postcode": "HD8 8NA", "uprn": "83194785"},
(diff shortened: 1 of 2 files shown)


Cadair:
Thanks so much for fixing this, been thinking about doing it myself for ages.

I've downloaded it and tested it works as a drop-in replacement with my existing config (after the uprn change).

markvp (approved):
Excellent work rewriting the source for the new `my.kirklees.gov.uk` JSON API — the step-by-step session flow is well-structured and clearly documented. The `predict` option is a thoughtful addition for non-HA consumers, and we already have a user confirming it works as a drop-in replacement.

Minor updates applied: swapped the UPRN-not-found `ValueError` for `SourceArgumentNotFoundWithSuggestions` (which surfaces the valid UPRNs at that postcode to the user), corrected the `fetch()` return type annotation to `list[Collection]`, and applied `black` formatting.

PR #6168, opened 2026-04-29
Add waste collection schedule support and GUI for Parramatta

PiexlPuck:


markvp (approved):
Thanks for adding Parramatta — the ArcGIS geocode + feature layer approach works well for polygon-based zone lookups, and the fortnightly area parity calculation is a neat solution.

Minor updates applied: replaced bare `Exception` raises with `SourceArgumentNotFound`, added `timeout=30` to both request calls, reverted the unrelated `config_flow.py` change (auto-creating sensors at setup is a separate discussion worth its own PR), reverted the generated files that CI handles post-merge, and applied `black` formatting.

PR #6170, opened 2026-04-29
Accept localities within Varberg/Falkenberg municipalities

nnajko:
## Summary

VIVAB serves the entire Varberg and Falkenberg municipalities, but the source previously only accepted addresses containing the substrings `varberg` or `falkenberg`. Residents in smaller localities within those municipalities (Veddige, Tvååker, Bua, Träslövsläge, Ullared, Vessigebro, Slöinge, etc.) could not configure the integration even though their addresses are served by the same regional API.

## Changes

- Add a `MUNICIPALITY_LOCALITIES` map listing the known tätorter for each municipality.
- Replace the two-branch substring check in `Source.__init__` with a word-boundary regex match against that list, so e.g. `"Storgatan 1, Veddige"` correctly resolves to the Varberg API.
- Update the user-facing error message and the docs (`doc/source/vivab_se.md`) to reflect the broader set of supported localities.

Existing `, Varberg` / `, Falkenberg` configs (and the `building_id` flow) continue to work unchanged.

## Test plan

- [x] Existing `TEST_CASES` (Varberg & Falkenberg) still pass via `test_sources.py -s vivab_se`.
- [x] Logic-level verification: addresses ending in Veddige, Tvååker, Bua, Ullared, Vessigebro now resolve to the correct regional API; unrelated cities (e.g. Stockholm) are still rejected.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/vivab_se.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/vivab_se.py
@@ -35,6 +35,20 @@ API_URLS = {
     "varberg": "https://minasidor.vivab.info/FutureWebVarberg/SimpleWastePickup/",
 }
 
+# Localities (tätorter) within each municipality served by VIVAB.
+# Used to resolve a street_address to the correct regional API.
+MUNICIPALITY_LOCALITIES = {
+    "varberg": [
+        "varberg", "bua", "kungsäter", "rolfstorp", "skällinge",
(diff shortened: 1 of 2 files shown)


markvp (approved):
Great fix — extending the locality matching to cover all tätorter within Varberg and Falkenberg municipalities is exactly the right approach, and the word-boundary regex prevents false matches on substrings. The updated error message is also much more helpful.

Minor update applied: reformatted the locality lists and one long line to satisfy `black`.

PR #6176, opened 2026-04-29
Update yarra_ranges_vic_gov_au.py

bdog720:
updated rubbish collection icon to a more appropriate icon


markvp:
Thanks for spotting the incorrect icon — `mdi:trash-can` is definitely the right choice over `mdi:sofa`! It looks like this fix made it into `master` via another route in the meantime, so your PR is now showing an empty diff. Safe to close this one — the change you intended is already in.

PR #6177, opened 2026-04-30
Add Baton Rouge, LA, US waste collection source from ArcGIS

KlfJoat:
* Adds a new source batonrouge_la_us for Baton Rouge, Louisiana using the city/parish's ArcGIS FeatureServer
* Geocodes the address and performs a spatial query on each of the garbage, trash, and recycling polygon feature layers
* Returns twice-weekly garbage collection dates, once-weekly trash collection date, and once-weekly recycling collection date, depending on zone
* Uses the reusable ArcGis service module


markvp (approved):
Great source — the ArcGIS geocode + polygon feature layer approach is clean, and returning 26 weeks of fortnightly dates is a sensible window. Good use of the existing `ArcGis` service module too.

Minor updates applied: renamed the doc file from `batonrouge_la_us.md` to `baton_rouge_la_us.md` to match the source module name (required for `update_docu_links.py` to link the doc correctly), and applied `black` formatting.

KlfJoat:
Thanks. I shamelessly copied from other US ArcGIS source files, like Hoover, AL, and Rochester, NY. IIRC, one of them also had multi-layer and the other had twice weekly. 

Can't believe I messed up the filename. 😅

PR #6188, opened 2026-04-30
Add MRC Joliette (QC) waste collection source

lit-af:
For https://github.com/mampfes/hacs_waste_collection_schedule/issues/6163


markvp (approved):
Thanks for adding MRC Joliette — clean implementation with good use of `CONFIG_FLOW_TYPES` and the icon/type maps. I made a few small fixes on your branch: corrected the `SourceArgumentNotFoundWithSuggestions` call (takes 3 args, not 4 — would have crashed at runtime), removed an unreachable validity check in `fetch()`, fixed a double-slash in the API URL, and switched to explicit UTC timestamps. Also added the missing `doc/source/mrcjoliette_qc_ca.md`. Merging now.

PR #6196, opened 2026-05-01
Add Ekonovus (ekonovus_lt) source for Lithuania

sanis:
Adds support for waste collection schedules from UAB "Ekonovus" (https://www.ekonovus.lt), serving residential addresses in Lithuania.

The schedule is sourced from the public Power BI report Ekonovus publishes on their website. The source takes a single 'address' argument matching the address shown in the public report.


markvp (approved):
Nice work — clean Power BI integration with good structure. Made a couple of small tweaks: removed the `lt` language key from `PARAM_DESCRIPTIONS` (only `en`/`de`/`it`/`fr` are supported), and applied black formatting. All tests pass. Approving!

PR #6198, opened 2026-05-01
Fix grafikai_svara_lt: rewrite for TanStack server-fn API

sanis:
## Summary                                             

  In late April 2026 the upstream Kauno švara website (grafikai.svara.lt) was rebuilt as a TanStack Start (TSR) React app. The previous REST endpoints (`/api/contracts`, `/api/schedule`) now return 404, so the existing source fails  
  for everyone with `JSONDecodeError: Expecting value` (it tries to JSON-parse a 404 HTML page). Cached events keep showing for a while after the change but the calendar effectively stops updating.
                                                                                                                                                                                                                                         
  The new architecture routes all data fetching through a single TanStack server function:

  ```
  GET /_serverFn/<sha256>?payload=<seroval-encoded-json>
  x-tsr-serverFn: true                                                                                                                                                                                                                   
  ```
                                                                                                                                                                                                                                         
  This PR rewrites the source against that new contract. 

  ## What changed

  - **Discover the data-fetching server-fn id fully dynamically.** The 64-char sha256 is baked into the JS bundle at build time and changes on every upstream rebuild — so there is no hardcoded id at all. The source scrapes `/`, finds
   every referenced JS bundle, extracts every quoted 64-char hex literal, and probes each candidate against the cheap `/schedule/getdistricts` endpoint to pick the one with the data-fetcher shape.
  - **Per-instance cache with auto-invalidation.** The discovered id is cached on the `Source` instance. When upstream redeploys and the cached id stops working (any non-200 / wrong-shape response), the source invalidates the cache  
  and re-discovers exactly once before failing. Net effect: an upstream rebuild causes one transient failure on a single polling cycle, then automatic recovery — no code change needed.                                                 
  - **Hand-rolled tiny Seroval encoder/decoder in pure stdlib.** No new dependency. Only the small subset of node types these endpoints actually return is handled (`t=0` number, `t=1` string, `t=2` bigint, `t=9` array, `t=10/11`
  object).                                                                                                                                                                                                                               
  - **Public input arguments unchanged** (`region`, `street`, `house_number`, `district`, `waste_object_ids`), so existing user configs keep working without edits.
  - **Doc updates.** Dropped the stale `waste_object_ids` example values (the numeric IDs changed during the rebuild). Replaced the obsolete "copy URL from Atsisiųsti button" discovery trick with a DevTools-based recipe, and noted   
  that pre-rebuild IDs are no longer valid.                                                                                                                                                                                              
  - **TEST_CASES kept** — both upstream addresses (Demokratų g. 7, Alytaus g. 2) verified to return non-empty results against the new API; just removed the now-stale hardcoded `waste_object_ids` from the Demokratų entry.             
                                                                                                                                                                                                                                         
  ## Test plan                                           
                                                                                                                                                                                                                                         
  - [x] Verified end-to-end against three addresses including both upstream `TEST_CASES`. Demokratų g. 7 returns 7 contracts, Alytaus g. 2 returns 3 contracts, my own address returns 38 schedule entries across 6 months.
  - [x] Verified the dynamic discovery with the hardcoded id removed: cold start picks the right server-fn from the live JS bundle.                                                                                                      
  - [x] Verified cache reuse: subsequent fetches don't redo discovery.                                                                                                                                                                   
  - [x] Verified stale-id recovery: injecting a fake fn id and calling `fetch()` triggers re-discovery and returns correct data.                                                                                                         
  - [x] Verified the failure case: when discovery legitimately can't find a working id, the source raises a clear, descriptive error rather than silently using a broken id.                                                             
  - [x] Running on my own HA instance — `calendar.kauno_svara` is back to `on` with today's "Mišrios komunalinės" event.                                                                                                                 
                                                                                                                                                                                                                                         
  ## Caveats                                                                                                                                                                                                                             
                                                         
  The dynamic discovery scrapes a JS bundle, which is fragile by nature — TanStack internals can change the way server-fn ids are emitted. The current heuristic is "any 64-char hex literal in any `/assets/*.js` referenced from `/`". 
  If TSR ever stops emitting plain string literals (e.g. moves to import-maps or splits ids across template strings), the scrape returns no candidates and the source raises a loud, actionable error. That seems strictly better than   
  silently freezing on a stale id.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/grafikai_svara_lt.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/grafikai_svara_lt.py
@@ -58,20 +58,24 @@ def _seroval_encode_request(api_path: str, tenant_id: int = _TENANT_ID) -> str:
     return json.dumps(
         {
             "t": {
-                "t": 10, "i": 0,
+                "t": 10,
+                "i": 0,
                 "p": {
                     "k": ["data"],
-                    "v": [{
-                        "t": 10, "i": 1,
-                        "p": {
-                            "k": ["apiPath", "tenantId"],
-                            "v": [
-                                {"t": 1, "s": api_path},
-                                {"t": 2, "s": tenant_id},
-                            ],
-                        },
-                        "o": 0,
-                    }],
+                    "v": [
+                        {
+                            "t": 10,
+                            "i": 1,
+                            "p": {
+                                "k": ["apiPath", "tenantId"],
+                                "v": [
+                                    {"t": 1, "s": api_path},
+                                    {"t": 2, "s": tenant_id},
+                                ],
+                            },
+                            "o": 0,
+                        }
+                    ],
                 },
                 "o": 0,
             },
@@ -148,7 +152,7 @@ def _candidate_fn_ids(session: requests.Session) -> list[str]:
         html = session.get(_BASE_URL + "/", timeout=15).text
     except requests.RequestException:
         return []
-    bundles = re.findall(r'/assets/[A-Za-z0-9._-]+\.js', html)
+    bundles = re.findall(r"/assets/[A-Za-z0-9._-]+\.js", html)
     seen: set[str] = set()
     ordered: list[str] = []
     for bundle in bundles:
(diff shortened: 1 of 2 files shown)


markvp (approved):
Nice work handling the TanStack rewrite — dynamic server-function ID discovery is the right call here. Made a small formatting pass with black/isort. Approving!

Issue #3019, opened 2024-11-12
[Source Request]: Sherbrooke, Québec, Canada


mlevac21:
### Municipality / Region

Sherbrooke/Québec/Canada

### Collection Calendar Webpage

https://www.sherbrooke.ca/fr/services-a-la-population/collecte-des-matieres-residuelles/calendrier-des-collectes

### Example Address

667 Patricia

### Collection Data Format

As ICS/ICAL file

5ila5:
Do you know if the ICS links stay valid forever?
e.g.:
https://www.sherbrooke.ca/Fichiers/3337a882-4a53-e611-80ea-00155d09650f/Sites/333ceda8-915d-e611-80ea-00155d09650f/Documents/Collecte-matieres-residuelles/Calendriers-ics/2024-calendrier-01-mardi.ics


It looks like it's pretty easy to write a dedicated source if they don't

markvp:
The Sherbrooke ICS links appear to be yearly static files (e.g. `2024-calendrier-01-mardi.ics`). These can be used with the existing ICS source, but you'd need to update the URL each year. If anyone from Sherbrooke can share the current 2026 ICS URL for their zone, we can document it.

mlevac21:
https://api.sherbrooke.ca/collection-schedules/calendars/1/sectors/01/days/Mardi/ics I think it’s not yearly static anymore!

markvp:
Great find! That looks like a proper REST API. Unfortunately the endpoint returns 403 from outside Canada, so we can't verify it directly. Could you confirm:

1. Does that URL return valid ICS data when you open it in your browser?
2. What do the URL segments mean — `calendars/1` = calendar type, `sectors/01` = your zone, `days/Mardi` = your collection day? Are there other valid values?
3. How did you find this URL — is there a page where you select your address and it generates the link?

If this works, we can add it as an ICS source with the URL pattern documented.

lit-af:
1. Here's the content of the file when clicking on the link. Added .txt to comply with github upload rules.
[2026-04-29_calendrier_01_Mardi.ics.txt](https://github.com/user-attachments/files/27216527/2026-04-29_calendrier_01_Mardi.ics.txt)

markvp:
Thanks for the ICS confirmation! Just need a couple more details to get this working:

[...]

PR #6221, opened 2026-05-02
Add Sherbrooke (QC)

lit-af:
- Add Sherbrooke, QC, CA as source
- Fix for Laval, QC, CA. Added missing default_param regex.


lit-af:
Close #3019

markvp (approved):
Thanks for adding Sherbrooke and fixing the Laval regex default! Both changes look correct — moving `regex` into `default_params` for Laval is exactly the right approach, and the Sherbrooke YAML follows the established pattern nicely. Appreciated!

PR #6222, opened 2026-05-02
Add AHP to waste collection mappings for southglos_gov_uk.py

ZuluWhiskey:
South Gloucestershire council now have AHP (Absorbent Hygiene Product) collections.
The introduction of AHP broke the integration for me as AHP wasn't mapped.
I've tested this fix locally and it has resolved my issue.


markvp (approved):
Thanks for this — AHP collections are a new one! The fix is exactly right: both maps updated consistently with the existing pattern, and `mdi:diaper-outline` is a sensible icon choice. Appreciated!

ZuluWhiskey:
> Thanks for this — AHP collections are a new one! The fix is exactly right: both maps updated consistently with the existing pattern, and `mdi:diaper-outline` is a sensible icon choice. Appreciated!

Yeah, AHP wasn't something I'd heard about before -  I had to Google it!

Thanks for the great feedback 😃

Issue #4298, opened 2025-06-23
[Source Request]:


sacredland:
### Municipality / Region

Saguenay - Chicoutimi, Québec 

### Collection Calendar Webpage

https://ville.saguenay.ca/services-aux-citoyens/environnement/recyclage/horaire-de-la-collecte

### Example Address

Need the neightborhood, Civic #, name of the street

### Collection Data Format

As html on a webpage

### Additional Information

Aldo in pdf

lit-af:
Technically possible to extract the data from HTML. 

API endpoint found for the PDF generation: https://ville.saguenay.ca/collecte_calendrier?batiment=8773&annee_suivante=0&adresse=514,%20Rue%20Malraux,%20Chicoutimi

sacredland:
Intergration exist: https://github.com/DavidBilodeau1/saguenay_collection

But need to be fixed https://github.com/DavidBilodeau1/saguenay_collection/issues/23

markvp:
This was implemented in PR #6224 and is available now via the `saguenay_ca` source. You'll need your building ID (`batiment`), which you can find in the URL of the collection calendar page (e.g. `batiment=8773`):

```yaml
waste_collection_schedule:
  sources:
    - name: saguenay_ca
      args:
        batiment: 8773  # replace with your building ID
```

To find your building ID, visit https://ville.saguenay.ca/services-aux-citoyens/environnement/recyclage/horaire-de-la-collecte, enter your address, and note the `batiment` value in the resulting URL.

PR #6224, opened 2026-05-02
Add Ville de Saguenay source (HTML scraper, batiment param)

lit-af:
For #4298


markvp (approved):
Great addition — the HTML scraper is clean and the bilingual howto instructions are a nice touch. Made a few small tweaks: removed a shadowed typing import, added German and Italian translations, swapped the bare ValueError on invalid building ID for SourceArgumentNotFound, and extended the fetch to cover both the current and next year for better year-end coverage. Thanks for the contribution!

PR #6226, opened 2026-05-03
Update instructions on how to get the streetId for the NZ-Wellington City Council calendar

leslieb256:
Updates the instructions so people can find their streetId for Wellington City Council, New Zealand

The blue "next week" button in the original instructions no longer lets you copy the URL.
The iCalendar button does still provide a copyable URL and the URL has the streetId.

Instructions updated to copy the Calendar button link.


markvp (approved):
Thanks for catching this — the updated instructions using the iCalendar button are much clearer.

PR #6234, opened 2026-05-03
Add configurable fetch_interval_days to control fetch frequency

khauns:
### Motivation
- I had issues with ics source: doc/ics/awb_landkreis_karlsruhe_de.md. It stopped working after some time giving me a "403 Forbidden“ / " 403 Forbidden You don't have permission to access this ressource". message.
- Increasing the fetch interval to 30 days, solved it for me. Running smoothly for more than 6 months. 

### Summary
- Allow configuring the interval (in days) between data fetches so integrations can reduce polling frequency or preserve the existing daily behavior.
- Support this new option in both YAML and the UI options flow while retaining the default daily fetch behavior for backward compatibility.

### Description
- Add new constant `CONF_FETCH_INTERVAL_DAYS` and default `CONF_FETCH_INTERVAL_DAYS_DEFAULT = 1` to `const.py` and expose it in configuration and UI schemas.
- Wire the option into the UI: add `fetch_interval_days` to the options form in `config_flow.py` with validation, and pass the option into `WCSCoordinator` from `init_ui.py`.
- Add YAML support in `init_yaml.py` to accept `fetch_interval_days` in `configuration.yaml` and pass it to `WasteCollectionApi` constructor.
- Implement interval-aware fetching in two places: for YAML-managed configuration, `WasteCollectionApi` now stores `_fetch_interval_days` and `_last_fetch_date` and skips fetching when the interval has not elapsed; for config-flow/coordinator-managed configuration, `WCSCoordinator` uses the `update_interval` of `DataUpdateCoordinator` when `fetch_interval_days > 1`, while preserving existing timer-based daily fetch semantics when `fetch_interval_days == 1`.
- Update documentation (`doc/configuration.yaml` and `doc/installation.md`) to document the new `fetch_interval_days` option and default behavior.

### Testing
- No automated tests were added or executed as part of this change.


custom_components/waste_collection_schedule/config_flow.py
```diff
@@ -1016,6 +1019,12 @@ async def async_step_init(self, user_input: dict[str, Any] | None = None):
                         CONF_FETCH_TIME, CONF_FETCH_TIME_DEFAULT
                     ),
                 ): TimeSelector(),
+                vol.Optional(
+                    CONF_FETCH_INTERVAL_DAYS,
+                    default=self._entry.options.get(
```
markvp, line 1024: `fetch_interval_days` is missing from the translation files. Please add entries to the `options > step > init > data` and `data_description` sections of `translations/en.json` (alongside the existing `fetch_time` entries), and add matching entries to `de.json`, `fr.json`, and `it.json` (English fallback text is fine if you don't have translations for the other locales).

Note: only the `options:` section at the bottom of those files is hand-maintained — the source-parameter sections above are generated by `update_docu_links.py` and must not be edited.

custom_components/waste_collection_schedule/config_flow.py
```diff
@@ -1105,6 +1114,8 @@ async def async_step_init(self, user_input: dict[str, Any] | None = None):
                 cv.time(user_input[CONF_DAY_SWITCH_TIME])
             except vol.Invalid:
                 errors[CONF_DAY_SWITCH_TIME] = "time_format"
+            if user_input.get(CONF_FETCH_INTERVAL_DAYS, 1) < 1:
```
markvp: This check is dead code. The `vol.Range(min=1)` in the schema above already rejects values below 1 before this branch is ever reached. Please remove it.

custom_components/waste_collection_schedule/wcs_coordinator.py
```diff
@@ -60,17 +62,26 @@ def __init__(
...
+                hass,
+                self._fetch_callback,
+                self._fetch_time.hour,
+                self._fetch_time.minute,
+                self._fetch_time.second,
+            )
+
         # start timer for day-switch time
         if self._day_switch_time != self._fetch_time:
             async_track_time_change(  # TODO: cancel on unload
```
markvp, line 79: When `fetch_interval_days > 1`, the `_fetch_tracker` is never set, so `fetch_time` and `random_fetch_time_offset` are silently ignored — the fetch fires at an uncontrolled time from HA startup rather than at the user's configured hour. The YAML path behaves differently (fires at `fetch_time` daily, but skips the actual API call when the interval hasn't elapsed). Please document this difference in `doc/installation.md`, and consider whether the UI path should anchor its first fetch to `fetch_time` for consistency.

markvp (changes requested):
Thanks for this PR — reducing polling frequency is a real need for sources that rate-limit or return 403s. A few things to address before this can merge.

markvp (approved):
Thanks for the follow-up — all three points are addressed. I added the missing fetch_interval_days entries to it.json (English fallback for now, to be polished by an Italian speaker later); fr.json has no options: section on master so there's nothing to anchor to there. Everything else looks good. Approving.

PR #6235, opened 2026-05-03
Add source for Shawinigan (QC)

JordanM1738:


markvp (approved):
Thanks for this, @JordanM1738 — nice implementation using the shared ArcGIS service. Made a couple of small tweaks: reverted the generated `sources.json` (the CI handles that post-merge), removed the standalone test file (it couldn't run due to the HA import layer), and dropped a leftover wrapper method that was only there for those tests. The source logic itself looks good — approved!

Issue #4479, opened 2025-07-24
[Source Request]: waste collection for St-Basile le grand, Canada


patgiguer:
### Municipality / Region

St-Basile le grand

### Collection Calendar Webpage

https://www.villesblg.ca/wp-content/uploads/2024/12/Calendrier_collectes-et-depots_2025.pdf

### Example Address

77 rue Des Chenes St- basile le grand QC.CA j3n 1g3

### Collection Data Format

As PDF Download

### Additional Information

https://www.villesblg.ca/calendrier/

lit-af:
@markvp this issue can be closed.

PR #6246, opened 2026-05-04
Add Ville de Saint-Basile-Le-Grand (QC) Canada

lit-af:
Source for #4479


markvp (approved):
Great addition — Saint-Basile-le-Grand is a nice clean RSS-based source. Made a couple of small tweaks before approving:

- **Date format fix:** the feed uses `DD/MM/YYYY` (e.g. `27/11/2026`), not `MM/DD/YYYY`, so any collection day > 12 was being silently skipped. Fixed the unpacking order to `day, month, year`.
- Removed a shadowed `from typing import Collection` import (it was immediately overridden by the WCS import on the next line).
- Trimmed some inline comments that the code already made obvious.
- Removed a duplicate example block from the doc file.

PR #6260, opened 2026-05-06
Fix Stockport UK Parsing Issues

0xTim:
Fixes some parsing issues with Stockport that would cause incorrect or unavailable dates to be shown in Home Assistant.

I've also added some tests but the pytest.ini file currently only runs the test_source_components.py file and not any of the other tests. Happy to add the fix for this into this PR as well, or in a separate PR or leave it out


markvp (approved):
Great fix — replacing the string-search approach with proper BeautifulSoup traversal is clearly the right call, and the unit tests do a nice job documenting the regression. Made a couple of small tweaks: removed an unreachable normalisation branch, switched the logger calls to %-style, and updated the test import path to match the project convention. Thanks for the contribution!

PR #6273, opened 2026-05-07
Add source: Mulhouse Alsace Agglomeration (mulhouse_alsace_fr)

damienwolfer67:
Covers the 39 m2A municipalities by querying the official open data portal (data.mulhouse-alsace.fr). Collection dates are projected from the published frequency/weekday rules, then adjusted with the holiday reschedules and cancellations parsed from the dataset ferie field.


markvp (approved):
Nice piece of work. Tested against the live open-data portal — all three TEST_CASES return sensible schedules (Wittelsheim 65, Mulhouse/Centre Ville 150, Habsheim 39 entries), the regex-based `ferie` parsing correctly handles the "reporté au …" reschedules and the "ne sont pas maintenues" cancellations against the actual data observed today, and lint/structural tests are all green. Doc page is included. Approving as-is.

PR #6304, opened 2026-05-11
feat: add Brossard, Québec ICS source

lit-af:


markvp (approved):
Thanks for the bilingual howto and the clear sector list — really nice touch. Pushed a trailing newline to the YAML file; otherwise this is good to merge.

PR #6305, opened 2026-05-11
feat: add MRC Matawinie (QC) source for gmrmatawinie.org

lit-af:
Same website structure as MRC Joliette (QC)


markvp (approved):
Thanks for the addition - tested all three sector test cases against the live API and everything works cleanly. Same pattern as your MRC Joliette source. Approving.

PR #6306, opened 2026-05-11
Add RMR Lac-Saint-Jean (QC) source with fr/en language support

lit-af:
Scrapes the RMR Lac-Saint-Jean AJAX calendar endpoint at calendrier.rmrlac.qc.ca/calendrier-de-collectes using address lookup (no geocoding required). Parses collection dates from data-dates-* attributes on the calendar div, supporting Garbage/Recycling/Organic collections with both French and English collection type names via the language parameter.


markvp (approved):
Thanks Alexandre, nicely done. Clean structure, live AJAX endpoint with proper bilingual type mapping, exception handling uses the right class, and the doc file is in place. Live-tested both fr and en TEST_CASES and got 35 collections each. Approving as-is.

PR #6307, opened 2026-05-11
feat: Add Source - Palm Coast, FL US (palmcoast_fl_gov)

ChrisMancini:
```bash
> python .\test_sources.py -s palmcoast_fl_gov
Testing source palmcoast_fl_gov ...
  found 104 entries for Ripcord Lane
```


markvp (approved):
Thanks Chris — nice clean source. Small note: the doc file ended with some stray AI tool-call output (`</content>` and `<parameter>` lines) — looks like an accidental paste. Stripped those and the file is good to go. Approving.

Issue #4566, opened 2025-08-08
[Source Request]: SEAB Biella


baxbax2504:
### Municipality / Region

Mongrando

### Collection Calendar Webpage

https://www.seab.biella.it/aree-servite/

### Example Address

Via Trucco 15, Mongrando

### Collection Data Format

As PDF Download

PR #6334, opened 2026-05-15
#4566

lucamot:
This PR address issue #4566 by adding support and parsing for PDFs of the waste scheduls for SEAB Biella.


markvp (approved):
Nice work — thanks for tackling #4566. Verified locally: the PDF parser pulls 123 collection entries from the Ailoche calendar, lint is clean, and the structural tests pass. `pypdf` is already a project dependency so no manifest change needed.

markvp:
Pushed a couple of small tweaks: removed an unused `logging` import and added `PARAM_TRANSLATIONS`.

PR #6348, opened 2026-05-17
Add Mid-Western Regional Council (NSW, Australia) waste collection source

rodgrech:
Adds support for Mid-Western Regional Council NSW, Australia.

Supported areas:
- Mudgee North Monday
- Mudgee North Thursday
- Mudgee South Tuesday
- Mudgee South Wednesday
- Gulgong Monday
- Gulgong Thursday
- Kandos/Rylstone Friday

Includes:
- Weekly /general waste
- Weekly FOGO
- Fortnightly recycling
- Fortnightly paper/cardboard

Tested locally with Home Assistant and TrashCard.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/midwestern_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/midwestern_nsw_gov_au.py
@@ -1,0 +1,59 @@
+from datetime import date, timedelta
+
+from waste_collection_schedule import Collection
+
+TITLE = "Mid-Western Regional Council"
+DESCRIPTION = "Waste collection source for Mid-Western Regional Council NSW"
+URL = "https://www.midwestern.nsw.gov.au/"
+
+TEST_CASES = {
+    "Mudgee North Monday": {"area": "mudgee_north_monday"},
+}
+
+ICON_MAP = {
+    "Landfill": "mdi:trash-can",
+    "FOGO": "mdi:leaf",
+    "Recycling": "mdi:recycle",
+    "Paper/Cardboard": "mdi:newspaper",
+}
+
+AREAS = {
+    "mudgee_north_monday": 0,
+    "mudgee_north_thursday": 3,
+    "mudgee_south_tuesday": 1,
+    "mudgee_south_wednesday": 2,
+    "gulgong_monday": 0,
+    "gulgong_thursday": 3,
+    "kandos_rylstone_friday": 4,
+}
+
+
+class Source:
+    def __init__(self, area):
+        if area not in AREAS:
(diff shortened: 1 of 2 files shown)


markvp (changes requested):
Thanks for the contribution Rodney, and for putting together a doc file — much appreciated.

Unfortunately I can't merge this in its current shape. The source generates the schedule locally from `weekday()` and `isocalendar().week % 2` rather than fetching from MWRC, and the in-code comment notes the fortnightly anchor hasn't been confirmed yet. Project policy is that every source must pull from a live provider endpoint (API, ICS feed, or scraped page) — hardcoded or synthesised schedules drift the moment the council shifts a week (public holidays, Christmas roll-over, etc.) and we have no way to keep them right.

A few directions worth investigating for MWRC:

- Their site runs on Granicus OpenCities — sometimes there's a JSON-backed widget on the waste page (open the calendar page in a browser, watch the Network tab for an XHR returning collection data).
- Check whether MWRC publishes an ICS calendar or PDF lookup, or if their "My Local Services" / similar mobile app has an API behind it.
- If they offer a downloadable street-level lookup, `c_trace_de`-style platforms or our ICS YAML loader (`doc/ics/yaml/*.yaml`) might fit.

A few smaller things to fix in the next revision regardless of which approach you take:

- `COUNTRY = "au"` is required (lowercase).
- Add `PARAM_TRANSLATIONS` and `PARAM_DESCRIPTIONS` dicts.
- Use `SourceArgumentNotFoundWithSuggestions` from `waste_collection_schedule.exceptions` instead of `ValueError`.
- Wire `ICON_MAP` into the `Collection(icon=...)` argument so the icons actually apply.
- Add a trailing newline to both new files.

Happy to take another look once it's fetching real data. If you hit a dead end finding a public endpoint, drop what you've found in a comment and I can take a look too.

rodgrech:
Thanks for the feedback and guidance.

I've reworked the source to use the live MWRC OpenCities calendar API instead of locally generated schedules.

Changes in the updated revision include:

* Fetching live collection data from the MWRC API endpoint
* Server-side handling of fortnightly recycle/paper alternation
* Added COUNTRY = "au"
* Added PARAM_TRANSLATIONS and PARAM_DESCRIPTIONS
* Replaced ValueError with SourceArgumentNotFoundWithSuggestions
* Added icon mapping via Collection(icon=...)
* Cleaned up formatting/newlines

The source now pulls authoritative collection data directly from MWRC and no longer relies on locally synthesised schedules or ISO week calculations.

Confirmed working areas currently include:

* Mudgee North Monday/Thursday
* Mudgee South Tuesday/Wednesday
* Gulgong Monday/Thursday
* Kandos/Rylstone Friday

Thanks again for pointing me toward the OpenCities widget/API approach — that ended up being the correct path.

PR #6356, opened 2026-05-17
Bugfix for South Tyneside UPRN lookup - 

CartesianJohn:
Removed zfill padding UPRN to 12 digits before lookup,
Added Test_004 for UPRN less than 12 digits.


markvp (approved):
Thanks for the fix and the extra test case. Clean, minimal change — happy to merge.

markvp:
One more small tweak — tightened the substring UPRN check to exact match and raised `SourceArgumentNotFound` when no match is found instead of falling through to an `AttributeError`. Thanks again for the original fix.

PR #6357, opened 2026-05-18
Add Aprica S.p.A. (apricaspa.it) source

### LINKED ISSUE #4064: [Source Request]: Aprica SPA
### Municipality / Region

Italy

App PULIamo

### Example Address

_No response_

### Collection Data Format

Dedicated IOS/Android APP

### Additional Information

The app is really ugly (no notifications, slow), viewing and using the data it provides in a different way would help many people.

https://play.google.com/store/apps/details?id=com.modomodo.mobile.a2a&hl=it
https://apps.apple.com/it/app/puliamo/id524983253

nu00:
Adds support for [Aprica S.p.A.](https://www.apricaspa.it/) (Gruppo A2A), serving waste collection schedules for municipalities in the provinces of Brescia, Bergamo, Como, Cremona, Crema, Lodi and others.
Uses the same /api/service/area-services/ REST API as the existing amsa_it source (both are A2A group companies sharing the same backend).
Configuration: address, house_number, city (all required).
Tested with: public addresses in Brescia and Bergamo — returns correct calendar data (indifferenziato, organico, vetro e metallo, carta e cartone, plastica).
Closes #4064

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/apricaspa_it.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/apricaspa_it.py
@@ -1,12 +1,12 @@
 import datetime
-from typing import Any, Dict, List, Optional, Tuple
+from typing import Any, Dict, List, Tuple
 
 import requests
 from waste_collection_schedule import Collection
 from waste_collection_schedule.exceptions import (
-    SourceArgumentNotFound,
     SourceArgumentException,
     SourceArgumentExceptionMultiple,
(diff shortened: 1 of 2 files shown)


markvp (approved):
Nice work — drops in cleanly alongside amsa_it.py and live-tested fine against all three test cases. I pushed a couple of small tweaks: dropped an unused import, ran black/isort, and added the doc/source/apricaspa_it.md page that's needed for the README listing. Thanks for picking this up!

PR #6367, opened 2026-05-18
feat: add "Ignore Duplicate Entries per Day" option to UI

christianh87:
## Summary

- Adds a new boolean option **"Ignore Duplicate Entries per Day"** (default: off) to the per-source options UI
- When enabled, collection entries sharing the same waste type and date are deduplicated at fetch time, keeping the first occurrence
- Deduplication runs at the end of the `SourceShell.fetch()` pipeline — after whitespace stripping, filtering, customization, and day-offset — so aliases and offsets are respected before comparison

## Motivation

Some waste collection providers (e.g. via shared platforms or buggy APIs) return the same waste type for the same day multiple times. This causes sensors and the HA calendar to show redundant entries. Rather than fixing every source individually, this opt-in flag gives users a simple one-click fix via the integration's options flow.

## Changes

| File | What changed |
|---|---|
| `const.py` | Added `CONF_IGNORE_DUPLICATES` constant and `CONF_IGNORE_DUPLICATES_DEFAULT = False` |
| `source_shell.py` | `ignore_duplicates` param added to `__init__` and `create()`; dedup loop appended to `fetch()` |
| `init_ui.py` | Reads the new option from `entry.options` and passes it to `SourceShell.create()` |
| `config_flow.py` | Imports `BooleanSelector`; adds the checkbox after "Day Offset" in `async_step_init` |
| `translations/en.json` | Label + description for the new field in `options.step.init` (hand-maintained section) |

## Reviewer notes

- Default is `False` → zero behaviour change for existing users
- No migration needed (the option simply defaults to `False` when absent)
- No generated files were touched

## Test plan

- [ ] Install integration, open Options for a source → new "Ignore Duplicate Entries per Day" toggle is visible
- [ ] Enable the toggle for a source that returns duplicates → duplicate entries disappear from sensor/calendar
- [ ] Disable the toggle → duplicates reappear
- [ ] Sources without duplicates are unaffected when the flag is on

### DIFF
--- a/custom_components/waste_collection_schedule/config_flow.py
+++ b/custom_components/waste_collection_schedule/config_flow.py
@@ -18,6 +18,7 @@ from homeassistant.config_entries import (
 from homeassistant.const import CONF_NAME, CONF_VALUE_TEMPLATE
 from homeassistant.core import callback
 from homeassistant.helpers.selector import (
+    BooleanSelector,
     DurationSelector,
     DurationSelectorConfig,
     IconSelector,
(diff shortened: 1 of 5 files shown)


christianh87:
Closing — PR was opened by mistake. Will be re-submitted after testing in the contributor's fork.

PR #6368, opened 2026-05-18
feat: add "Ignore Duplicate Entries per Day" option to UI

christianh87:
## Issue

Some waste collection providers return the same waste type for the same day multiple times, causing sensors and the HA calendar to show redundant entries (e.g. \"Restmüll\" or \"Schadstoffmobil\" appearing 2–3× on the same day).

## Solution

Adds a new boolean option **"Ignore Duplicate Entries per Day"** (default: off) to the per-source options UI. When enabled, collection entries sharing the same waste type and date are deduplicated at fetch time, keeping the first occurrence.

### ❌ Before — duplicates visible (option disabled)

Function_disabled

Function_disabled_Result

### ✅ After — deduplicated (option enabled)

Function_enabled

Function_enabled_Result

## Changes

| File | What changed |
|---|---|
| `const.py` | Added `CONF_IGNORE_DUPLICATES` constant and `CONF_IGNORE_DUPLICATES_DEFAULT = False` |
| `source_shell.py` | `ignore_duplicates` param added to `__init__` and `create()`; dedup loop appended to `fetch()` |
| `init_ui.py` | Reads the new option from `entry.options` and passes it to `SourceShell.create()` |
| `config_flow.py` | Imports `BooleanSelector`; adds the checkbox after "Day Offset" in `async_step_init` |
| `translations/en.json` | Label + description for the new field in `options.step.init` |
| `translations/de.json` | German translation |
| `translations/it.json` | Italian translation |

## Notes

- Default is `False` → zero behaviour change for existing users
- No migration needed (the option simply defaults to `False` when absent)
- No generated files were touched
- Deduplication runs after whitespace stripping, filtering, customization, and day-offset — so aliases and offsets are respected before comparison

## Test plan

- [x] New "Ignore Duplicate Entries per Day" toggle is visible in the options UI
- [x] Enable the toggle → duplicate entries disappear from sensor/calendar
- [x] Disable the toggle → duplicates reappear
- [x] No errors in HA logs


markvp (approved):
Nice, useful option — and the dedup runs after customise/day-offset so aliases are honoured. Reordered the new `const` imports to keep isort happy, otherwise looks great. Thanks!

christianh87:
You're welcome and thanks for the feedback, I'll keep that in mind for next time. I'm happy to contribute something as well. :blush:

PR #6375, opened 2026-05-18
South Lanarkshire: add collection details for this region

### LINKED ISSUE #5670: [Source Request]:
### Municipality / Region

South Lanarkshire Scotland 

https://www.southlanarkshire.gov.uk/downloads/file/18301/east_kilbride_cambuslang_and_rutherglen_bin_collection_calendar_2026

### Example Address

_No response_

### Collection Data Format

As PDF Download

### Additional Information

_No response_

campbellcopeland:
Resolves #5670


markvp (changes requested):
Hey @campbellcopeland — thanks so much for taking this one on. South Lanarkshire has been a long-standing request and you've clearly put a lot of work into it; the form-handling and session/UPRN flow are pieces I'd be happy to see in a future revision.

Before we can merge this one though, I'd love to talk through a few things in the schedule-generation logic — they're substantial enough that I think the easiest path is to rework them rather than patch in place. Happy to discuss approach if any of this is unclear:

- **The PDF isn't actually informing the output.** In `_parse_pdf_schedule`, dates are extracted from the PDF but bin types are then assigned via `date_obj.isocalendar()[1] % 4` against a hardcoded `[Black, Grey+Burgundy, Black, Blue+Burgundy]` cycle — the bin keywords next to each date are never used. So the result is effectively a baked-in 4-week pattern aligned to ISO weeks, which won't necessarily match any given user's actual schedule. It'd be worth either parsing the PDF for real (associating each date with the bin colour next to it in the text), or finding a way to do without the PDF entirely.
- **The PDF URL parameter will go stale yearly.** The URLs include the year (`..._2026.pdf`), so each January every user's config breaks. If we do keep PDFs, the source should derive the URL itself from the listing page at `/downloads/download/791/...`.
- **Bin pattern varies per household.** South Lanarkshire publishes different calendars for different household configurations ("4-bin" vs "food and garden same week as general"), so a single hardcoded cycle can't fit everyone. Worth considering how to detect or expose that.

Two smaller items while you're in there:
- `SourceArgumentNotFound` / `SourceArgumentNotFoundWithSuggestions` from `waste_collection_schedule.exceptions` are preferred over generic `Exception` / `ValueError` — they surface nicely in the HA UI.
- A `doc/source/southlanarkshire_gov_uk.md` is needed for every new source.
- `tests/test_southlanarkshire_enhanced.py` (833 lines) is a lot heavier than other per-source tests in the repo — compare `test_midlothian_gov_uk.py` (~150 lines) for the lighter pattern most other sources use. Once the schedule logic is reworked, happy to drop the test file or pare it down.

A couple of nearby sources that might be useful references for "the site only shows one week" problems: `south_ayrshire_gov_uk.py` and `east_renfrewshire_gov_uk.py` — both either parse a wider window from the source page or hit the underlying API the site itself uses, which avoids the PDF dependency entirely.

Thanks again for the effort — happy to chat through approach if helpful before you take another pass.

campbellcopeland:
Resolves https://github.com/mampfes/hacs_waste_collection_schedule/issues/5670

This reworks the South Lanarkshire source to use the council's waste services portal API instead of deriving schedules from PDF calendars.

Summary:

replace the PDF-based schedule generation with the live SelectPrem portal flow
fetch a CSRF token from the dashboard, submit postcode and UPRN, and parse the returned appointments data
use postcode-scoped premises suggestions when the supplied UPRN is invalid
add request timeouts and deterministic sorting to match established source patterns
add focused unit coverage for parsing, deduplication, sorting, timeout usage, and negative-path error handling
include the required source documentation
Why:

avoids hardcoded cycle assumptions and yearly PDF URL breakage
follows the same approach used by nearby council integrations that query the underlying site flow directly
aligns error handling and testing with existing patterns in this repository
Validation:

formatted with black and isort
verified with the dedicated South Lanarkshire unit tests
live-tested against the source TEST_CASES during development
Note:

this supersedes the earlier PDF-based approach discussed in https://github.com/mampfes/hacs_waste_collection_schedule/pull/6375 and uses the current branch southlanarkshire-derived-labels instead.

markvp (approved):
Thanks for the thorough rework, @campbellcopeland — this is a big step up. Hitting the PublicDashboard `SelectPrem` flow directly removes the hardcoded cycle, the yearly PDF URL drift, and the household-pattern guessing in one go. Error handling and the suggestions fallback are spot on, and the test file size now lines up with other per-source tests in the repo. Live-tested all four TEST_CASES against the council site and they return clean, deduplicated, sorted schedules. Made a small tweak to migrate ICON_MAP to the canonical `Icons` enum (your branch predated #6384). Nice work.

Issue #3709, opened 2025-01-31
[Source Request]:  [Source Request]: Town of Bassendean, WA, Australia


crm940:
### Municipality / Region

Bassendean, Perth, Australia

### Collection Calendar Webpage

https://maps.bassendean.wa.gov.au/IntraMaps97/ApplicationEngine/frontend/mapbuilder/default.htm?configId=00000000-0000-0000-0000-000000000000&liteConfigId=ce26eebb-7d54-4e49-8d4e-438bc3b10456&title=RmluZCBZb3VyIEJpbiBDb2xsZWN0aW9uIERheQ==

### Example Address

2 JUBILEE AVENUE, EDEN HILL

### Collection Data Format

As html on a webpage

crm940:
@markvp Sorry to be a pain, but this configuration doesn't fully work for me. It gives the correct dates for one type of bin (FOGO) but incorrect dates for the "general waste" and "Recycling" services. E.g. the bins are collected Friday, the FOGO is reporting a correct date of: 2026-05-22 but the General Waste is reporting a date of 2026-05-27.
Are you able to help with this? or do I raise a new issue?

PR #6386, opened 2026-05-19
Update waste collection schedule base dates by 3 days

olie-ob:
Fixes dates for waste collection on Town of Bassendean integration.
Dates were 3 days behind (ie a proper bin day of Tuesday was shown as a Saturday) on general and recycling bins. Updated the dictionary of initial dates to have starting Mondays Tuesdays and so on, so that the dates are reflected in the calendar. 
It seems to work for this week with some test addresses.


markvp (approved):
Thanks Olie — nicely diagnosed. The originals had the day-of-week labels glued onto dates whose actual weekday didn't match, which is exactly the 3-day skew users were seeing; the new dates line up properly across all five weekdays for both bins. Verified the math against crm940's report on #3709 — that case now resolves correctly. Approving.

Issue #1408, opened 2023-11-11
[Feature]: Hide default waste calendar if all waste types use dedicated calendars 


nurtext:
### I propose a feature for:

The integration in general

### Describe your wanted feature

I'm currently using dedicated calendars for every type of waste. This results in the default waste calendar being empty. I therefore propose to hide the default calendar, if all waste types use dedicated calendars.

5ila5:
not sure how hard this would be to implement, but you could just change the calendar_title to of the source and add use_dedicated_calendar to all but one types

nurtext:
Great workaround idea, cheers!

markvp:
Thanks for the suggestion, Cedric! This is a reasonable UX improvement — when every waste type has `use_dedicated_calendar: true`, the default aggregated calendar ends up empty and is just clutter.

In the meantime, there is a partial workaround: keep `use_dedicated_calendar: false` on one waste type (the one you care least about having in its own calendar) — that type will remain in the default calendar and the others will each have their own. Alternatively, you can suppress the default calendar from view using HA's calendar visibility settings.

The proper fix would be in `calendar.py`'s `create_calendar_entries()` — skip creating the default calendar entity when all types are covered by dedicated calendars. If anyone wants to pick this up, the relevant code is in `custom_components/waste_collection_schedule/calendar.py` around line 146. A PR is very welcome!

Leaving this open as a tracked enhancement.

nurtext:
@markvp Tried my best to make a tiny PR at #6387 - thanks for pointing out the place in the code!

PR #6387, opened 2026-05-20
Skip creating default calendar if all calendars are marked dedicated

nurtext:
Tried to fix the issue of creating the default calendar if all calendars are marked as dedicated ones. Therefore the default calendar won't get created at all.  See #1408 for details. 

Disclaimer: I'm not very familiar with Python coding in general, so feel free to correct my code if I made a mistake.


markvp (approved):
Thanks Cedric — nicely scoped fix and exactly the right spot to touch. Hoisting the aggregator and the alias-mapped set, then gating the default calendar on `aggregator.types != dedicated_calendar_types`, does the job without any extra moving parts. Approving as-is.

Issue #3968, opened 2025-03-27
[Source Request]: Could you plese add Airdrie in Alberta, Canada


Dashtalks:
### Municipality / Region

Airdrie, Alberta, Canada

### Collection Calendar Webpage

https://www.airdrie.ca/index.cfm?serviceID=1880

### Example Address

https://www.airdrie.ca/index.cfm?serviceID=1880

### Collection Data Format

As html on a webpage

lit-af:
This issue can be closed @markvp

PR #6406, opened 2026-05-21
feat: add source for Airdrie waste collection schedule (airdrie_ca)

lit-af:
Source for: #3968


markvp (approved):
Thanks Alexandre — really clean implementation, and the live test against Coopers Crossing pulled a full year of entries without complaint. Made a couple of small tweaks: switched `TITLE` to `"Airdrie (AB)"` to match the province-suffix style used by the other Canadian sources, and fixed an `ICON_MAP` key so organics events pick up the leaf icon (the API returns `"Organics collection"` with a lowercase c). Approving on that basis.

Dashtalks:
Thank you!!!!

PR #6415, opened 2026-05-22
feat: add Derbyshire Dales

moonthug:
Add Derbyshire Dales district council source


markvp (approved):
Thanks for adding Derbyshire Dales! Made a couple of small formatting tweaks (black/isort) and this is good to go.

PR #6420, opened 2026-05-23
fix(colchester_gov_uk): drop textiles and combine recycling streams

Canis-L-Sapien:
## Summary

  Colchester changed two parts of its kerbside collection schedule in 2026:

  - **2026-04-03**, textiles moved from kerbside collection to a [bookable
  service](https://www.colchester.gov.uk/recycling-and-rubbish/book-textile-collection/) and no longer appear on the recycling calendar. Removes the
  `Textiles` entry from `ICON_MAP`.
  - **2026-06-01**, paper/card, plastics, tins and cans are now collected together as a single "Mixed recycling" stream (per the [council's strategy page](https://www.new.colchester.gov.uk/recycling-and-rubbish/recycling-and-waste-strategy-2025-2040-implementation/calendar-collection)). The API has already cut over (no collection entries returned before the cutover date), so the old `Cans` and `Plastics` row names no longer appear in the response.
  The API still emits `"Paper/card"` as the row name for the new combined stream, so a small `NAME_MAP` rewrites the display name to `"Mixed recycling"`, and the icon is switched from `PAPER` to `RECYCLING`. The now-unused `Cans` and `Plastics` entries are removed from `ICON_MAP`.

  The two commits are kept separate so each change can be reverted independently.

  ## Test plan

  - [x] `python -m pytest tests/test_source_components.py`: 6 passed
  - [x] `python test_sources.py -s colchester_gov_uk -l`: both `Church Road, Colchester` and `The Lane, Colchester` TEST_CASES return collections with the
  expected `Mixed recycling` label, no Textiles/Cans/Plastics entries
  - [x] `black`, `isort`, `flake8`: clean

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
@@ -17,7 +17,6 @@ ICON_MAP = {
     "Black bags": Icons.GENERAL_WASTE,
     "Glass": Icons.GLASS,
     "Cans": Icons.METAL,
-    "Textiles": Icons.TEXTILE,
     "Paper/card": Icons.PAPER,
     "Plastics": Icons.PLASTIC_PACKAGING,
     "Garden waste": Icons.GARDEN,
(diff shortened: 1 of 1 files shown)


markvp (approved):
Thanks for keeping this in step with Colchester's service changes — the NAME_MAP + icon swap reads cleanly and the two-commit split is appreciated. Approving.

PR #6421, opened 2026-05-23
OKC official API change to new ArcGIS REST API 

totallydifferent:
**Problem Solved**
- OKC official API (data.okc.gov) was no longer working
- Needed to use a new address-based approach using ArcGIS REST services
- Legacy Object ID fields were confusing and unnecessary

**Solution Implemented**
- New ArcGIS REST API integration with address geocoding and spatial queries
- Simplified configuration: Only 3 fields (try_official, address, objectID)
- Complete waste coverage: Trash, Recycling, and Bulky Waste schedules
- Clean user experience: Street address only (no city/state/zip needed)

**Key Changes**
- Updated API endpoints to use gis.okc.gov/arcgis/rest/services/Public
- Added address geocoding using Street_Geocoding_Service
- Implemented spatial queries for waste zone detection
- Fixed parsing for both PickupDay and PickUpDay field variations
- Removed legacy Object ID fields (bulkyObjectID, recycleObjectID, trashObjectID)
- Updated documentation with clear address-based examples

### DIFF
--- a/README.md
+++ b/README.md
@@ -3296,7 +3296,7 @@ If you already have enough information for your municipality/region, you are ver
 - [City of Los Angeles, CA](/doc/source/lacity_gov.md) / lacitysan.org
 - [City of Lowell, MA](/doc/ics/recollect.md) / lowellma.gov
 - [City of McKinney, TX](/doc/ics/recollect.md) / mckinneytexas.org
-- [City of Oklahoma City](/doc/source/okc_gov.md) / data.okc.gov
+- [City of Oklahoma City](/doc/source/okc_gov.md) / okc.gov
 - [City of Philadelphia, PA](/doc/source/phila_gov.md) / phila.gov
 - [City of Pittsburgh](/doc/source/pgh_st.md) / pgh.st
 - [City of Plano](/doc/source/plano_gov.md) / plano.gov/630/Residential-Collection-Schedules
--- a/custom_components/waste_collection_schedule/source_metadata.json
+++ b/custom_components/waste_collection_schedule/source_metadata.json
@@ -6099,7 +6099,7 @@
   "okc_gov": {
     "docs_url": "https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/okc_gov.md",
     "howto": {
-      "en": "For official mode with try_official=true, provide specific Object IDs from OKC Open Data Portal (bulkyObjectID, recycleObjectID, trashObjectID). Object IDs can be found by exploring the OKC waste collection datasets at https://utility.arcgis.com/usrsvcs/servers. If try_official is disabled, provide objectID for the unofficial source."
+      "en": "Provide the OBJECTID of one or more of your collection zones from the OKC Open Data Portal (trashObjectID, recycleObjectID, bulkyObjectID). At least one is required. Open the FeatureServer layers linked in the documentation, use the Query page to locate the zone that covers your address, and read its OBJECTID."
     },
     "urls": {}
   },
--- a/custom_components/waste_collection_schedule/source_metadata.json
+++ b/custom_components/waste_collection_schedule/source_metadata.json
@@ -6099,7 +6099,7 @@
   "okc_gov": {
     "docs_url": "https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/okc_gov.md",
     "howto": {
-      "en": "Using your browser, go to https://data.okc.gov/portal/page/viewer?datasetName=Bulky%20Waste%20Zones&view=map and search for your address. Go to the table tab and filter by address. Find your Object ID. Next use the dropdown in the top right hand side of the screen to change the dataset from Bulky Waste Zones to Recycle Zones. Once again Filter by Map to find your Object ID. Do this once more for your Trash Zone. Once you have these three Object IDs, enter them in bulkyObjectID, recycleObjectID, and trashObjectID and enable try_official. If try_official is disabled, provide objectID for the unofficial source."
+      "en": "For official mode with try_official=true, provide specific Object IDs from OKC Open Data Portal (bulkyObjectID, recycleObjectID, trashObjectID). Object IDs can be found by exploring the OKC waste collection datasets at https://utility.arcgis.com/usrsvcs/servers. If try_official is disabled, provide objectID for the unofficial source."
     },
     "urls": {}
   },
--- a/custom_components/waste_collection_schedule/translations/de.json
+++ b/custom_components/waste_collection_schedule/translations/de.json
@@ -25540,6 +25540,7 @@
           "bulkyObjectID": "Bulky Object Id",
           "objectID": "Object Id",
           "recycleObjectID": "Recycle Object Id",
+          "recycle_reference_date": "Recycle Reference Date",
           "trashObjectID": "Trash Object Id",
           "try_official": "Try Official"
         },
@@ -25555,6 +25556,7 @@
           "bulkyObjectID": "Bulky Object Id",
           "objectID": "Object Id",
           "recycleObjectID": "Recycle Object Id",
+          "recycle_reference_date": "Recycle Reference Date",
           "trashObjectID": "Trash Object Id",
           "try_official": "Try Official"
         },
--- a/custom_components/waste_collection_schedule/translations/de.json
+++ b/custom_components/waste_collection_schedule/translations/de.json
@@ -25538,11 +25538,8 @@
         "data": {
           "calendar_title": "Kalender Titel",
           "bulkyObjectID": "Bulky Object Id",
-          "objectID": "Object Id",
           "recycleObjectID": "Recycle Object Id",
-          "recycle_reference_date": "Recycle Reference Date",
-          "trashObjectID": "Trash Object Id",
-          "try_official": "Try Official"
+          "trashObjectID": "Trash Object Id"
         },
         "data_description": {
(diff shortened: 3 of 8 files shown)


markvp (changes requested):
Thanks for taking this on — the original `data.okc.gov` endpoint going dark is a real problem and the ArcGIS approach is a nice direction. A few things I'd like to see before this can land:

**Breaking change for existing users.** This removes `bulkyObjectID` / `recycleObjectID` / `trashObjectID` entirely. Anyone currently running `try_official: true` with those three zone IDs will get a constructor error after upgrading. Could we either keep both flows side by side (legacy zone IDs *or* new address-based), or call this out explicitly in the PR description and the doc as a breaking config change with migration guidance?

**Cleanup needed.** The file has some accidental duplication that I think slipped in during a merge:
- Lines 1-6 import `re`, `datetime`, and `quote` twice
- The entire `fetch()` method is defined twice at lines 476-486
- `json_data = response.json()` is duplicated at line 267
- `_fetch_rows` / `_fetch_record` still reference `API_BASE_URL`, but you removed that constant — those methods would now `NameError` if anything called them. They look like dead code; can be removed or the constant restored.

Same kind of duplication in `doc/source/okc_gov.md`: the `# City of Oklahoma City` header appears twice at the top, and the `**objectID**` config variable block is duplicated.

**CI will fail on the icons.** `ICON_MAP` regressed from the canonical `Icons` enum back to raw `mdi:*` strings. Please put it back to `Icons.GENERAL_WASTE` / `Icons.RECYCLING` / `Icons.BULKY` (see #6384 and `waste_collection_schedule/icons.py`) — `test_icon_map_uses_canonical_icons` will reject the raw strings.

**No tests for the new code path.** `TEST_CASES` still only covers the unofficial `objectID` mode. Please add at least one address-based test case so we can verify the new ArcGIS flow against the live API in CI.

**A couple of smaller things:**
- `_fetch_official` swallows every per-layer exception with a bare `except Exception: continue` — that risks returning a silently incomplete schedule when one layer is unavailable. Worth narrowing it to specific exceptions or at least logging.
- The geocoder param key `"Single Line Input"` (with spaces) looks suspicious — ArcGIS GeocodeServer usually accepts `SingleLine`. Worth confirming against the live endpoint.
- Don't worry about the generated files (`README.md`, `info.md`, `sources.json`, `source_metadata.json`, `translations/*.json`) — those are regenerated by CI post-merge, you can revert them with `git checkout upstream/master -- <file>`.

Happy to take another look once these are sorted.

totallydifferent:
thanks for your quick review and input @markvp - really appreciate it!

totallydifferent:
had the ArcGIS working and they restricted access to the endpoints! So had to revert back to the previous methodology - it is working now through a publicly accessible feed. also added another field to handle the starting date for the alternate week logic for recycling pickup. sorry for the multiple commits - its late and had a bunch of things to iteratively squash.

markvp (changes requested):
Appreciate you sticking with this and the quick turnaround on the earlier review. Reading the latest state and your follow-up comment, I think there's still some untangling to do before this can land. A few things from the previous review remain, and the recent commits seem to have re-added the ArcGIS / address path even though you mentioned reverting it.

Could you take another pass with this in mind:

**Pick one official path.** Your comment said the ArcGIS endpoints got locked down, but `_fetch_official` still calls Nominatim + `utility.arcgis.com` feature servers, and `Test_004` (`address: "13917 Marsala Ct"`) targets that path. If those endpoints don't work, please remove the address-based flow and `Test_004` entirely and keep the zone-ID approach. If they do work, please drop the unused Incapsula-bypass helper (`_make_arcgis_request`), the `USER_AGENTS`, `ARCGIS_BASE_URL`, `WASTE_LAYER_URL`, `OPEN_DATA_BASE_URL`, and the unused `from urllib.parse import quote`. There's also an unreachable branch in `_geocode_address` (`service_name != "Oklahoma City"` is always true) and a stray `print()` which should come out.

**Geocoder fallback to the city centre is a footgun.** If Nominatim can't pinpoint the address, returning OKC centre coords will silently return whatever zone happens to cover the city centre — much better to raise `SourceArgumentNotFoundWithSuggestions` and let HA surface a useful error.

**Generic exceptions.** A few spots (constructor when args are missing, `_geocode_address` when nothing matches, `_fetch_official` when zones aren't found) should use `SourceArgumentNotFound` / `SourceArgumentNotFoundWithSuggestions` from `waste_collection_schedule.exceptions`. The bare `except Exception: continue` in the per-layer loop is also worth narrowing.

**Doc duplication.** `doc/source/okc_gov.md` has the "Unofficial example" block twice (lines 115-124 and 126-135). Probably a merge artefact.

**Generated files.** `README.md`, `info.md`, `sources.json`, `source_metadata.json`, and `translations/{de,en,fr,it}.json` are still in the diff — those are regenerated by CI post-merge, so please revert with `git checkout upstream/master -- <file>`.

**Rebase.** Your branch is ~1000 files behind upstream `master` (we landed the canonical `Icons` enum since you forked), which is why the diff shows changes to files you didn't touch.

**`recycle_reference_date`.** The new "supply a known anchor date" field looks like a workaround for parsing the recycling pickups incorrectly. The unofficial endpoint already returns explicit `pickups[]` arrays — could we just use those?

Quite a bit there, sorry — happy to chat through any of it. Once it's stabilised I'll take another look.

markvp:
Thanks for sticking with this, @totallydifferent. Since the official data.okc.gov / ArcGIS endpoints are locked behind Incapsula, I've reworked the source to use only the public okc.schizo.dev feed: it reads the explicit pickups[] dates per section directly (so the weekday/ordinal inference and the recycle_reference_date anchor are gone), uses SourceArgumentNotFound for error paths, and moves ICON_MAP to the Icons enum. The official/ArcGIS/address machinery is removed. I've pushed this to the branch.

Important caveat: I could NOT live-test it — okc.schizo.dev is currently returning 502 Bad Gateway on every endpoint (including /health), so the backend appears to be down right now. The parser is built against the JSON shape from the previous code (each section with a pickups[] of {date}). Could you run `python test_sources.py -s okc_gov -l` against your setup once the feed is back up and confirm it returns the correct dates? If the shape has changed we can tweak the parser. Thanks!

totallydifferent:
hey @markvp sorry i took so long - have been a bit tied up. i've just submitted some changes based on your feedback.
also the unofficial site is down for me as well - so I am unable to use that at the moment. hence the perseverance with the official pathway. also with regards to recycle_reference_date - as the unofficial endpoint is down, without providing a starting date for this it can't work out the schedule correctly unfortunately.

totallydifferent:
Fixed the official API by switching from blocked data.okc.gov to working ArcGIS endpoints
Added recycle_reference_date field for every-other-week recycling sync
Now ready for review

markvp (approved):
Thanks for tracking down the new ArcGIS endpoints — I confirmed all three FeatureServer layers (Trash/Recycle/Bulky) respond cleanly with live zone data, so the official path is solid. I took the liberty of pushing a couple of small tweaks on top: dropping the old okc.schizo.dev fallback and the recycle_reference_date workaround entirely (now that the official API resolves the schedule directly), wiring the errors through SourceArgumentNotFound, and refreshing the TEST_CASES/docs with real verified Object IDs. Live test returns the expected Friday/Monday/4th-Monday pickups. Approving once these land.

totallydifferent:
@markvp Hah just as luck would have it - recycling date was off by a week. doh! June 12 reported. Should've been June 19. Don't worry - I managed to get the entire street to follow suit so I don't feel too foolish. What was the tweak you made for the recycling date piece?

markvp:
Good news — you actually gave me everything I needed. Post-mortem: when we moved to the official ArcGIS zone layers, the source ended up only reading the `PickupDay` weekday and returning its next occurrence. I checked the live recycle layer and it exposes only `OBJECTID`, `Route`, `PickupDay` and `ServiceProvider` — no date or week-A/B field anywhere, so nothing in the official feed can say *which* week of the fortnight is "on". Hence right weekday, wrong week.

Since June 19 is your confirmed good date, I've put up a small follow-up that reinstates a `recycle_reference_date` anchor — set it to one real recycling date and the source projects the every-other-week cadence from there (verified: June 19, then July 3, skipping June 26): #6621

Once it lands, add to your config:
```yaml
recycle_reference_date: "2026-06-19"
```
And if you're happy to share your recycle zone's OBJECTID, I'll add a proper live test for a Friday zone (the current test case is the Monday zone 1215).

totallydifferent:
Here you go:
[Bulky Waste Zones ](https://data.okc.gov/portal/page/viewer?datasetName=Bulky%20Waste%20Zones&view=map)(Dataset ID 86) 
Object ID 14, Pickup Day 4th Monday, Provider Waste Management

[Recycle Zones](https://data.okc.gov/portal/page/viewer?datasetName=Bulky%20Waste%20Zones&view=map) (Dataset ID 89)
Object ID 1366, Pickup Day Friday, Provider Waste Management

[Trash Zones](https://data.okc.gov/portal/page/viewer?datasetName=Bulky%20Waste%20Zones&view=map) (Dataset ID 87)
Object ID 315, Pickup Day Friday, Provider Waste Management

markvp:
Perfect, thanks! I checked all three against the live FeatureServer layers and they line up exactly with what you described: Recycle 1366 → Friday, Trash 315 → Friday, Bulky 14 → 4th Monday (all Waste Management).

I've added the Friday recycle zone (1366) as a live test for the every-other-week path, anchored to your confirmed 19 June date, so we've now got a real Friday zone alongside the existing Monday one. Verified it returns 2026-06-19. Follow-up PR: #6631.

PR #6437, opened 2026-05-25
Fix grafikai_svara_lt: send Sec-Fetch-Site: same-origin

sanis:
Upstream grafikai.svara.lt now rejects _serverFn requests with HTTP 403 unless Sec-Fetch-Site: same-origin is present. Without this header both the discovery probe and the data calls fail, so the integration produces no events and the calendar stays empty.

Add the header to both _probe_data_fetcher and _raw_call.


valdaspauliukonisflowforma (approved):

markvp (approved):
Thanks for tracking this down — clean, minimal fix and the inline comment explaining the 403 is exactly what future-me wants to see when this comes up again.

PR #6439, opened 2026-05-25
Odpad Slovensko 

frantisektalapka-create:
Waste slovakia

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/odvoz_odpadu_sk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/odvoz_odpadu_sk.py
@@ -1,0 +1,36 @@
+import datetime
+import json
+import requests
+from waste_collection_schedule import Collection
+
+TITLE = "Odvoz Odpadu"
+DESCRIPTION = "Zdroj pre harmonogram odvozu odpadu z portálu odvoz-odpadu.eu."
+URL = "https://odvoz-odpadu.eu"
+COUNTRY = "sk"
+
+# TEST_CASES slúži pre automatický test na GitHube, aby overil funkčnosť obce Dolná Ves
+TEST_CASES = {
+    "Dolná Ves": {"city": "dolna_ves"}
+}
+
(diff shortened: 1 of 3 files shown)


markvp (changes requested):
Hi @frantisektalapka-create — thanks for picking up Slovak coverage, `odvoz-odpadu.eu` is a great target.

A few things need attention before we can merge:

1. **The Python source can't fetch anything.** `odvoz_odpadu_sk.py` POSTs to `https://odvoz-odpadu.eu/api/v1/get-by-municipality`, but that endpoint returns 404 with the app's own Slovak "Požadovanú stránku nebolo možné nájsť!" page — there is no public JSON API at that path. The portal *does* serve a working ICS feed at `https://odvoz-odpadu.eu/miesto/<id>/kalendar.ics`, which returns valid `text/calendar` data. The simplest and most maintainable path is to **drop the Python source entirely** and ship this as an ICS provider only, via a single `doc/ics/yaml/odvoz_odpadu_sk.yaml` file.

2. **The two YAML files need cleanup.** Both are missing the `.yaml` extension, so `update_docu_links.py` will skip them and your provider won't appear in `README.md` / `info.md` / `sources.json`. One filename also contains spaces, parentheses, and a trailing space, and ends with a stray `a` character. Please replace both with a single file at `doc/ics/yaml/odvoz_odpadu_sk.yaml`.

3. **Add an `extra_info` block.** ICS-provider YAMLs need an `extra_info:` list (entries with `title`, `url`, `country`) so the provider is listed correctly. Have a look at `doc/ics/yaml/svoz_litovle_cz.yaml` for a similar small-provider example.

4. **Branch is behind upstream.** Your fork's `master` is well behind `mampfes:master` and has picked up an unrelated commit (`Update st_helens_gov_uk.py`). Could you create a fresh feature branch off the latest `mampfes:master` and re-open the PR from there? That'll also clear the noise from the diff view.

Heads-up: this is also a duplicate of #6441 (same contributor, same target). I'll close #6441 as a duplicate so we can focus the conversation here.

Happy to help review once you've had a chance to address these — the ICS-only approach should be quick to land.

PR #6446, opened 2026-05-26
odvoz_odpadu_sk.yaml

frantisektalapka-create:


markvp (approved):
Thanks for adding this! Made one small tweak — the ICS YAML files need `title`, `url`, and `country` at the top level (the doc generator and the structural test both key off those), so I moved them out of `extra_info` and dropped the wrapper. Merging shortly.

frantisektalapka-create:
Thanks

PR #6459, opened 2026-05-28
Golden Plains Shire: fix HTTP 403 caused by stricter request validation (add browser headers + session handling)

smaurer3:
**Issue**

All requests to the Golden Plains Shire Council site began returning:

`403 Client Error: Forbidden`

This affected both:

- address autocomplete endpoint
- schedule lookup page

The site still functions normally in a browser, indicating the issue is related to request validation rather than API changes.

**Cause**

The server appears to have introduced stricter request filtering for non-browser clients, likely at a global level.

Requests made via requests were blocked due to missing browser-like characteristics, including:

- User-Agent header
- XHR / AJAX request headers
- Referer header
- Session cookies (not established in stateless requests)

**Changes**

- Introduced requests.Session() to maintain cookies across requests
- Added browser-like headers:
   - User-Agent
   - Accept
   - Referer
   - X-Requested-With
- Applied session handling to all requests (autocomplete + schedule page)

**Result**

- Restores functionality for both address lookup and schedule retrieval
- Aligns request behaviour with browser-like access patterns
- No changes to parsing logic or data model

**Notes**

This appears to be a server-side security or bot mitigation change rather than an API change. The site continues to work normally via a browser, suggesting the restriction is based on request fingerprinting rather than endpoint availability.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/goldenplains_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/goldenplains_vic_gov_au.py
@@ -1,7 +1,7 @@
 import re
 from datetime import datetime
 
-import requests
+from curl_cffi import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import (
@@ -27,17 +27,6 @@ CARD_TYPES = {
     "fogo": ("Glass", "mdi:bottle-wine"),
 }
 
-HEADERS = {
-    "User-Agent": (
-        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) "
-        "AppleWebKit/537.36 (KHTML, like Gecko) "
-        "Chrome/136.0.0.0 Safari/537.36"
-    ),
-    "Accept": "application/json, text/javascript, */*; q=0.01",
-    "Referer": "https://www.goldenplains.vic.gov.au/address-search",
-    "X-Requested-With": "XMLHttpRequest",
-}
-
 
 class Source:
     def __init__(self, address: str):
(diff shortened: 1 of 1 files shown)


markvp (changes requested):
Thanks for digging into this — confirming the 403 is real and your session/header change does clear it (I get a clean 200 from the autocomplete endpoint now instead of the forbidden response).

One snag though: after the 403 is gone, the autocomplete endpoint comes back with an empty result ([]) for the test address and for every partial query I tried, so fetch() still ends up raising SourceArgumentNotFoundWithSuggestions and the source returns no collections. I verified this isn't bot-fingerprinting or geo — curl_cffi with Chrome impersonation gets the same empty array, and the endpoint URL still matches the page's data-autocomplete-path. It looks like the address lookup query itself (params/dataset) has changed server-side on top of the 403.

Could you take another look at the autocomplete request — what query params the live site's address field actually sends — and confirm the source returns collections end-to-end against a known Golden Plains address? Since the host is behind Cloudflare, it'd also be worth switching to curl_cffi (requests.Session(impersonate="chrome"), see east_renfrewshire_gov_uk.py) while you're in there, as it's more resilient than static headers.

smaurer3:
Thanks Mark - it actually look like that the Council Website  is currently not working correctly as the Address search on the website is also responding with no results:
https://www.goldenplains.vic.gov.au/address-search

I can confirm that the other day when I fixed the 403 issue, it was workign up until today and I've already looked at it.  It may be worth just sitting on this PR for a few days until the Council Website is working correctly again from a browser, I will drop another comment to confirm if the integration is working again.

Been a weekend I don't expect anything will be fixed until at least Monday but who knows?  They have likely been changing / updating things on the website (hence the stricter rules creating the 403 error) but perhaps thy have inadvertently broken something

smaurer3:
Council website was back up and working on Sunday and still working now.  I have modified the code to use curl_cffi as per your suggestion.  Everything tested again and working successfully at 5:30pm AEST on 1/6/2026.

markvp:
sorry! just missed the release :'( should be up next week

markvp (approved):
Thanks for the quick turnaround — the switch to curl_cffi with Chrome impersonation is exactly what was needed, and routing both the autocomplete and schedule requests through the one impersonated session is clean. Glad the council site came back up and you were able to confirm it end-to-end. Structural tests pass locally. Approving — sorry again for missing the last release window, this'll go out in the next one.

PR #6467, opened 2026-05-29
fix(colchester_gov_uk): handle renamed waste types in calendar feed

Canis-L-Sapien:
## Summary

  Colchester's calendar API has renamed two of its waste streams:

  - `Black bags` -> `Non-recyclable rubbish`
  - `Paper/card` -> `Mixed recycling`

  Verified live against both `TEST_CASES` (`Church Road` and `The Lane`) on 2026-05-29 - both addresses now return the new names.

  Without this change `ICON_MAP.get("Non-recyclable rubbish")` and `ICON_MAP.get("Mixed recycling")` both miss, so those collections lose their icons. The
  new `Mixed recycling` label also gets mangled by the default `.title()` casing into `Mixed Recycling`, regressing the label that the previous fix (#6420)
  deliberately set.

  ## Changes

  - `ICON_MAP` gains entries for the two new names so icons resolve correctly.
  - `NAME_MAP` maps each new name to itself, bypassing the default `.title()` casing so `Mixed recycling` stays sentence case (matching #6420).
  - `NAME_MAP` keeps the legacy `Paper/card` mapping and adds `Black bags -> Non-recyclable rubbish` so users mid-rollout see the same canonical labels
  regardless of which feed version their address is on.
  - Adds `tests/test_colchester_gov_uk.py` covering the new and legacy feed shapes, an integration test with the exact JSON returned by the live API, and a
  date-filter regression test.ter / extrapolation behaviour.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
@@ -13,17 +13,25 @@ TEST_CASES = {
     "The Lane, Colchester": {"llpgid": "7cd96a3d-6027-e711-80fa-5065f38b56d1"},
 }
 
-# Colchester's API still emits "Paper/card" for the combined mixed-recycling
-# stream (paper/card, plastics, tins and cans) that replaced the separate
-# rounds.
+# The provider's calendar feed renamed two streams in mid-2026:
+#   "Black bags" -> "Non-recyclable rubbish"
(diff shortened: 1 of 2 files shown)


markvp (approved):
Thanks for the careful fix and the thorough regression tests. Verified live against both TEST_CASES — the API does now return "Non-recyclable rubbish" and "Mixed recycling", and the source maps them with the correct sentence-case labels and icons. Keeping the legacy Black bags/Paper/card mappings for the rollout window is a nice touch. Structural test suite passes. Approving as-is.

PR #6514, opened 2026-06-01
Colchester postcode lookup

Canis-L-Sapien:
## Summary

  Lets users set up the Colchester source with their postcode and house number/name
  instead of having to navigate the council site and copy a GUID. The existing
  `llpgid` argument continues to work unchanged — existing configs are untouched.

  The recycling calendar page already chains two anonymous endpoints:
  postcode -> address list (Dynamics 365 portal OData at `/_api/new_llpgs`) and
  then llpgid -> calendar. This PR teaches the source module to walk that chain
  itself when the user provides `postcode` + `house` instead of `llpgid`. The
  resolved llpgid is cached on the `Source` instance so subsequent polls still
  only hit the calendar endpoint.

  Also adds the missing `COUNTRY = "uk"` field as a separate prerequisite commit.

  ## Commits

  - `Add COUNTRY field to colchester_gov_uk source` — single-line prerequisite,
    avoids relying on the `*_gov_uk` filename fallback in `update_docu_links.py`.
  - `Add postcode/house lookup to colchester_gov_uk` — the feature.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/colchester_gov_uk.py
@@ -1,17 +1,52 @@
 import json
 from datetime import datetime, timedelta
+from typing import Optional
 
 import requests
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import (
+    SourceArgAmbiguousWithSuggestions,
+    SourceArgumentNotFound,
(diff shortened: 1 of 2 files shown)


markvp (approved):
Thanks for this — the postcode/house path is a really nice usability win, and keeping `llpgid` working unchanged is exactly right. I pushed a couple of small tweaks on top of your branch: used `SourceArgumentExceptionMultiple` for the "provide one of these" case so the error maps to real argument names, added a `timeout` to the calendar request (switching to `r.json()`), and let the exact-match branch also match on `new_name` so named properties resolve as the docs describe. Looks good to merge.

PR #6524, opened 2026-06-03
fix(sector27_de): resolve streets correctly

zaolin:
The upstream searchForStreets endpoint truncates the result list to a handful of matches, so sending only the first whitespace-separated token of the configured street causes a SourceArgumentNotFound when the street starts with a generic German article/preposition (e.g. 'Im', 'Am', 'An', 'Auf', 'Zum') and the distinctive word is the second one ("Im Overkamp" in Datteln being the first reported case).

The street lookup is refactored into _lookup_street_id, which iterates over every distinct token of the configured street, calls the upstream API for each, and returns the id of the first response whose name is an exact, case-insensitive, whitespace-stripped match for the configured street. If no token yields a match the existing SourceArgumentNotFound(WithSuggestions) error is raised with the union of suggestions gathered across the attempted searches.

Adds 'Datteln Im Overkamp' to TEST_CASES to lock in the regression. Removes an unused 'import logging' along the way.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sector27_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sector27_de.py
@@ -1,6 +1,5 @@
 import datetime
 import json
-import logging
 import re
 
 import requests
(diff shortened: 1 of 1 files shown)


markvp (approved):
Thanks for the clear write-up and the regression test case — the per-token search with an exact case-insensitive match is a clean way to work around the truncated searchForStreets results, and dropping the unused logging import is a nice touch. Verified locally: black/isort clean and the structural suite passes (6 passed). Looks good to merge.

PR #6527, opened 2026-06-03
fix: Update parsing of data for wigan_gov_uk

lurimax-north:
The data from the endpoint has changed recently. This updated the parsing to match the new data returned


markvp (approved):
Thanks for keeping this one current — the selector rename matches the site's updated markup. Structural tests pass on my end. Approving.

PR #6553, opened 2026-06-06
feat(rsag_de): add source for RSAG Rhein-Sieg Abfallwirtschaftsgesellschaft

### LINKED ISSUE #3551: [Source Request]:
### Municipality / Region

Königswinter, Germany

https://www.rsag.de/abfallkalender/abfuhrtermine

### Example Address

Winzerstrasse 84, 53639 Königswinter

### Collection Data Format

As PDF Download

### Additional Information

_No response_

markvp:
## Summary

- Adds a new source `rsag_de` for RSAG (Rhein-Sieg Abfallwirtschaftsgesellschaft), covering 19 municipalities in the Rhein-Sieg-Kreis, Germany.
- Uses the provider's public REST API to resolve city and street names to IDs, then fetches an ICS calendar.
- Parameters: `city` (e.g. `Königswinter`) and `street` (e.g. `Winzerstraße`).

Closes #3551

## Test plan

- [ ] `python -m pytest tests/test_source_components.py -q` — passes (6/6)
- [ ] Live: `python test_sources.py -s rsag_de -l` — returns 228 entries for Königswinter/Winzerstraße
- [ ] Verify street not found → `SourceArgumentNotFoundWithSuggestions` raised with correct suggestions
- [ ] Verify city not found → `SourceArgumentNotFoundWithSuggestions` raised

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/rsag_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/rsag_de.py
@@ -1,0 +1,146 @@
+import datetime
+
+import requests
+from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import (
+    SourceArgumentNotFoundWithSuggestions,
+)
+from waste_collection_schedule.service.ICS import ICS
+
+TITLE = "RSAG Rhein-Sieg Abfallwirtschaftsgesellschaft"
+DESCRIPTION = "Source for RSAG waste collection in the Rhein-Sieg-Kreis, Germany."
+URL = "https://www.rsag.de"
+COUNTRY = "de"
+
+TEST_CASES = {
+    "Königswinter, Winzerstraße": {
+        "city": "Königswinter",
+        "street": "Winzerstraße",
+    },
+    "Siegburg, Annostraße": {
+        "city": "Siegburg",
+        "street": "Annostraße",
+    },
+}
+
+ICON_MAP = {
+    "Restmüll": Icons.GENERAL_WASTE,
+    "Biotonne": Icons.ORGANIC,
+    "Bio": Icons.ORGANIC,
+    "Papier": Icons.PAPER,
+    "Wertstoff": Icons.PLASTIC_PACKAGING,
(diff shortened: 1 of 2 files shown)


David-Development:
@markvp Out of curiosity - how does this differ from the existing RSAG integration? ([This one](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/custom_components/waste_collection_schedule/waste_collection_schedule/source/muellmax_de.py)).

markvp:
Good question — I checked both live, and they actually hit the **same backend**: RSAG uses Müllmax as its portal vendor, and `rsag.de/api` is essentially a JSON/ICS front-end over the same dataset. For Alfter / `Ahrweg (105-Ende/94-Ende)`, both return 218 collections with identical dates and waste types — the native API even uses the exact same street-naming. The only data difference is that Müllmax prefixes every type with `"RSAG "`.

So `rsag_de` isn't replacing a dead feed — the Müllmax `Rsa` path still works. The advantage is UX: it takes plain `city`/`street` and resolves IDs via `rsag.de/api`, instead of `service="Rsa"` + the exact `mm_frm_str_sel` bracket-range string. Fair point on the overlap, though — since they're the same provider, I'm going to deprecate the Müllmax `Rsa` entry in favour of `rsag_de` so there aren't two listings for the same service.

Thanks for flagging it!

amazing-42:
@markvp : Not sure if this is the right place to adress this issue: I recognized that with the [release/3.0.0](https://github.com/mampfes/hacs_waste_collection_schedule/tree/release/3.0.0) a change from master at the rsag_de was intruduced, which apprently brakes the integration for me:
the def _clean_type(label: str) -> str:
replaces the detailed types to friendly names, which makes it impossible to filter on a specific type at the configuration of the sensor (i.e. "Restmülltonne 2-wö." becomes "Restmülltonne" and by this cannot be distinct form "Restmülltonne 4-wö." anymore.
Changing the function testwise to a simple
return lable
fixes this issue, but the
customize:
  alias: ...
at the sources: does not work to replace the ugly by friendly names anymore in this case (which was the way to go with the former muellmax_de source).
Any hint what I am missing or doing wrong in trying to replace the muellmax_de by rsag_de?

PR #6555, opened 2026-06-06
Add kositeast_sk source for KOSIT EAST waste collection (Slovakia)

jan-tdy:
- Parses PDF schedules from kositeast.sk/obyvatelia/harmonogram-zberu-odpadu-v-obciach/
- Uses pdfminer to detect color-coded collection markers on the calendar PDF
- Supports all municipalities listed on the website via 'town' parameter
- Maps 6 waste types to canonical Icons enum
- Includes TEST_CASES for Adidovce and Andrejova

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/kositeast_sk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/kositeast_sk.py
@@ -3,7 +3,7 @@ import re
 import requests
 from bs4 import BeautifulSoup
 from pdfminer.high_level import extract_pages
-from pdfminer.layout import LTTextContainer, LTTextLine, LTRect
+from pdfminer.layout import LTTextContainer, LTTextLine, LTRect, LTPage
 from datetime import date
 import logging
 from waste_collection_schedule import Collection, Icons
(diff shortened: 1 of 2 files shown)


markvp (changes requested):
Thanks for tackling KOSIT EAST and for adding the doc file and TEST_CASES — the structural side (COUNTRY, Icons enum, translations) is all in good shape.

Before we can merge, a couple of things need rethinking on the PDF parsing:

1. `year = 2024` fallback — if the "ROK YYYY" header isn't matched, every date silently lands in 2024. Please derive the year reliably and raise/skip rather than defaulting to a hardcoded year.
2. Layout-dependent magic numbers — the column pitch (86.6), x-offset (40), the y > 485 half-year split, the rect-size window, and the hardcoded RGB COLOR_MAP are all tied to one exact PDF template, and only pages[0] is parsed. For a source meant to cover every municipality, a template or scaling change (or a multi-page PDF) will silently produce wrong or empty dates with no error. Could you make the parsing more robust to layout variation, or confirm with test output across several different towns that the same offsets hold?

If the colour/coordinate approach is genuinely the only way the provider exposes this, let's discuss — but as written it's quite fragile. Happy to take another look once these are addressed.

jan-tdy:
I reached my Antigravity quota mid-job, sorry. Either someone else will do
it, or i will do it next weekend.

jan-tdy:
@gemini-code-assist can you look at it

jan-tdy:
@Claude can you look at it

jan-tdy:
@markvp It is now ready

markvp (approved):
Thanks for the rework, Jan — this is a big improvement. The hardcoded 2024 fallback is gone (it now derives the year from the "ROK" header and raises if it's missing), the magic offsets are replaced with data-driven clustering, and every page is parsed now. I pushed a tiny formatting pass (black/isort) on top. Looks good to merge.

PR #6557, opened 2026-06-06
CONTRIBUTING.md is at wrong place

jan-tdy:
It is not that easy to find until you read readme.md, it is annoying because I know it is here, but only now I realized it is in some folder.... And [Community page of that one repo](https://github.com/mampfes/hacs_waste_collection_schedule/community ) does not even know about it.


markvp:
Thanks Jan! Rather than copying the file, I've moved it to the root and updated all the internal links — same outcome for GitHub's community features, but only one file to maintain. Also took the opportunity to add CODE_OF_CONDUCT.md, SECURITY.md, and a PR template while we were at it. Merging now.

Issue #6561, opened 2026-06-07
RFC: Reduce source boilerplate — BaseSource pipeline, canonical WasteTypes, typed PARAMS


markvp:
## Summary

This is a request for comment on a multi-phase plan to **significantly reduce the repetition and boilerplate** involved in writing and maintaining source modules. The goal is to make a new source as small as possible, around 15 to 40 lines of source-specific logic, down from today's 60 to 150.

The plan builds on the `arch/base-source-waste-types` branch (draft PR #6562). This issue describes the current design and proposes how to extend it into a full migration strategy. The change history of the branch is in the progress-update comment below; this description tracks the current state.

Tagging @mampfes @5ila5 @dt215git @bbr111 for input. You have all contributed heavily to the source ecosystem and your feedback on feasibility and priorities would be really valuable.

---

## The pipeline: retrieve, parse, preprocess, transform

The core idea is four composable, swappable steps:

```
retrieve → parse → preprocess → transform
```

`preprocess` (usually the default) normalises the parsed output into the records the transformer consumes, and is where recurring schedules are expanded into dates.

Sources declare which standard steps to use. For most sources the only source-specific code is `__init__`. Everything else is declarative:

```python
@final
class Source(BaseSource):
    TITLE = "Example Council"
    DESCRIPTION = "Waste collections for Example Council."
    URL = "https://example.com"
    COUNTRY = "au"
    API_URL = "https://api.example.com/collections"
    PARAMS = [config_params.uprn()]

    transform = JsonTransformer(
        date_key="collectionDate",
        type_key="binType",
        type_value_map={"refuse": GENERAL_WASTE, "recycling": RECYCLABLES},
    )

    def __init__(self, uprn: str | None = None):
        super().__init__(uprn=uprn)
```

No `fetch()`. No `ICON_MAP`. No `PARAM_TRANSLATIONS`. No `Collection` import. `WASTE_TYPES` is auto-derived from the transformer, and `classify()` remains as an escape hatch for sources too complex for a standard transformer. Every new-arch `Source` is `@typing.final` so the type checker fully verifies its `BaseSource` contract.

---

## What the branch implements

The branch is on green, deterministic (offline) CI and provides the following.

### Retrievers (`retrievers.py`)

`curl_cffi` (Chrome impersonation) is the **default**; it works on regular and Cloudflare-protected sites alike and is already a project dependency.

| Retriever | Backend | When to use |
|---|---|---|
[...]

PR #6562, opened 2026-06-07
RFC: BaseSource pipeline, canonical WasteTypes, typed PARAMS (arch prototype)

markvp:
## What this is

This PR carries the `arch/base-source-waste-types` branch for the RFC in #6561. It is feature-complete with green, deterministic offline CI. Per #6561, merging it to `master` is the planned next step (legacy sources keep working unchanged, no flag day), pending maintainer alignment on the design. See #6561 for the proposal and discussion; the change history for this branch is in the comments below, and this description is a snapshot of the current state.

---

## The architecture

A source is a composition of four typed, swappable pipeline steps:

```
retrieve → parse → preprocess → transform
```

Sources declare *which* standard steps to use. For most sources the only source-specific code is `__init__`; everything else is declarative class attributes.

```python
@final
class Source(BaseSource):
    TITLE = "Example Council"
    DESCRIPTION = "Waste collections for Example Council."
    URL = "https://example.com"
    COUNTRY = "au"
    API_URL = "https://api.example.com/collections"
    PARAMS = [config_params.uprn()]

    transform = JsonTransformer(
        date_key="collectionDate",
        type_key="binType",
        type_value_map={"refuse": GENERAL_WASTE, "recycling": RECYCLABLES},
    )

    def __init__(self, uprn: str | None = None):
        super().__init__(uprn=uprn)
```

No `fetch()`. No `ICON_MAP`. No `PARAM_TRANSLATIONS`. No `Collection` import. `classify()` is still available as an escape hatch for sources too complex for a standard transformer. `WASTE_TYPES` is auto-derived from the transform, and every new-arch `Source` is `@typing.final` so the type checker fully verifies its `BaseSource` contract.

---

## What the branch provides

**Retrievers** (`retrievers.py`): `curl_cffi` (Chrome impersonation) is the default and clears most Cloudflare blocks. `http_get` / `http_post`, configured `HttpGetRetriever`/`HttpPostRetriever`, `TwoStepRetriever` (a lookup call feeds the schedule fetch, with optional `headers=` such as `Accept: application/json` on both calls), and `Legacy*` plain-requests fallbacks for the rare documented case. All curl_cffi retrievers issue their request through the source's single shared session, so connection, cookies and TLS are reused and no source constructs its own session.

**Parsers** (`parsers.py`): `JsonParser(*keys)`, `HtmlParser(selector, skip=, from_json_key=)` (the last for HTML rendered inside a JSON field, e.g. the OCAPI `wasteservices` endpoint), `IcsParser`, `IcsEventsParser`, `TextParser`, `PdfTextParser`, `XmlParser(namespaces=)`, `CsvParser`. Each accepts an optional, intent-named validation argument that raises `ResponseShapeError` (response logged) on a mismatch: `shape=` (a `TypedDict`/`list[...]`, structural) for `JsonParser`; `require=[...]` (required selectors/columns) for `HtmlParser`/`CsvParser`; and `min_events=`/`min_nodes=`/`min_chars=` (count thresholds) for the ICS/XML/PDF/text parsers.

**Date parsers** (`date_parsers.py`): `auto` (dateutil), `for_format(fmt)`, and `from_epoch(unit="s"|"ms")` for Unix-timestamp APIs.

**Preprocessors and recurrence** (`preprocessors.py`, `recurrence.py`): an optional `preprocess` step between parse and transform (`RecurrenceExpander`, `Schedule`, `Compose`, `HolidayShift`) plus a `recurrence` core whose weekday/month names are sourced from Babel's CLDR data across many locales (including inflected forms such as Polish genitive months), so worded dates resolve without hand-maintained tables; `recurring`, `recurring_from_anchor`, `recurring_within`. Projects a weekday-plus-cadence into dates, including seasonal/windowed and financial-year calendars, with no per-source date arithmetic.

**Transformers** (`transformers.py`): `JsonTransformer`, `KeyValueTransformer`, `ICSTransformer`, `HtmlTransformer`. Each takes an optional `parse_date` and `type_value_map`; a map value may be a single `WasteType` or a `list[WasteType]` for a combined round covering several types on one date.

**Canonical waste types** (`waste_types.py`): 11 `WasteType` dataclasses (id, icon, colour, names in en/de/fr/it). Resolution is multilingual and lossless: an unmapped label resolves across languages and is otherwise preserved verbatim, never collapsed to `OTHER`. The displayed label follows the Home Assistant UI language (English fallback), so a French instance shows `Ordures ménagères`. The type carries the icon, so pipeline sources declare no `ICON_MAP`.

**Regions** (`regions.py`): a source is one structure (pipeline + `PARAMS`) applied to one or more regions. `REGIONS` is a typed `list[Region]` (`region(title, **params)`, optional `url`/`country` overrides), each becoming its own README / `sources.json` listing with its params pre-filled; it may be a callable for large external registries. The legacy `EXTRA_INFO` dict list is adapted into `Region`s at the boundary.

**Typed config params** (`config_params.py`): `coords`, `uprn`, `postcode`, `address`, `municipality`, `dropdown`, `dependent_select`, `multi_value_lookup`, `text_field`, `api_key`, `alternatives`. `text_field`/`api_key` take `default=` (optional + pre-filled, e.g. an embedded public key) and `text_field`/`dropdown` take `optional=True` (optional, no default); `alternatives(*groups)` declares mutually-exclusive input groups (validation requires exactly one full group); `dependent_select` drives a cascading dropdown via the source's `get_choices(parent)` / optional `get_parent_choices()`. One `PARAMS` list replaces the three per-source translation dicts and drives both the config-flow UI (its `labels`/`descriptions` feed the generated translations) and up-front validation.

**Response-shape validation** (`response_shape.py`): one shape declaration drives pyright typing, offline-fixture validation, and a runtime `ResponseShapeError` when a provider changes its API. Applied across all new-arch sources, once per shared service envelope (ArcGis, IntraMaps, Abfallnavi) plus per-source for the standard parsers.

**Image-only PDF calendars** (`service/PdfImageCalendar.py`): `PdfImageRetriever` + `ColourGridCalendarParser` read colour-coded calendar images without OCR (Pillow imported lazily). Reference source `amarsul_pt`.

**Shared lookups** (`lookups.py`): `resolve()` does normalised name-to-value matching with suggestions-on-miss (street/town to weekday), removing per-source boilerplate.

**Shared-platform services** (`service/`): the reused platforms are now componentised as retriever + parser + preprocessor sets, so every source on each converts declaratively: `AchieveForms` (auth + `LookupStep`/`GetStep` + field-map/dynamic-row preprocessors), `RiSKommunalAT` (multi-ICS-feed retriever + `lookahead_days` + `paginate_list`), `IntraMaps` (`IntegrationClientRetriever` + `nominatim_reproject`), `SiteparkIES`, `ArcGis`, and the ICS `AthosWasteManagementRetriever` / `PollingIcsRetriever`. New config-param widgets `integer()` / `boolean()` / `raw_object()` (Number / Boolean / Object selectors) back the generic `ics` engine, and `recurrence` gained `monthly_nth_weekday` + `us_federal_holidays`.

**Deterministic offline tests** (`tests/cassette.py`): gating CI records/replays HTTP at the `Session.request` layer for both stacks, clock frozen to the recording date, so no gating test hits a live provider. Live tests are marked `@pytest.mark.live` and excluded from the gate. 74 cassettes across 30 sources (including a dependent-select `get_choices` replay).

**Backward compatibility** (`collection.py`): legacy `Collection(date, t, icon=...)` keeps working via an adapter; `BaseSource` is purely opt-in and all existing sources are unchanged.

---

## Sources migrated

Around 255 of ~900 sources are now on the pipeline (~28.5%, up from ~3.8% when this branch opened). Every reusable/shared-platform service is componentised and its sources converted:

- **ICS**: all `.py` sources, and the generic `ics` engine itself (`alternatives(url|file)` + `raw_object` params/headers + a source-defined `retrieve`); the 178 `doc/ics/yaml` providers fold into `ics.REGIONS`. Added `AthosWasteManagementRetriever` (~15 councils) and `PollingIcsRetriever` (UK htmx portals).
- **AchieveForms** (20 councils), **RiSKommunalAT** (33), **IntraMaps** (14), **SiteparkIES** (10), **ArcGis** (28).
- Smaller shared services: junker, Pozi, WhatBinDay, uk_cloud9_apps, A_region_ch, Sepan, FirmstepSelfService, EcoHarmonogramPL.

Only `CitiesAppsCom` stays legacy by design (login-required). The ~600 single-provider sources are unchanged, and the legacy `fetch()` contract remains permanently supported.

The demonstrators from the original consistency pass remain (recurrence and seasonal, image-PDF, text-PDF, XML, CSV multi-type, cascading `dependent_select`, `alternatives` input, worded Polish dates, epoch dates, OCAPI two-step, `holidays` calendars, and the shared-service split reference `abfall_neunkirchen_siegerland_de`).

New dependencies: `Babel` (recurrence locale data) and `holidays` (regional holiday calendars), alongside `Pillow` (image-PDF).

---

## Design-review consistency pass

Following a full review of the branch against this RFC, a consistency pass landed (CI green, offline). It removed the divergence the review found rather than adding features, so the documented pipeline is the only pipeline:

- retrievers now share one session;
- the pipeline step attributes are uniform verbs (`preprocess`, `transform`);
- typed `PARAMS` labels now actually drive the generated config-flow translations;
- parser validation arguments are intent-named (`require=` / `min_events=` / `min_nodes=` / `min_chars=`, with `shape=` reserved for the structural JSON check);
- `Collection` identity is locale-independent (keyed on `WasteType.id`);
- four unused preprocessors were removed;
- `ArcGis.geocode` returns the point enriched with attributes, so a source needing a city/zone no longer hand-rolls geocoding, and the multi-layer retriever raises when every layer fails instead of returning an empty schedule;
- the converted sources were moved onto the canonical idioms (no `strptime`, no raw `requests`, `EXTRA_INFO` → `REGIONS`, no pass-through `classify()`).

The contributor guide gained an "Anti-patterns" table so new sources do not reintroduce the old habits.

Contributor tooling is now uniform across both styles: `test_sources.py` reads `TITLE`/`TEST_CASES` off a `BaseSource` class as well as the module, and codeowners use one name (`SOURCE_CODEOWNERS`) for legacy and pipeline sources. Migrating a source built on a shared service is documented as a re-implementation (split the service into a retriever + parser first, so each source on the platform converts declaratively); `abfall_neunkirchen_siegerland_de` on the new `SiteparkIESRetriever` is the reference.

The `transformer`/`preprocessor` → `transform`/`preprocess` rename and the parser-argument rename are breaking for the converted sources, all updated in the same change; since nothing is merged there is no migration burden.

---

## Status

Gating CI is green and deterministic (2935 passed, offline); pyright is clean across all pipeline sources; every converted source's provider/region counts are preserved (independently verified); and `update_docu_links.py` runs end-to-end. The one outstanding pre-merge item is a live smoke-test of ~10 sources that could only be verified structurally here (a few providers were unreachable at conversion time). Per #6561, merging this to `master` is the next step; the remaining ~600 single-provider sources migrate incrementally and opt-in afterwards.

---

## Versioning (resolves RFC discussion #6622)

This PR also enacts the versioning policy from the RFC in discussion #6622, adopting **strict semantic versioning (Option A)**. The BaseSource pipeline is what makes it workable: canonical `WasteType`s, typed `PARAMS` and response-shape validation turn "breaking vs non-breaking" into a precise, machine-checkable distinction.

Added:

- `doc/versioning.md`: the policy. Patch = fixes; minor = additions (new sources, platforms, params, countries) and new deprecations; major = breaking changes. A pipeline source migration changes waste-type labels, so it is breaking and is batched into a major, which is why this change ships as the next major (released first as an `X.0.0-alpha`/`beta`). Includes the deprecation lifecycle and the Removed/Deprecated/Added/Fixed/Other release-note ordering (keeping Option B's deprecation-visibility benefit without loosening the scheme).
- `DEPRECATIONS.md`: a tracking file so compatibility shims do not silently accumulate.
- The same policy is mirrored into `CLAUDE.md`, `AGENTS.md`, the contributing guide, and the release/source agents.

(#6622 is a Discussion, so it cannot be auto-closed by a PR keyword; it will be marked resolved manually.)


5ila5:
Sorry for my long time away.

## General notes
I generally like the idea of having these simple constructs we can just connect together.

Maye we could even put them into a source for advanced users that want to test some configuration before it is officially part of any release but this will reach its limit for complicated multi step sources where we would probably need to the fetch logic ourself but I guess thats fine.


## Some more detailed notes

I think it would be good if we could have more types and eforcement of correct usage of parsers/transformers ...

I thougt we could even remove the __init__ function all together by moving all confiuguration logic into the retriever / transformer and parser


I did some testing but it hits some limitation such as error handling / rasing good exceptions.

If you want you can have a look  at https://github.com/5ila5/hacs_waste_collection_schedule/commit/f2032ae14fb3a7a94d95366a71320f4b79f3bfcf maybe there is something we can acutally use maybe not

markvp:
Great to have you back, no apologies needed! And thanks for not just commenting but actually prototyping it — `f2032ae` is a really useful push on this.

The big things in your branch I've folded in:

- **Threading the types through (`BaseSource[ParserType, TransformerType]` + a matching `Preprocessor`).** This is exactly the "enforce correct parser/transformer usage" you wanted — wiring incompatible stages together is now a type error instead of a runtime surprise.
- **The `preprocessors` layer.** Much nicer than the `isinstance(records, dict)` special-case I had buried in `fetch()`. `Json`/`KeyValue`/`Identity` (plus a `DefaultPreprocessor` that reproduces the old normalisation) as explicit, swappable steps.
- **Config as resolvable callables in the retriever** (`HttpGetRetriever(url=lambda uprn: …)` reading `source.params`), with zero-config `http_get`/`http_post` defaults so sources that only set `API_URL` keep working.

On dropping `__init__` entirely: I went with the *spirit* but stopped just short, because your sticking point — error handling and good exceptions — is the reason. `BaseSource.__init__` now validates the declared `PARAMS` and raises the typed `SourceArgument*` exceptions up front (`config_params.validate()`), so retrievers/transformers can assume clean params and the common source needs no custom `__init__`. Optional and mutually-exclusive params use `required=False` with the cross-field check staying in the source (see `stirling_wa_gov_au` for the address-OR-coords case). That closes the gap you hit.

The "expose the building blocks as a configurable source for advanced users to try things pre-release" idea is lovely — agreed it breaks down for complex multi-step fetches, but for the single retrieve→parse→transform majority it could be powerful. Keeping it on the list as a follow-up once the core stabilises.

I've pushed the reworked version (`6d35560`): every pipeline stage is now a typed callable (`Parser`/`DateParser`/`RetrieverFunc`/`Preprocessor`/`BaseTransformer`), the generics enforce parser/transformer compatibility, and the example sources + the new-architecture test suite are migrated to the new API (158 deterministic tests green). Your prototype is credited in the commit. For the imperative case (compute-then-fetch), a source can still override `retrieve(self, source)` as a method and delegate to `http_get` — see `stirling_wa_gov_au`, which resolves the `headers=… # TODO` you ran into. Have a look when you get a chance and let me know if the direction feels right.

5ila5:
I think this looks a lot better

I noticed that type check will only find potential issues in the source when marking them with `typing.final` as we never instantiate them directly, so I think we should probably mark them as final

markvp:
## Change history

Kept here so the PR description stays a clean current-state snapshot.

**Original prototype.** BaseSource pipeline (retrieve, parse, preprocess, transform), five prototype conversions (one per pipeline shape), canonical WasteTypes, typed PARAMS, and a 170-test suite.

**Services as components, lossless multilingual types, recurrence (2026-06-20, `db7741e0`).** Reusable platforms split into a raw `retrieve` plus a `parse` step (ArcGis, RiSKommunalAT, AbfallnaviDe, IntraMaps). `WasteType` gained per-language aliases with lossless `resolve()`/`preserved()`; added HAZARDOUS and ELECTRONICS (11 canonical types). Added the `recurrence` core and the preprocess step. Around 22 sources converted; pyright scoped via `pyrightconfig.json`.

**Image-PDF calendars, seasonal recurrence, more sources (2026-06-21).** Folded in the former PoC #6608: `service/PdfImageCalendar.py` (`PdfImageRetriever` + `ColourGridCalendarParser`, Pillow lazy), region-of-interest `crop` and financial-year support, `recurrence.recurring_within` + `Schedule.not_before`/`until`, the shared `lookups.resolve`, and sources `amarsul_pt`, `lawrence_ma_us`, `isaac_qld_gov_au`, `hohokus_nj_us`.

**Deterministic offline tests + response-shape validation (2026-06-21).** `tests/cassette.py` records/replays HTTP at the session layer for both stacks, clock frozen to the recording date; live tests excluded from the gate via `-m "not live"`. Typed response shapes drive pyright, fixtures, and a runtime `ResponseShapeError`. `@final` on every new-arch Source.

**Toolkit expansion: two-step retrieve, PDF/XML/CSV parsers, multi-type collections (2026-06-21).** `TwoStepRetriever` (lookup then fetch); `PdfTextParser`, `XmlParser`, `CsvParser`; a `type_value_map` value may be a `list[WasteType]` (combined rounds); `config_params.api_key()`. Example migrations, each with an offline cassette and a declared response shape: `stkh_hu` (text-PDF), `cannock_chase_dc_gov_uk` (XML), `tibi_be` (CSV, multi-type), `leeds_gov_uk`, `gemeinde24_at`, `muttenz_ch`.

**Alternatives, param defaults, dependent_select choices, locale dates (2026-06-21).** `config_params.alternatives()` for mutually-exclusive input groups (`reading_gov_uk`: UPRN or postcode plus house); `text_field`/`api_key` `default=` (`leeds_gov_uk`: embedded public key); the `dependent_select` parent-options contract (`get_parent_choices`) plus a config-flow consumer and an offline `get_choices` replay harness (`gemeinde24_at`); Polish month and weekday recurrence vocab with the `zys_harmonogram_pl` migration. Migration tooling (an AST converter) was considered and dropped in favour of manual migration with the template and the updated `/new-source` agent.

markvp:
## Change history (continued): design validation, regions, localisation

Validated the design by migrating five randomly-picked legacy sources (one per data shape). All proved expressible; the exercise surfaced and filled several reusable gaps, and prompted two design decisions.

**Toolkit gaps found and filled.** `date_parsers.from_epoch(unit="s"|"ms")` (Unix-timestamp APIs); `HtmlParser(from_json_key=...)` for HTML rendered inside a JSON field (the OCAPI `wasteservices` endpoint many AU councils share); `TwoStepRetriever(headers=...)` (curl_cffi content-negotiated to XML without an `Accept` header); `text_field`/`dropdown` `optional=True` (an optional field with no default, replacing `dataclasses.replace(..., required=False)`); `alternatives()` (mutually-exclusive input groups) now also used by `stirling_wa_gov_au`.

**Locale data via Babel + holidays.** `recurrence` weekday/month names now come from Babel's CLDR data across many locales (manual tables retired; inflected forms like Polish genitive resolve), and the `holidays` library supplies regional holiday calendars (mulhouse, Alsace-Moselle). Both added to `manifest.json` / `requirements.txt`.

**A source is one structure covering many regions.** Introduced a typed `Region` (`regions.region(title, **params)`) and `BaseSource.REGIONS`: one structure (pipeline + `PARAMS`) applied to one or more regions, each a discoverable README / `sources.json` listing. The toolchain processes `Region`s natively; the legacy `EXTRA_INFO` dict list is adapted into `Region`s at the boundary (a thin shim, not a second native format). The shared service platforms and the platform YAML registries are the same pattern at scale and remain a deliberate follow-on.

**Waste-type labels now localise.** `Collection.type` follows the Home Assistant UI language (English fallback), so a source migrated to canonical `WasteType`s keeps a sensible local label (a French instance shows `Ordures ménagères`, not `General Waste`). Note that converting a legacy source still changes its labels from the author's hard-coded strings to the canonical names: config arguments are unchanged, but type-based filters/customisation need updating, to be called out per source at migration time.

**Five migrations:** `mrcjoliette_qc_ca`, `yarra_ranges_vic_gov_au`, `lancaster_gov_uk`, `mulhouse_alsace_fr`, `phila_gov` (the last is live-only: `api.phila.gov` 202-gates automated access from CI). Cassettes recorded for the four reachable ones; the contributor guide, `CLAUDE.md`, the source-implementer agent, and `/new-source` are updated; gating suite green.

markvp:
@bbr111 this branch will get stale very quickly - could you try pulling it and migrating a few sample sources to see what the experience is like / validate whether we can/should go down this path? if you're happy with it, it would be good to switch over sooner rather than later.

bbr111:
@markvp thanks for your reminder. will do some investigations shortly.

bbr111:
Did a real conversion trial against this branch: `abfall_neunkirchen_siegerland_de.py` (uses the shared `SiteparkIES` service — ~11 sources depend on it), full pipeline + `ICSTransformer`, verified byte-for-byte parity against the legacy `fetch()` (45/45 entries, both `TEST_CASES`). A few things worth flagging from actually doing it:

**1. Display language is correctly locale-driven, but easy to misread in ad-hoc testing.** `Collection.type` resolves through `waste_types.display_name()`, which reads a module-global `_display_language` set once via `set_display_language(hass.config.language)` in `init_yaml.py`/`init_ui.py`. That's correct inside a real HA instance, but instantiating `Source(...).fetch()` directly outside HA (any manual smoke test) silently falls back to English with no signal that anything was skipped. Cost me a moment to realize the "wrong" language wasn't a bug — just an uninitialized global. Might be worth a one-line callout in `contributing_source.md`'s testing section so the next person doesn't chase a ghost.

**2. `retrieve()` duplicates shared-service logic unless the service is split first.** My first pass at converting a source built on `SiteparkIES` — whose `fetch_ics()` bundles the HTTP GET *and* the ICS parse into one call — had to reimplement the request params (`ModID`, `call=ical`, `kat`, `alarm`, `Referer`) by hand inside `retrieve()`, since there was no raw-response entry point to delegate to. Net effect: the file grew from 58 lines (legacy) to 105 (pipeline) — the opposite of what the architecture promises. Splitting `fetch_ics()` into `fetch_ics_response()` (raw) + `fetch_ics()` (parses the former) let `retrieve()` delegate cleanly, landing at 89. I'd guess most of the other shared service clients have the same "retrieve+parse in one call" shape, so any source built on top of them will hit this unless the service is pre-split. Might be worth flagging as a prerequisite step in the conversion guide, or auditing the shared services up front rather than discovering it source-by-source.

**3. Should `test_sources.py` learn about `BaseSource`, or should there be a separate tool?** Right now `python test_sources.py -s <name> -l` hard-asserts module-level `TITLE` etc., so it fails immediately on any pipeline source — reproduces identically on already-merged `kwinana_wa_gov_au.py`, so it's not specific to my conversion, it's a gap for every pipeline source today. My inclination is to extend the existing tool to detect a `BaseSource` subclass and read `TITLE`/`TEST_CASES` off the class instead of the module — one tool, one contributor-facing entry point. The alternative is a parallel `test_basesource.py`, which avoids touching the legacy code path but means contributors need to know upfront which tool applies to which source style. Happy to put together the small patch either way — which do you want in the branch?

**4. `SOURCE_CODEOWNERS` vs `CODEOWNERS` — keep two names, or unify?** Noticed `update_docu_links.py` already reads both (`getattr(source_cls, "CODEOWNERS", None) or getattr(module, "SOURCE_CODEOWNERS", [])`), so legacy sources keep `SOURCE_CODEOWNERS` and pipeline sources introduce a differently-named `CODEOWNERS` for the same concept. That's a deliberate fallback, not an oversight, but it does mean two spellings of one idea live side by side going forward, and every place that reads codeowners needs to know to check both. Worth a call: keep it split (legacy stays as-is, only new architecture uses the shorter name), or converge on one name — e.g. have pipeline sources also accept `SOURCE_CODEOWNERS` (drop the new alias), or the other direction (rename the legacy module-level var too, a mechanical ~600-file search-and-replace with no behavior change). No strong opinion from me, just flagging that it's currently a fork in naming rather than a decided convention.

markvp:
@bbr111 this is exactly the hands-on validation the RFC needed, thank you. A real byte-for-byte conversion (and catching the SiteparkIES fetch+parse bundling) surfaced things no amount of reading would have. All four points were on the mark.

Status on each:

**1, 3 and 4 are done on the branch.**
- Display language: added a note to the testing section of `contributing_source.md` so the next person does not chase the English-fallback ghost when testing outside HA.
- `test_sources.py` now reads `TITLE`/`TEST_CASES` off a `BaseSource` class as well as the module, so `-s <name> -l` works for pipeline sources (verified on `kwinana_wa_gov_au` and the migration below). I went with extending the existing tool, as you suggested: one entry point.
- Codeowners: converged on a single name, `SOURCE_CODEOWNERS`, for both styles (pipeline sources set it as a class attribute), which closes the two-spellings fork without a 600-file rename.

**2 is the interesting one, and I think you have put your finger on something bigger than the one service.**

The 58→105 growth is real, but I read it as a symptom rather than the disease: the conversion wrapped the legacy service through a `retrieve()` override instead of re-implementing the service in the pipeline's own terms. The unit of migration for a shared-platform source is the *service*, not the individual source. Once the platform is expressed as a `Retriever` (HTTP only) plus a `Parser` (bytes to records), every source on it becomes a thin declaration and each conversion shrinks.

To make sure that is not just assertion, I took your exact source and did it that way on the branch:
- Split `SiteparkIES` as you proposed (`fetch_ics_response()` raw, `fetch_ics()` delegating to it, so the ~10 legacy sources keep working) and added a `SiteparkIESRetriever` that runs the pois lookup and returns the raw ICS response.
- Rewrote `abfall_neunkirchen_siegerland_de` as a declarative pipeline: `retrieve = SiteparkIESRetriever(...)`, `parse = parsers.IcsParser()`, `transform = ICSTransformer(...)`. No `retrieve` override, no ICON_MAP, no hand-rolled request params.
- 44/44 entries per test case, dates identical to the legacy fetch, types now canonical WasteTypes (the three Restmüll variants collapse to General Waste). Offline fixture recorded so it replays deterministically. You are credited in the commit.

I also wrote it up so we do not rediscover it per source: a "Migrating a source built on a shared service" section in `contributing_source.md` (with the split-the-service-first recipe), an anti-pattern row for the `retrieve()`-reissues-a-service smell, and the same steer in the source-implementer and pr-reviewer agents. Migration is a re-implementation, not a wrapper, and that is now written down.

Your instinct to split `fetch_ics()` was right; the only change is doing the split in the service so all its sources benefit, rather than in the one source.

You have now actually driven this, which none of the rest of us have yet, so the real question back to you is a confidence one: having done a proper conversion, do you feel this is the right direction to commit to, or is there more we should work through first?

If you have the appetite, it might be worth another go at a source on an already-migrated service (ArcGis, RiSKommunal AT, Abfallnavi, IntraMaps) or an otherwise simple one. That path skips the service-splitting entirely and is the experience most future conversions will have: assign a retriever and a parser, add a type map, done. A good gut-check on whether the declarative case feels as light as it should.

If you think it is close, I would like to line up the merge to master so it stops going stale. If something is still nagging, I would much rather hear it now than after we land it.

bbr111:
@markvp i'm a few days off. will take some time to answer this.

markvp:
### Progress update — 2026-07-12: the migration has been executed on this branch

The incremental source migration described in the plan has now largely been carried out here, rather than left to post-merge community work. Coverage is now **~255 of ~900 sources (~28.5%, up from ~3.8%)**.

**Every reusable/shared-platform service is componentised and its sources converted:**

- **ICS** — all `.py` sources, and the generic `ics` engine itself is now a `BaseSource` (`alternatives(url|file)` + `raw_object` params/headers + a source-defined `retrieve`); the 178 `doc/ics/yaml` providers fold into `ics.REGIONS`. Added shared `AthosWasteManagementRetriever` (the "WasteManagementServlet" wizard, ~15 councils) and `PollingIcsRetriever` (UK htmx portals).
- **AchieveForms** (20 councils; new `AchieveFormsRetriever` + `LookupStep`/`GetStep` + field-map/dynamic-row preprocessors), **RiSKommunalAT** (33), **IntraMaps** (14, + new `IntegrationClientRetriever`), **SiteparkIES** (10), **ArcGis** (28).
- Smaller shared services: junker, Pozi, WhatBinDay, uk_cloud9_apps, A_region_ch, Sepan, FirmstepSelfService, EcoHarmonogramPL.

**Framework extended only where a shape recurred** (never a per-source hack): `config_params.integer()/boolean()/raw_object()` (Number/Boolean/Object selectors); `IcsParser` `offset`/`regex`/`split_at`/`title_template`; `recurrence.monthly_nth_weekday` + `us_federal_holidays`; RiSKommunal `lookahead_days`/`paginate_list`; `Region` `doc_filename`/`howto`/`source_owners` (per-region docs, used by the ICS YAML fold).

**Quality:** gating CI green (2935 passed, offline), pyright clean across all pipeline sources, every converted source's provider/region counts preserved (independently verified), and `update_docu_links.py` runs end-to-end. Several latent legacy bugs were fixed in passing.

**Only `CitiesAppsCom` stays legacy by design** (login-required). The ~600 single-provider sources are unchanged, and the legacy `fetch()` contract remains permanently supported.

**Before merge:** the one outstanding item is a live smoke-test of ~10 sources that could only be verified structurally here (a few providers were unreachable or rate-limited at conversion time). A separate handful currently fail live for pre-existing provider outages, not regressions.

markvp:
@bbr111 would love you to take another look if you can 😊

bbr111:
@markvp 
I’ve taken another look at the current PR, also considering #6819.

From my side, the overall architecture is now in a good state. Most of the original design questions have been addressed through the typed pipeline, preprocessors, response-shape validation, expanded PARAMS and the lossless handling of unknown waste types.

The main remaining architectural point I see is collection metadata.

PR #6819 now exposes `Collection.description` and `Collection.location` on Home Assistant calendar events. The current standard transformers in this PR appear to construct collections from date and waste type only. We should therefore decide how pipeline sources can declaratively preserve metadata such as description and location.

My preference would be optional `description_key` and `location_key` arguments on the relevant transformers, accepting either a field name or a callable in the same way as `date_key` and `type_key`. If one record produces multiple waste types, the metadata could be copied to each resulting Collection.

This is especially relevant to ICS. We should clearly document the difference between `IcsParser` and `IcsEventsParser`, and ensure there is a standard pipeline combination that preserves ICS DESCRIPTION and LOCATION rather than reducing every event to `(date, summary)`. If we want support this.

The other points I think we should verify before merging are:

* whether the new Collection compatibility layer covers all dict-like operations currently used elsewhere in the integration, not only `get()` and item access;
* what the expected error policy is for malformed individual records versus a changed provider response;
* what minimum fixture coverage we require for each migrated standalone source and for shared-service migrations;
* and how we want to release this change, ideally with a clear beta/rollback path because of the number of converted sources.

I don’t see the size of the PR itself as a reason to reject the direction, provided we are comfortable with those points and the remaining live smoke tests.

markvp:
@bbr111 thank you, this is a genuinely useful pass. Good to hear the direction reads as settled. Status on each point:

**Metadata (the main one): done on the branch, and your instinct was right.** All the standard transformers now take optional `location_key`/`description_key` (a field name or a callable, read the same way as `date_key`/`type_key`), copied to each Collection of a combined round; `HtmlTransformer` spells them `location_getter`/`description_getter`. For ICS I went the zero-config route you were pointing at: pairing `IcsEventsParser` with the standard `ICSTransformer` now preserves `LOCATION`/`DESCRIPTION` automatically, so a source only reaches for `classify()` when it needs to *inspect* those fields. Documented the `IcsParser` vs `IcsEventsParser` split in `contributing_source.md`. This slots straight onto #6819.

The other four:

**1. Collection dict-compatibility.** Good catch to make me check properly. The pre-refactor Collection subclassed `dict`, so `dict(c)`, `{**c}`, iteration, `len()` and `.items()/.keys()/.values()` all worked (mostly for user value/date templates). The refactor's shim only covered `get`/`[]`/`in`/`update`, so the rest had regressed. I've restored the read-side Mapping protocol over `as_dict()`. Auditing first-party code found nothing relying on the missing ops, so the exposure was templates only; the sole behaviour not restored is a bare `json.dumps(collection)`, which nothing uses (`as_dict()` is the path).

**2. Error policy.** Today: a malformed *record* is skipped (`skip_unparseable_dates`, or a getter returning None), an unknown *label* is preserved verbatim rather than collapsed to OTHER, and a changed *provider response* raises `ResponseShapeError` via the declared response shape / `min_events`. That is the right split (one bad row shouldn't fail the whole fetch; a shape change should surface loudly). I'll write it up explicitly so it's a stated contract rather than emergent behaviour.

**3. Minimum fixture coverage.** Proposal: every migrated standalone source ships one recorded cassette covering its `TEST_CASES`; a shared-service migration additionally keeps one cassette per distinct response shape the service produces. I'll add that to the contributing guide as the floor.

**4. Release, rollback, and the versioning scheme.** This is where the architecture actually resolves the open question in #6622, so let me put a recommendation on the table: **Option A (strict SemVer)**.

My reasoning is specific to this work. The new architecture is what finally makes "breaking vs non-breaking" precise enough for strict SemVer to mean something: canonical `WasteType`s make a label change a well-defined breaking event, typed `PARAMS` and response-shape validation make config/schema changes explicit, and the componentised services make capability additions clean. The main practical objection to A (that in the old hard-coded-string world you couldn't reliably tell what was breaking) is dissolved by the architecture itself. A is also what HACS and most consumers assume a version string means.

Option B's one genuine advantage is deprecation visibility, and we can keep that under A without adopting B, by taking @marsdeat's release-note ordering (Removals, Deprecations, Additions, Fixes, Other). And @marsdeat's "where does a major rework of a single source fit" question is answered by the deprecation lifecycle: a single-source rework that would change entity IDs goes through a compat shim (non-breaking, so it ships as a minor), and only the eventual batched removal lands in a major. So a per-source change never forces an integration-wide major.

Concretely for this branch under A:
- Migrating a source changes its waste-type *labels* (canonical names), so this is a breaking change and ships as the **next major**, released first as `X.0.0-alpha`/`beta`. The legacy `fetch()` contract stays supported as the rollback path, so any source can revert individually.
- I'll fold this recommendation into #6622 so we can close the policy out and move forward on it.

Everything above except the doc write-ups for 2 and 3 is on the branch now (offline suite green, pyright clean). Does the metadata shape look right to you, and are you happy with the fixture floor in 3?

bbr111:
@markvp 
I reviewed the latest updates again.

The main points I raised previously now appear to be addressed:

* `description` and `location` are supported by the standard transformers through optional keys/callables and are copied to all Collections produced by a combined round.
* ICS metadata is covered through `IcsEventsParser` and `ICSTransformer`.
* The Collection compatibility layer now restores the read-side Mapping behaviour needed by existing templates, including iteration, `dict(collection)`, `items()`, `keys()` and `values()`.
* The strict semantic-versioning policy and the planned major alpha/beta release provide an appropriate rollout path for the breaking waste-type label changes.
* Current CI is green.

From an architecture perspective, I no longer see a major unresolved design issue.

The remaining points are mainly release-readiness checks:

1. document and, where possible, complete the remaining live smoke tests;
2. confirm that nothing still relies on direct `json.dumps(Collection)` behaviour, since Collection is no longer a dict subclass;
3. ensure the metadata tests cover JSON/callable extraction, combined waste types and ICS LOCATION/DESCRIPTION;
4. define the criteria for moving from alpha to beta and then stable;
5. clearly communicate that canonical/localised waste-type labels may affect user automations or templates that compare exact strings.

With those points covered, I would be comfortable proceeding with the planned major alpha/beta release.

markvp:
@bbr111 thanks, appreciate the thorough pass. Agreed these are the right release-readiness gates. Where each stands:

**Already on the branch:**
- **(2) `json.dumps(Collection)`** — audited: nothing in first-party code serialises a Collection directly; `as_dict()` is the only path. Confirmed clean.
- **(3) Metadata tests** — covered in `test_new_architecture.py`: JSON extraction by field key and by callable, waste-type combined rounds (metadata copied to each Collection), and the `IcsEventsParser` → `ICSTransformer` LOCATION/DESCRIPTION carry-through with no extra config.
- **(4) alpha → beta → stable criteria** — good call, these weren't stated. Now added to `doc/versioning.md`: alpha = offline suite green against recorded fixtures + pyright clean + no known label regression; beta = live-verified across the migrated set with labels confirmed; stable = one full release cycle with no rollback to legacy `fetch()` and no open label-regression report. A source that misses its bar reverts to `fetch()` for that release rather than holding back the whole major.
- **(5) Label-change communication** — also in `versioning.md`: every migrating major must lead its release notes (and the source's doc) with a callout that canonical/localised waste-type labels have changed and may affect automations or templates comparing exact strings.

**Release-prep, tracked before the alpha:**
- **(1) Live smoke tests** — the fixture floor is documented; I'll record the outstanding live runs and note any source that can't be live-verified.

That leaves nothing blocking the design. Thanks again for the careful review.

markvp:
@bbr111 how are you feeling about this branch? Ready to alpha release it yet?

bbr111:
yes. ready for alpha release. we need to test it.
@markvp maybe we find some timeslot between our timezones to discuss some points in video or audio call.
could be faster than writing here. If you are interested. Please let me know. if have teams or discord available.

markvp:
Absolutely that would be good. I have those too... Just not sure the best way to exchange contact details securely....

PR #6577, opened 2026-06-08
Docs update for Norfolk councils (UK) and Dartford Council (UK)

marsdeat:
## Summary

Update the documentation for councils in Norfolk, UK; also fix a documentation error for Dartford Council.

### `dartford_gov_uk.md`
Fix the heading on this. Documentation accidentally referenced North Norfolk

### {`great_yarmouth`, `north_norfolk`, `norwich`, and `west_norfolk`} `_gov_uk.md`
Add a note about local government reorganisation and continuing to use the current sources until new sources are available.
This is not a current problem in the INTEGRATION, but there are already people in real life confused as to whether their council has changed, and this will only become more of a problem over the next 2 years. It is best to have this documentation ready to point people towards.
I also made minor wording adjustments to the geographic text at the top of the source files. Current wording was occasionally conflating council and geographic terminology.

### `south_norfolk_and_broadland_gov_uk.md`
Same as with the other Norfolk councils above. In addition, clarified the difference in data level between Broadland and South Norfolk addresses, both for the information of users and for maintainers who may not be familiar with the local context.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [x] Documentation update
- [ ] Other

## Checklist

[Not applicable - docs only update]

### DIFF
--- a/doc/source/breckland_gov_uk.md
+++ b/doc/source/breckland_gov_uk.md
@@ -1,6 +1,6 @@
 # Breckland Council
 
-Support for schedules provided by [Breckland Council](https://www.breckland.gov.uk/mybreckland), serving the district of Breckland, UK.
+Support for schedules provided by [Breckland Council](https://www.breckland.gov.uk/mybreckland), serving the district of Breckland, in Norfolk, UK.
 
 ## Local Government Reorganisation note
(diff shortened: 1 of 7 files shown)


marsdeat:
Changed the wording a little, specific to each area. (South Norfolk and Broadland are being split three ways, so the existing wording is correct for them; the rest are wholly incorporated into a future authority so don't need the additional wording)

Also added the note to Breckland, which got missed on the first pass because I searched for "Norfolk" and the term doesn't show up in those docs. I'm adding "Norfolk" to the leadmatter of both Breckland and Norwich to make it easier to find them by searching in future.

PR #6593, opened 2026-06-09
fix(northumberland_gov_uk): add required postcode step to fix broken CSRF flow

### LINKED ISSUE #6573: [Source Defect]: Northumberland County Council
### Source Name

northumberland_gov_uk

### What's happening?

When trying to add an example address using the UPRN I get the following error (tested with multiple UPRN's) 

"Argument is invalid: "We could not find values for the argument 'uprn' with the value '100110637553', No collection data found. Please check your UPRN is correct."

This unfortunately has not worked for me since Northumberland was implemented.

### Error Logs

```Shell
No error logs are being produced.

Checked home-assistant.log after multiple reboots and updates but no error gets posted to any logs.
```

### Source Configuration

```YAML
(Added via UI) 

waste_collection_schedule:
    sources:
    - name: northumberland_gov_uk
      args:
        uprn: "10091036842"
```

### Checklist

markvp:
## Summary

Fixes #6573 — Northumberland County Council bin collection lookups have been broken since the site tightened its CSRF flow.

- **Root cause:** The source was POSTing directly to `/address-select` using a CSRF token obtained from a GET to `/postcode`. The site now requires an intermediate POST to `/postcode` (with the actual postcode) to establish a session-bound address list and issue a fresh CSRF token before `/address-select` will accept a selection.
- **Fix:** Implement the correct three-step flow: GET `/postcode` → POST `/postcode` (with `postcode` field) → POST `/address-select` (with `address`/UPRN field). Adds `postcode` as a new required parameter.
- **Evidence:** Network trace provided by @marsdeat in #6573 confirms the exact payload expected at each step.
- Updates `doc/source/northumberland_gov_uk.md` to document the new required `postcode` parameter.

## Test plan

- [ ] `python -m pytest tests/test_source_components.py -q` passes (9/9)
- [ ] `python -m black` and `isort` — no changes needed
- [ ] Live test against `uprn=100110637553, postcode="NE46 1UF"` from a UK IP (geo-blocked from outside UK) — please confirm on merge

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/northumberland_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/northumberland_gov_uk.py
@@ -12,7 +12,10 @@ DESCRIPTION = "Source for Northumberland County Council waste collection schedul
 URL = "https://www.northumberland.gov.uk"
 COUNTRY = "uk"
 TEST_CASES = {
-    "30, NE46 1UF (UPRN 100110637553)": {"uprn": "100110637553"},
+    "30, NE46 1UF (UPRN 100110637553)": {
+        "uprn": "100110637553",
+        "postcode": "NE46 1UF",
+    },
 }
 
 ICON_MAP = {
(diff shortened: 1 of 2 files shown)


marsdeat:
I can test this after merge

bbr111:
you can also test it before merge.
just download the files from repo or check-out PR

marsdeat:
Not working

`test_sources.py -l -i -s northumberland_gov_uk`:
```
Testing source northumberland_gov_uk ...
  30, NE46 1UF (UPRN 100110637553) failed: We could not find values for the argument 'uprn' with the value '100110637553', No collection data found. Please check your UPRN and postcode are correct.
```

marsdeat:
Can't do a direct live test as my HA device has decided to pick now to play up and not let me SSH in properly, if you need a direct integration test that will have to be tomorrow at the earliest I think.

markvp:
> Can't do a direct live test as my HA device has decided to pick now to play up and not let me SSH in properly, if you need a direct integration test that will have to be tomorrow at the earliest I think.

Don't stress. Won't be releasing until the weekend

marsdeat:
Nope, not working live either

image

Same error with and without spaces, with + or - instead of space, and upper/lowercase postcode (and all combinations thereof).

markvp:
Thanks @marsdeat for testing this through twice — really helpful. Unfortunately this confirms the merged change doesn't actually fix the lookup: the request flow reaches the site but we're still getting "No collection data found", which means either the post-/postcode step needs an extra address-selection request we're missing, or the results-page parser isn't matching the real markup.

I've reopened #6573 to track this. The blocker is that the site geo-blocks non-UK IPs, so I can't capture the real responses to fix it blind. @marsdeat — when you next get a moment, could you grab the raw HTML (or a HAR) of the three steps from a UK connection: the GET /postcode, the POST to /postcode with your postcode, and the resulting page after selecting your address? With the actual markup in hand I can get the flow and parser right and put up a follow-up PR. Apologies for merging ahead of the verification.

PR #6603, opened 2026-06-10
Add Repentigny qc canada

### LINKED ISSUE #2481: [Source Request]: Repentigny, Quebec, Canada
### Municipality / Region

Repentigny, Quebec, Canada

https://repentigny.ca/sites/default/files/2023-12/ville-repentigny_calendrier_collectes_2024.pdf

### Example Address

Sector A,B,C,D,E,F

### Collection Data Format

As PDF Download

### Additional Information

_No response_

lit-af:
## Summary

New source for Repentigny, Qc, Canada. 

Closes: https://github.com/mampfes/hacs_waste_collection_schedule/issues/2481

## Type of change

- [X] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [X] `python -m black` and `python -m isort --profile black` run on changed source files


markvp (changes requested):
Thanks for putting this together, and for the thorough checklist and TEST_CASES — much appreciated.

Two things before this can go in:

1. The fetch currently returns no collections. Running the live test against all three sectors gives "found 0 entries" each time. The page at repentigny.ca renders its collection calendar client-side from a JSON endpoint (there's a fetch(...).then(r => r.json()) reading data.features in the page), so the table.calendrier / <caption> / p.date HTML the parser looks for isn't present in the static response. Could you find the underlying JSON/GeoJSON endpoint the page calls and parse that directly, mapping the sector argument onto it? That'll make the source robust.

2. A couple of smaller things to fold in while you're there: please raise SourceArgumentNotFound (from waste_collection_schedule.exceptions) instead of a bare ValueError for a missing sector, and reconcile the "Quebec open data portal" wording once the real endpoint is wired up.

One nit I've already noted on my side: the doc file needs to be doc/source/repentigny_ca.md (matching the source id) rather than repentigny.md, otherwise the structural test flags it as orphaned — easy to fix on the next push.

markvp (approved):
Thanks for the quick turnaround on this. Switching to the calendar JSON endpoint did the trick — I'm now getting a full schedule for every sector (60+ entries each), the SourceArgumentNotFound change and the repentigny_ca.md doc filename are both sorted, and the structural tests pass. Approving. Nice work.

PR #6605, opened 2026-06-11
fix(preston_gov_uk): require street for UPRN lookup; clarify parameter docs

### LINKED ISSUE #6337: fix: preston_gov_uk broken - council changed website
## Source broken: Preston City Council (`preston_gov_uk`)

The Preston City Council source (`preston_gov_uk.py`) is broken. The council appears to have changed their website approximately 2 weeks ago (reported ~2026-05-15).

The source currently scrapes `https://selfservice.preston.gov.uk/service/Forms/FindMyNearest.aspx?Service=bins` using a BeautifulSoup/form-POST approach. This endpoint may have been removed or restructured.

**Steps to reproduce:** Configure the integration with any valid Preston address/UPRN — fetch will fail.

**Original report:** https://github.com/mampfes/hacs_waste_collection_schedule/discussions/6335

markvp:
## Summary

Partial follow-up to #6419 — fixes the residual UPRN-only regression reported in #6337 (comment 2026-06-10).

- Add early validation: when `street` is empty, raise `SourceArgumentNotFound` with the `street` argument immediately, rather than sending the UPRN as an address search string and silently failing
- Update `TEST_CASES`: `Test_002` now provides both `street` (postcode `PR1 2RL`) and `uprn`, matching realistic usage
- Add `PARAM_TRANSLATIONS` / `PARAM_DESCRIPTIONS` (en) making it clear that `street` is required and `uprn` is an optional disambiguator
- Add `HOW_TO_GET_ARGUMENTS_DESCRIPTION` with a link to findmyaddress.co.uk for UPRN lookup

**Root cause:** The Preston self-service form's `txtAddress` field requires a postcode or street name. Passing a raw UPRN number here yields no dropdown, so `select_el` is `None` and `SourceArgumentNotFound` is thrown.

**Note:** I could not test this against the live Preston site (it geo-restricts non-UK IPs), so the live TEST_CASES are unverified from my environment. Asking the original reporter to validate from master once merged.

## Test plan

- [ ] `python test_sources.py -s preston_gov_uk -l` passes both TEST_CASES (requires UK network access to selfservice.preston.gov.uk)

Closes #6337 (residual)

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/preston_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/preston_gov_uk.py
@@ -20,7 +20,25 @@ SRV_URL = (
 
 TEST_CASES = {
     "Test_001": {"street": "town hall, lancaster road"},
-    "Test_002": {"street": "", "uprn": "10002220003"},
+    "Test_002": {"street": "PR1 2RL", "uprn": "10002220003"},
+}
+
+PARAM_TRANSLATIONS = {
+    "en": {
+        "street": "Street / Postcode",
+        "uprn": "UPRN (optional)",
+    }
+}
+
+PARAM_DESCRIPTIONS = {
+    "en": {
+        "street": "Your street name or postcode (required). Used to search for your address.",
+        "uprn": "Your Unique Property Reference Number (optional). When provided alongside street/postcode, it picks the exact matching property from the search results. Leave blank if you are unsure.",
+    }
+}
+
+HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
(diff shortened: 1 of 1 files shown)


westwoodx:
Thanks for the very prompt response and explanation.  All good now when I just enter a post code.
With the mods you describe am I right in thinking the Street field should now make it clear that it's a mandatary field?  Only I don't see any prompt for that yet after downloading 2.27 on HACS.  Should I try again at a later time?

The fact it's now working for me is great thanks, if only you could get the council to actually turn up to empty the bins on every occasion too!!!

markvp:
It will be in the next release

PR #6609, opened 2026-06-11
fix(northumberland_gov_uk): correct table column mapping and add yearless date parsing

### LINKED ISSUE #6573: [Source Defect]: Northumberland County Council

markvp:
Fixes #6573

## Root cause

The schedule results page at `bincollection.northumberland.gov.uk/address-select` renders
collections in a three-column table (`date | day-of-week | bin type`) with dates in the
format `"15 June"` (no year). The previous fix (#6593) had two bugs in `_parse_schedule`:

1. **Pattern A** looked for `class=re.compile(r"\bcard\b")` — the actual div class is
   `ncc-bin-calendar`. Zero cards matched, so Pattern A was always skipped.
2. **Pattern B** read `cells[0]` as bin type and `cells[1]` as date — the columns are
   actually `date / day / bin-type` (indices 0 / 1 / 2). Dates in this table also have no
   year, so `_parse_date` returned `None` for every row regardless.

Both bugs together meant `_parse_schedule` always raised `SourceArgumentNotFound`
("No collection data found").

## Changes

- `_parse_date`: add yearless formats (`%d %B`, `%d %b`) using year 2000 as a safe
  placeholder to avoid a Python 3.15 `DeprecationWarning`, then replace with the current
  year (or next year if the date is already in the past).
- `_parse_schedule` Pattern A (now primary, table): fix column mapping to `cells[0]` = date,
  `cells[2]` = bin type; skip the header row; update guard from `< 2` to `< 3`.
- `_parse_schedule` Pattern B (now fallback, ncc-bin-calendar divs): match
  `ncc-bin-calendar` class, extract bin type from `p[0]` and date from `p[2]`
  (which includes the year in this element).

Confirmed against marsdeat's fresh HAR (NE46 1UF, 11 June 2026): 6 entries returned.

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/northumberland_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/northumberland_gov_uk.py
@@ -1,6 +1,6 @@
 import logging
 import re
-from datetime import datetime
+from datetime import date, datetime
 
 from bs4 import BeautifulSoup
 from curl_cffi import requests
@@ -74,7 +74,7 @@ class Source:
             )
         csrf_token1 = csrf_input.get("value", "")
 
-        # Step 2: POST postcode to /postcode — this returns the address-select page
+        # Step 2: POST postcode to /postcode — this redirects to /address-select,
         # which contains a fresh CSRF token for the next step.
         r2 = session.post(
             f"{BASE_URL}/postcode",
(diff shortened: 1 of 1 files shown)


marsdeat:
Will test tomorrow, am now shut down for the day

marsdeat:
Tested working. 

Image

Matches site output for those bins.

PR #6614, opened 2026-06-11
chore(muellmax_de): deprecate RSAG Rhein-Sieg-Kreis entry in favour of rsag_de

markvp:
Follow-up to #6553 (RSAG via `rsag_de`).

@David-Development pointed out that the new `rsag_de` source overlaps the existing Müllmax `Rsa` listing. Investigation confirmed they hit the **same backend** — RSAG uses Müllmax as its portal vendor and `rsag.de/api` is a JSON/ICS front-end over the same dataset (identical dates and waste types across the same 19 Rhein-Sieg-Kreis municipalities).

To avoid two listings for one provider, this removes the `RSAG Rhein-Sieg-Kreis` (`service_id: "Rsa"`) entry from `MuellmaxDe.SERVICE_MAP` and the corresponding commented-out test case in `muellmax_de.py`. Users are steered to `rsag_de`, which has a friendlier `city`/`street` config.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/MuellmaxDe.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/MuellmaxDe.py
@@ -7,11 +7,9 @@ SERVICE_MAP = [
         "url": "https://www.awista.de/",
         "service_id": "Dus",
     },
-    {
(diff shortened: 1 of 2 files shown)


David-Development:
Terrific, thank you so much! :)

PR #6617, opened 2026-06-11
fix republicservices_com: weekly recurrence projection never fired ('W' vs 'week'); bulk waste icon

jasonjhofmann:
## Summary

Two fixes to the `republicservices_com` source:

**1. Weekly recurrence projection never fired (`'W' != 'week'`).** The Republic Services `publicPickup` API returns `numberOfPickupsPeriodUnit: "W"`, but the projection condition compared it against the full word:

```python
period_unit.lower() == "week"   # API sends "W" → always False
```

so the recurring-schedule projection added in the period-fields change was dead code — every weekly container fell through to the `else` branch and emitted only its single `nextServiceDays` date. Users saw exactly one upcoming pickup per cart instead of the projected ~6 months. Fixed by accepting both forms: `period_unit.lower() in ("w", "week")`.

Live-test evidence (`test_sources.py -s republicservices_com`), entries per test case before → after:

| Test case | Before | After |
|---|---|---|
| Scott Country Clerk | 1 | 26 |
| Branch County Clerk | 1 | 26 |
| Residential Collection | 1 | 26 |
| 4420 Oasis Hill Ave, North Las Vegas | bulk dates only | 65 |

(Non-weekly containers, e.g. Las Vegas bulk service, list explicit future dates in `nextServiceDays` and were unaffected.)

**2. Method-2 "Bulk Waste" used the yard-waste icon.** `mdi:leaf` was a copy-paste from the YARD/ORGANICS branch directly above. Changed to `mdi:sofa`, matching the canonical `Icons.BULKY` value used across the project.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (9 passed)
- [x] `python -m black` and `python -m isort --profile black` run on changed source files (no changes needed)
- [ ] `doc/source/<name>.md` created for new sources (n/a — existing source)
- [x] TEST_CASES use real, publicly accessible addresses (not my own) (no TEST_CASES changes; all existing cases pass live)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/republicservices_com.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/republicservices_com.py
@@ -1,7 +1,7 @@
 from datetime import date, datetime, timedelta
 
 import requests
-from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Republic Services"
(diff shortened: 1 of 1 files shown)


bbr111:
why use static icon mappings?
you can use Icons.BULKY an so on instead.

markvp (approved):
Thanks for tracking down the 'W' vs 'week' bug — nice catch, and the before/after entry counts make the impact really clear. I pushed a small follow-up that takes bbr111's suggestion the whole way: swapped the remaining inline mdi:* strings in this source for the matching Icons enum members (the rendered icons are identical). Good to merge from my side.

PR #6627, opened 2026-06-14
Switch sensor_select and customize_select to LIST mode

### LINKED ISSUE #3611: [Feature]: Reduce drop down menue length
### I propose a feature for:

The integration in general

Hello great waste_collection_schedule team :)

is it possible to reduce the length of the drop down menue for the sensor selection?
In my browser it is to long, so that I can see all, even if I scroll down. You can see it in the picture below.

In my case I can't choose "New sensor" or I have to zoom the hole page ...

Image

Thanks a lot for this project!!!
Best regards

delight-f:
Dropdown overlays become unusable with a large sensor list. Switching to SelectSelectorMode.LIST renders a bounded, scrollable list box instead.

## Summary
Summary

Switches the `sensor_select` and `customize_select` selector fields from HA's default `DROPDOWN` overlay mode to `SelectSelectorMode.LIST`, rendering them as a bounded, scrollable list box instead.

### Problem

When a user has many sensors configured (common with waste collection schedules covering many waste types), the `sensor_select` and `customize_select` fields in the options flow render as oversized dropdown overlays that overflow the viewport, making the config form unusable.

### Changes

**File:** `custom_components/waste_collection_schedule/config_flow.py`

| Location | Field | Before | After |
|---|---|---|---|
| `WasteCollectionConfigFlow.async_step_customize_select` (L838) | `CONF_TYPE` | `SelectSelectorMode.DROPDOWN` | `SelectSelectorMode.LIST` |
| `WasteCollectionOptionsFlow.async_step_init` (L1097) | `"sensor_select"` | implicit `DROPDOWN` (no mode set) | `SelectSelectorMode.LIST` |
| `WasteCollectionOptionsFlow.async_step_init` (L1126) | `"customize_select"` | implicit `DROPDOWN` (no mode set) | `SelectSelectorMode.LIST` |

### Testing

- Formatted with `black` and `isort --profile black`
- No generated files in the diff
- No source modules were touched — test suite relevant to source contracts is unaffected
## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [X] `python -m black` and `python -m isort --profile black` run on changed source files


markvp:
As a change to the user interface, please include screenshots for before/after. Thanks 😊

markvp:
Thanks @delight-f, and apologies for the slow turnaround. I've rebased this onto current master (the branch had fallen a long way behind) and the change applies cleanly.

One thing before it goes in: two of the three selectors you switched keep `custom_value=True` — the `customize_select` CONF_TYPE (~line 837) and the uncustomized-types one (~line 1124). As far as I know, Home Assistant only honours `custom_value` in DROPDOWN mode; in LIST mode there's no free-text field, so users would lose the ability to enter a waste type that isn't already in the list. The third selector (the "add new sensor" list, `custom_value=False`) is a fixed set, so LIST suits it nicely.

Could you confirm custom types can still be entered with LIST mode on HA 2024.4? If they can't, the cleanest split would be to keep DROPDOWN for the two `custom_value=True` selectors and use LIST only for the fixed-choice one. Happy to get it merged once that's settled.

delight-f:
Thanks for the rebase @markvp!

I checked the Home Assistant frontend code for the 2024.4 release. It turns out that when custom_value=True, Home Assistant ignores the LIST mode and always renders the selector as a dropdown. In other words, those two selectors would never have been displayed as lists, regardless of the mode setting.

I've therefore reverted the LIST change for the two selectors that allow custom values, and kept it only for the fixed-choice "add new sensor" selector, where it does have an effect. I've pushed the updated commit.

markvp (approved):
Thanks @delight-f, and thanks for digging into the frontend source rather than guessing. Good to have it confirmed that LIST is ignored when custom_value is set, which means the split you have landed on (LIST only for the fixed-choice sensor selector) is exactly right.

Happy with this as it stands. The title and description still describe the original three-selector version, so I will tidy that up in the squash message.

Approving.

PR #6637, opened 2026-06-14
feat(toulouse_metropole_fr): add Toulouse Métropole waste collection source

fangedhex:
## Summary

Adds a new source for waste collection schedules from **Toulouse Métropole** open data portal, serving 37 communes in Haute-Garonne (31), France.

- Queries Opendatasoft API v2.1 for two datasets: ordures ménagèères and collecte selective
- Supports weekly and bi-weekly schedules (semaine paire/impaire)
- Street name lookup via libelle_voie field
- Waste types: Dechets non recyclables / Dechets recyclables
- HOW_TO_GET_ARGUMENTS_DESCRIPTION in en + fr
- PARAM_DESCRIPTIONS and PARAM_TRANSLATIONS in en + fr

## Testing

- test_sources.py -s toulouse_metropole_fr -l -i: 72 entries across 4 test cases
- python -m pytest tests/: 9/9 passed
- Formatted with black + isort
- Tested also on my own Home Assistant with my own street name and works without issue

(PR written with help of AI)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/toulouse_metropole_fr.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/toulouse_metropole_fr.py
@@ -1,0 +1,197 @@
+import datetime
+import logging
+
+import requests
+from waste_collection_schedule import Collection, Icons
+from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
+
+_LOGGER = logging.getLogger(__name__)
+
+TITLE = "Toulouse Métropole"
+DESCRIPTION = "Source pour la collecte des déchets de Toulouse Métropole (37 communes)."
+URL = "https://data.toulouse-metropole.fr"
+COUNTRY = "fr"
+
+TEST_CASES = {
+    "Colomiers - Avenue Henri Guillaumet": {
+        "street_name": "Avenue Henri Guillaumet",
+    },
+    "Toulouse - Rue de la Paix": {
+        "street_name": "Rue de la Paix",
+    },
+    # Confirmed in dataset: collecte sélective "semaine paire" (bi-weekly)
+    "Chemin Vié (bi-weekly CS)": {
+        "street_name": "Chemin Vié",
+    },
+    # Confirmed in dataset: ordures ménagères "Vendredi"
+    "Route de Fonbeauzard": {
+        "street_name": "Route de Fonbeauzard",
+    },
+}
+
+HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
+    "en": (
+        "Enter your street name exactly as it appears on the Toulouse Métropole open data portal. "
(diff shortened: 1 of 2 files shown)


markvp (changes requested):
Thanks for this, Sylvain, and for the thorough testing notes. The structure is in good shape (country, icons, en/fr translations, exceptions all check out), but I hit a correctness problem when testing against the live API that I'd like you to address before we merge.

`_parse_schedule` only recognises single bare weekday names (`Vendredi`, `Mercredi semaine paire`). The ordures-ménagères dataset, though, very often uses multi-day and daily values: `Le lundi et jeudi`, `Le lundi, mercredi et vendredi`, `7 jours / 7`. Those all fall through the `weekday is None` guard and get dropped silently, so affected streets return partial or empty schedules with no error. Your own `Toulouse - Rue de la Paix` test case shows it — one of its records is `Le lundi, mercredi et vendredi` and is currently lost.

Could you extend the parser to handle the `Le ` prefix, multiple days separated by `et` / commas, and the `7 jours / 7` daily case, then re-test against a few real multi-day streets?

Two smaller things while you're in there: (1) please confirm the API really has no per-date or ICS schedule (if it does, fetching real dates is preferable to generating them locally) and that `semaine paire` maps to even ISO week numbers on the actual calendar; and (2) please add a `doc/source/toulouse_metropole_fr.md` (and ideally `SOURCE_CODEOWNERS = ["@fangedhex"]`). Happy to help once those are in.

fangedhex:
﻿All review points addressed:

**1. _parse_schedule - multi-day and daily support**
- Replaced the fragile separator-splitting pipeline with a day_name in value loop over _DAY_MAP. Handles all separator variants (commas, et, Oxford comma, whitespace) without brittle string manipulation.
- Added a "7 jours / 7" guard returning all 7 weekdays.
- Re-tested against live API: Rue Sainte-Thérèse (3-day: "Le lundi, mercredi et vendredi") returns 32 entries, Rue des Lois (daily: "7 jours / 7") returns 56 entries, Rue de la Paix now returns 44 entries instead of the previous partial result.

**2. Added doc/source/toulouse_metropole_fr.md** with configuration docs and usage examples.

**3. Added SOURCE_CODEOWNERS = ["@fangedhex"]**

**4. Regenerated test cases** with comments explaining the schedule pattern each one exercises.

I rechecked for API dates, it doesn't give anything else except those string we need to parse. As written above, I rewrote with help of AI a more robust way possible to parse it.

*(This update was written with AI assistance.)*

markvp (approved):
Thanks for the quick turnaround, Sylvain. The reworked parser handles the multi-day and "7 jours / 7" cases cleanly now, and I confirmed the previously dropped records (e.g. "Le lundi, mercredi et vendredi") come through. doc file, codeowners, and the annotated test cases are all in good shape. Structural tests pass and formatting is clean. Looks good to merge.

PR #6638, opened 2026-06-14
Claude/sitepark ies shared service

bbr111:
## Summary

Introduces a shared `SiteparkIES` service that harmonises all sources running on the Sitepark IES / abto ("iKISS") waste-calendar platform, and adds one new municipality.

The Sitepark abto platform is used by a number of German municipalities. They all share the same two endpoints:
- `/output/autocomplete.php?out=json&type=abto&select=2&term=<street>` → `[[pois, "Street (Place)"], ...]`
- `/output/options.php?ModID=48&call=ical&pois=<pois>` → ICS calendar

Each request needs a `Referer` header pointing at the base domain (otherwise `403`). The shared service centralises autocomplete, street→pois resolution (with `ort` disambiguation), refid lookup via the `sf_locid` dropdown, and ICS download.

## Changes

- **New** `service/SiteparkIES.py` — shared client + `match_icon()` helper. The substring-based `match_icon()` is needed because the feeds append rhythm suffixes to waste types (e.g. `Restmülltonne 14-täglichen Rhythmus`), which a plain `ICON_MAP.get()` would miss.
- **New source** `muehlenkreis_de` — Mühlenkreis / Kreis Minden-Lübbecke (NRW). Autocomplete + iCal export confirmed working.
- **Converted from ICS YAML to Python sources** (now with street autocomplete): `ab_peine_de`, `gross_gerau_de`, `ilm_kreis_de`, `kreis_ploen_de`. Old `doc/ics/yaml/*.yaml` removed.
- **Harmonised** the 6 existing abto Python sources onto the shared service: `abfall_neunkirchen_siegerland_de`, `hilchenbach_de`, `kwb_goslar_de`, `landkreis_wittmund_de`, `lk_mecklenburgische_seenplatte_de`, `ostprignitz_ruppin_de`.
- All `ICON_MAP`s now use the `Icons` enum (no raw `mdi:` strings); all sources use `match_icon()`.
- Added optional `ort` parameter where the platform's global autocomplete can return the same street in multiple districts. Legacy parameter names (`city`/`street`/`location`) kept working.
- `doc/source/*.md` added for the 4 newly-converted sources and for muehlenkreis.

## Testing

- `python -m pytest tests/test_source_components.py -q` → all green (Icons enum, COUNTRY, PARAM_TRANSLATIONS allowlist).
- Sources tested live against their `TEST_CASES` via `test_sources.py`.


bbr111:
@markvp request your review, because this is a greater change for some sources.

if we don't want to to this with an extra service. i'm fine with that. we can also delete this.
just an idea.

markvp (approved):
Thanks for pulling all the abto/Sitepark sources together, Benjamin - this is a tidy consolidation and the shared match_icon() substring helper is a sensible answer to the rhythm-suffix problem. Structurally everything is green locally: black, isort, flake8 and tests/test_source_components.py (9 passed) all pass, the Icons enum is used throughout, COUNTRY is correct everywhere (and you've added the missing de to kwb_goslar and landkreis_wittmund), and the legacy city/street/location keyword names still resolve.

Since you asked for a call on the approach: I'm happy to go with the shared service - it removes a lot of duplicated autocomplete/ICS plumbing. One small thing to confirm before merge: for lk_mecklenburgische_seenplatte_de and landkreis_wittmund_de the positional argument order changed (the legacy keyword names are preserved, so YAML args: configs are fine, but anyone relying on positional order would shift). HA config is keyword-based so I think we're safe - just flagging it.

No blockers from me. Approving.

bbr111:
The legacy keyword names (city/street/location) are preserved, so existing args: configs keep working unchanged. Left the signatures as-is rather than reordering, since there's no caller that would be affected.

PR #6639, opened 2026-06-14
Migrate lint/format tooling to Ruff (pre-commit + VSCode)

bbr111:
Replace black, flake8 and isort with Ruff in .pre-commit-config.yaml and configure VSCode to use Ruff as formatter / import organiser.

Ruff is configured to mirror the previous tool chain so the diff stays minimal and behaviour is unchanged:
- line-length 88 (black default)
- lint select E,F,W,I (pycodestyle + pyflakes + isort profile=black)
- ignore E203,E501 (as flake8 did) and E721 (stricter in Ruff than pycodestyle, so kept off to avoid new findings)
- ruff-format runs on the full tree like black; the ruff lint hook excludes source/ exactly as the flake8 hook did
- pyupgrade, codespell, bandit, yamlfmt and mypy hooks are unchanged

The .py reformatting in this commit is the irreducible black->ruff formatting difference; no logic changes.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [ ] `python -m pytest tests/test_source_components.py -q` passes
- [ ] `python -m black` and `python -m isort --profile black` run on changed source files

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -15,6 +15,7 @@
 		"vscode": {
 			"extensions": [
 				"ms-python.python",
+				"charliermarsh.ruff",
 				"github.vscode-pull-request-github",
 				"ryanluker.vscode-coverage-gutters",
 				"ms-python.vscode-pylance"
@@ -25,12 +26,16 @@
 				"terminal.integrated.shell.linux": "/bin/bash",
 				"python.defaultInterpreterPath": "/usr/bin/python3",
 				"python.analysis.autoSearchPaths": false,
-				"python.linting.pylintEnabled": true,
-				"python.linting.enabled": true,
-				"python.formatting.provider": "black",
 				"editor.formatOnPaste": false,
 				"editor.formatOnSave": true,
 				"editor.formatOnType": true,
+				"[python]": {
+					"editor.defaultFormatter": "charliermarsh.ruff"
+				},
+				"editor.codeActionsOnSave": {
+					"source.fixAll.ruff": "explicit",
+					"source.organizeImports.ruff": "explicit"
+				},
 				"files.trimTrailingWhitespace": true
 			}
 		}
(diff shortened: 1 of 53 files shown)


bbr111:
**Still to be discussed:**

1. **Switch to Ruff to speed up linting.** Ruff replaces black + flake8 + isort in a single tool, which should noticeably cut down pre-commit and CI runtime. For now the config mirrors the old tool chain 1:1 (line-length 88, select E/F/W/I, same ignores) so the diff stays minimal and behaviour is unchanged.
2. **After minimal changes, we can make the setup stricter.** Once this lands as the behaviour-preserving migration, a follow-up can tighten things up, e.g.:
   - enable additional rule sets (`UP` for pyupgrade, `B` for bugbear, `SIM`, `C4`, …) — some overlap with existing hooks, so we'd consolidate
   - drop the temporary ignores (`E203`, `E501`, `E721`) once the affected code is cleaned up
   - decide whether `source/` should stay excluded from linting or get gradually included
   - consider folding the standalone codespell/bandit/pyupgrade hooks into Ruff where it has equivalents, to reduce the number of tools

markvp (approved):
Thanks Benjamin, this is a clean, well-scoped migration. The Ruff config mirrors the old black/flake8/isort behaviour 1:1, the source/ lint exclusion is preserved, and I confirmed the .py reformatting is purely the black->ruff formatting difference with no logic changes. ruff format --check and ruff check are both no-ops on HEAD, and the structural test suite passes (9 passed). Happy to land this as the behaviour-preserving baseline; the follow-up tightening you outlined (additional rule sets, dropping the temporary ignores, folding in codespell/bandit) makes sense as separate PRs. Approving.

markvp:
do you need to update the github actions too?

bbr111:
> do you need to update the github actions too?

good point.

PR #6640, opened 2026-06-14
fix(fylde_gov_uk): integration changed parameter

j-webb:
## Summary

Fylde Gov integration changed parameter from "Subject" to "Description"
Also added a fix for when UPRN is returned as a `float`

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m black` and `python -m isort --profile black` run on changed source files


markvp (approved):
Thanks for the fix — switching to the Description field and tolerating a float UPRN both look spot on. I pushed a couple of small cleanups on top: kept the type: ignore inline so mypy still picks it up, dropped the raw mdi: icon fallback in favour of the Icons enum, and switched the credential and UPRN failures to the typed SourceArgument exceptions. Good to merge.

Issue #1391, opened 2023-11-06
[Source Request]: Moreton Bay, Australia


Slippery-Silvia:
### Municipality / Region

Moreton Bay, Australia

### Collection Calendar Webpage

https://www.moretonbay.qld.gov.au/Services/Waste-Recycling/Collections/Bin-Days

### Example Address

4 Caspian Parade, Warner QLD 4500

### Collection Data Format

As ICS/ICAL file

5ila5:
Moreton Bay is supported by the [ICS source](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/ics.md)
[Documentation for Moreton Bay](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/ics/moretonbay_qld_gov_au.md) added in above PR

PR #6646, opened 2026-06-15
feat(ics): add optional impersonate arg to pass TLS-fingerprinting WAFs

### LINKED ISSUE #6645: [Feature]: Add optional impersonate arg to the generic ICS source for TLS-fingerprinting WAFs

**Add an optional `impersonate` argument to the generic ICS source so feeds behind TLS-fingerprinting WAFs (Incapsula/Imperva, Cloudflare) can be fetched.**

**Problem.** Some councils put their ICS/webcal feed behind a WAF that fingerprints the TLS/JA3 handshake, not just the `User-Agent`. The generic ICS source then gets a hard `HTTP 403` and no custom `headers:` value fixes it — I verified that a browser `User-Agent` header still 403s, and only curl_cffi's `impersonate` (which forges the full browser TLS fingerprint) returns the calendar:

```
no-UA/default fingerprint : 403  (398-byte "Access Denied" page)
browser User-Agent header : 403  (same)
impersonate="chrome"      : 200  (BEGIN:VCALENDAR ...)
```

**Concrete case: Moreton Bay (QLD, AU).** Requested in #1391 and resolved as "use the generic ICS source" with `webcal://www.moretonbay.qld.gov.au/bincal?externalId=<id>`. That path has since broken — the council added an Incapsula WAF, so the generic source now 403s and there's no config-only way to recover.

**Prior art in this repo.** Dedicated sources already work around the same WAF class with `requests.Session(impersonate="chrome")` — e.g. `banyule_vic_gov_au`, `goldenplains_vic_gov_au`, `valeofglamorgan_gov_uk` (the last two fixing 403s in #6324 / #6468). The integration already depends on `curl_cffi` (#6442), so the capability is present; the generic ICS source just doesn't expose it.

**Proposal.** Add an optional `impersonate: str | None = None` argument to the ICS `Source`, passed straight through to the curl_cffi `get`/`post` calls in `fetch_url`. Default `None` = current behaviour, so it's fully backward-compatible, and it's annotated like the existing optional ICS args (`regex`, `split_at`) so the config flow renders it automatically. A user behind such a WAF then just sets `impersonate: chrome`, and providers like Moreton Bay that were designated as generic-ICS keep working without a bespoke source.

I'm aware the established pattern for these 403s has been a dedicated source — happy to go that way for Moreton Bay specifically if you'd prefer, but the generic option seemed worth raising since it's a few lines, helps every WAF-blocked feed, and keeps Moreton Bay on the path #1391 chose. I have a PR ready for the generic approach and will link it here.

<sub>Drafted with AI assistance (Claude Code) and reviewed by me before posting.</sub>

jjsmackay:
## Summary

Adds an optional `impersonate` argument to the generic ICS source, passed through to the underlying `curl_cffi` `get`/`post` calls. This lets the client mimic a real browser's TLS/JA3 fingerprint so feeds behind a WAF (Incapsula/Imperva, Cloudflare) that returns `HTTP 403` to non-browser clients can be fetched.

Default is `None`, so existing behaviour is unchanged. It's annotated like the other optional ICS args (`regex`, `split_at`), so the config flow renders it automatically once `sources.json` is regenerated.

This is the generic counterpart to the per-source workaround already used by `banyule_vic_gov_au`, `goldenplains_vic_gov_au`, and `valeofglamorgan_gov_uk`. Motivating case: Moreton Bay QLD (#1391), resolved as a generic-ICS provider and since put behind an Incapsula WAF. See #6645 for the discussion.

Verified against the affected feed:

| call | result |
|------|--------|
| `impersonate=None` (default) | HTTP 403 |
| `impersonate="chrome"` | 200 — 55 collections parsed |

## Type of change

- [x] Bug fix / source fix
- [x] Other (small enhancement to the generic ICS source)

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (9 passed)
- [x] `python -m black` and `python -m isort --profile black` run on changed source files
- [ ] `doc/source/<name>.md` created for new sources — n/a (not a new source; updated `doc/source/ics.md`)
- [ ] TEST_CASES use real, publicly accessible addresses — n/a (no new source/test cases)

<sub>Implemented with AI assistance (Claude Code) and reviewed by me before submitting.</sub>

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ics.py
@@ -113,6 +113,7 @@ _LOGGER = logging.getLogger(__name__)
 PARAM_TRANSLATIONS = {
     "en": {
         "version": "(Deprecated) Version, has no effect anymore",
+        "impersonate": "Browser to impersonate (e.g. 'chrome') to pass TLS-fingerprinting WAFs",
     },
     "de": {
         "url": "URL",
@@ -127,6 +128,7 @@ PARAM_TRANSLATIONS = {
         "version": "(Veraltet) Version, hat keine Auswirkung mehr",
         "verify_ssl": "SSL-Verifizierung aktivieren",
         "headers": "Headers",
+        "impersonate": "Zu imitierender Browser (z.B. 'chrome'), um TLS-Fingerprinting-WAFs zu passieren",
     },
 }
 
@@ -146,6 +148,7 @@ class Source:
         version: int | None = None,
         verify_ssl: bool = True,
         headers: dict = {},
+        impersonate: str | None = None,
     ):
         self._url = re.sub("^webcal", "https", url) if url else None
         self._file = file
@@ -170,6 +173,7 @@ class Source:
         self._verify_ssl = verify_ssl
         self._headers = HEADERS
         self._headers.update(headers)
+        self._impersonate = impersonate
 
     def fetch(self):
         if self._url is not None:
@@ -227,6 +231,7 @@ class Source:
                 params=flat_params,
                 headers=self._headers,
                 verify=self._verify_ssl,
+                impersonate=self._impersonate,
             )
         elif self._method == "POST":
             r = requests.post(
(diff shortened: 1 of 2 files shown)


markvp (approved):
Nice, well-targeted addition — this is the right generic home for the curl_cffi impersonation trick we'd otherwise keep duplicating per source. Verified the impersonate=None default leaves existing behaviour untouched, the arg threads cleanly into both the GET and POST paths, the doc and translation entries are consistent, and the structural test suite passes (9 passed). Thanks for the thorough write-up and the before/after verification. Approving.

Issue #3500, opened 2025-01-08
[Bug]: Awista Düsseldorf managed by Awista Kommunal not muellmax anymore


ChiyoDragon:
### I Have A Problem With:

The integration in general

### What's Your Problem

It seems the Waste Collection Schedule in Home Assistant is no longer working for Düsseldorf as of January 1, 2025. The service provider for waste management has changed from "Müllmax" to "Awista Kommunal" for private households.

### Checklist Source Error

- [ ] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [ ] Checked that the website of your service provider is still working
- [ ] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [ ] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

Nachhall:
Same problem here!

EleRas:
You can use the calendar provided with the generic - ICS - import as a workaround. After selecting street / house number, the ICS link is static and can be used.

hpuac:
Yep, more detailed step by step:
Go to: https://www.awista-kommunal.de/abfallkalender
Search for your Address click on "Kalender anzeigen"
There is a Button saying "iCalendar-Datei herunterladen (ICS)".
Do a right click and copy the URL.
Example: 
`https://www.awista-kommunal.de/abfallkalender/2e3f0916-59b9-489d-b112-00c31fb83267/calendar.ics`

Now go to your configuration.yaml and change the `waste_collection_schedule` config.
Example:
```
waste_collection_schedule:
  sources:
    - name: ics
      args:
        url: https://www.awista-kommunal.de/abfallkalender/2e3f0916-59b9-489d-b112-00c31fb83267/calendar.ics
```
Enjoy

makanimike:
Does the entity give us all upcoming dates though? My skills only let me see the next one on the calendar entity.

makanimike:
I have my calendar populated.  
[...]

PR #6707, opened 2026-06-21
fix(muellmax): remove unusable Düsseldorf service

Zaunei:
## Summary

AWISTA Düsseldorf [switched](https://github.com/mampfes/hacs_waste_collection_schedule/issues/3500) the backend and you need to use the ICS source now. Since the Müllmax integration fails to work, it leads to user confusion during the setup process.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/MuellmaxDe.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/MuellmaxDe.py
@@ -2,11 +2,6 @@
 
 
 SERVICE_MAP = [
-    {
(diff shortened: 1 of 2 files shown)


markvp (approved):
Thanks for the cleanup. Removing the broken AWISTA Düsseldorf service from both SERVICE_MAP and the wizard is the right move now that they've moved off the Müllmax backend (per #3500), and it matches how RSAG was handled in #6553. Structural tests pass and the diff is clean. Approving.

PR #6733, opened 2026-06-30
Add human-readable daysToText attribute to sensors (#6534)

### LINKED ISSUE #6534: [Feature]: Add a human-readable attribute for displaying time until collection
### I propose a feature for:

Sensors

Hello,

First, thank you for maintaining this integration. It has been very reliable and is one of the most useful integrations in my Home Assistant setup.

I would like to suggest adding a human-readable attribute that represents the time remaining until the next collection.

Currently, the sensor state is a numeric value representing the number of days until pickup, which is ideal for automations, calculations, and visibility conditions. For example:

11

However, many users display waste collection information using Home Assistant Badges. Badges are a common choice because they can easily be shown or hidden based on the number of days remaining (for example, only displaying the badge when collection is within the next 5 days).

The problem is that Badges do not support templates or value formatting. They can only display the sensor state exactly as provided.

As a result, a badge displays:

11

instead of:

11 days

Users who want a more user-friendly display must create additional template sensors solely for presentation purposes.

A possible solution would be to add a human-readable attribute while keeping the existing numeric state unchanged. For example:

state: 11

attributes:
  formatted: "11 days"

or even:

state: 21

attributes:
  formatted: "3 weeks"

This would preserve full compatibility with automations while providing a simple way for dashboard users to display meaningful information without creating extra template entities.

Even a basic formatted attribute such as "11 days" would be a significant improvement and would reduce the need for presentation-only template sensors.

Thank you for considering this feature request.

---

Den versionen fokuserar på ett konkret UI-problem i Home Assistant snarare än på en personlig preferens, vilket brukar öka chansen att utvecklarna faktiskt överväger förslaget.

Coderzz69:
## Summary

Fixes #6534 

This PR adds a human-readable `daysToText` string to the `extra_state_attributes` of the sensor (e.g. "Today", "Tomorrow", "in X days"). 

Currently, the default state can be overwritten by users with a `value_template` to isolate the integer for conditions, but this strips out the human-readable formatting, making it difficult to use in dashboard elements like Badges that do not natively support templating. By surfacing `daysToText` directly in the attributes alongside the existing numeric `daysTo`, users can easily map the raw integer to automations and the formatted string to UI Badges without needing to create secondary template sensors.

Additionally, this PR introduces a standalone test suite in `tests/test_sensor.py` to ensure the core logic handles the integer mappings correctly across edge cases (`0`, `1`, `> 1` days).

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other (Feature enhancement & tests)

## Checklist

- [ ] `doc/source/<name>.md` created for new sources (N/A)
- [ ] TEST_CASES use real, publicly accessible addresses (not my own) (N/A)

### DIFF
--- a/custom_components/waste_collection_schedule/sensor.py
+++ b/custom_components/waste_collection_schedule/sensor.py
@@ -51,6 +51,15 @@ class DetailsFormat(Enum):
     hidden = "hidden"  # hide details
 
 
+def get_days_to_text(days: int) -> str:
+    """Return a human-readable string for days until collection."""
+    if days == 0:
+        return "Today"
+    if days == 1:
+        return "Tomorrow"
+    return f"in {days} days"
+
+
 PLATFORM_SCHEMA = PLATFORM_SCHEMA.extend(
     {
         vol.Required(CONF_NAME): cv.string,
@@ -383,6 +392,7 @@ class ScheduleSensor(SensorEntity):
         if len(upcoming1) > 0:
             if self._add_days_to:
                 attributes["daysTo"] = upcoming1[0].daysTo
+                attributes["daysToText"] = get_days_to_text(upcoming1[0].daysTo)
 
         self._attr_extra_state_attributes = attributes
         self._add_refreshtime()
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -1,0 +1,22 @@
+"""Tests for sensor logic."""
+
+__author__ = "Coderzz69"
+
+from custom_components.waste_collection_schedule.sensor import get_days_to_text
+
+
+def test_get_days_to_text_today():
+    """Test that 0 days returns 'Today'."""
+    assert get_days_to_text(0) == "Today"
+
+
+def test_get_days_to_text_tomorrow():
+    """Test that 1 day returns 'Tomorrow'."""
+    assert get_days_to_text(1) == "Tomorrow"
+
+
+def test_get_days_to_text_future():
+    """Test that >1 days returns formatted string."""
+    assert get_days_to_text(2) == "in 2 days"
+    assert get_days_to_text(7) == "in 7 days"
+    assert get_days_to_text(11) == "in 11 days"
--- a/tests/test_sensor.py
+++ b/tests/test_sensor.py
@@ -1,7 +1,5 @@
 """Tests for sensor logic."""
 
-__author__ = "Coderzz69"
-
 from custom_components.waste_collection_schedule.sensor import get_days_to_text
 
 


bbr111:
Good idea, but what about translations?

bbr111 (changes requested):
Thanks for this, and for adding tests — the ergonomics idea behind `daysToText` is reasonable. Before this can go in, though, there's a design question that needs resolving:

**Translations (blocking).** The helper hardcodes English `"Today"` / `"Tomorrow"` / `"in {days} days"` into `extra_state_attributes`. This integration is used predominantly in non-English locales, and Home Assistant doesn't translate arbitrary attribute *values*, so a fixed-English attribute would ship untranslated to everyone. Could you outline how you'd like to handle localisation here? Options include exposing the string only through a proper i18n path, or keeping the numeric `daysTo` (which users already template today) and dropping the baked-in English string.

Two smaller things while you're in there:
- `tests/test_sensor.py` won't actually run in CI — `pytest.ini` restricts discovery to `python_files = test_source_components.py`, so these assertions are collected only when the file is invoked by name. Worth confirming how you'd like them wired into the collected suite (or folding them into an already-discovered module).
- Please drop the unused `__author__` variable to keep the test module minimal (per the Copilot note).

Happy to re-review once the translation approach is settled.

markvp (changes requested):
Thanks for picking this up, and for the tests. The `__author__` removal in 2fbf9b6a covers one of the earlier notes, so that one's done.

@bbr111's translation question is still the blocker, and I agree with it. Home Assistant doesn't translate attribute values, so a hardcoded "Today" / "Tomorrow" / "in X days" would reach a mostly non-English userbase in English only. Worth settling the approach before more code goes in.

One thing that might help frame it: the integration already ships this exact string as a `value_template` preset ("in .. days / Tomorrow / Today" in `config_flow.py`, also in `doc/faq.md`). So the case for the attribute really rests on badges not supporting templates, which is the heart of #6534. If you can make that argument explicitly, and pair it with a localisation story, the feature has a much clearer path. An alternative worth considering is exposing the next collection as a timestamp and letting the frontend render relative time in the user's own locale.

Two smaller things for whenever the shape is settled: `tests/test_sensor.py` isn't picked up by the default pytest run (`pytest.ini` limits `python_files` to `test_source_components.py`), and the sibling test modules note that in a docstring and `sys.path.append` into the component directory rather than importing the package root. Copying that pattern would keep it consistent and avoid pulling in all of homeassistant. And `doc/installation.md` needs a row for the new attribute.

Happy to re-review once the localisation approach is decided.

bbr111 (changes requested):
Thanks again for the work here, and sorry for the slow follow-up.

Since this PR was opened, the `release/3.0.0` line has landed the change that #6534 was asking for, and it took the language-neutral route that @bbr111 and @markvp were pointing at. `sensor.py` now always sets `daysTo`, `date` (ISO), `next_types` and `color` on the next-collection sensor, with an explicit note that these are raw values so cards and templates can render them in the user's own language — and there's a new dedicated **Days until collection** sensor (numeric, `duration` device class, unit `d`) in the default overview set for automations. Both are documented in `doc/installation.md` under "Attributes of a sensor". For the badge case specifically, the `date` attribute lets the frontend render relative time in the user's own locale without a template.

That means the translation question that was blocking this PR has effectively been answered the other way: rather than adding a formatted English string to the attributes, we keep the attributes language-neutral. The formatted string stays available where it can be localised — the `"in .. days / Tomorrow / Today"` `value_template` preset in `config_flow.py` (and `doc/faq.md`) already produces exactly this text.

Practically, the PR also no longer applies: it conflicts in `sensor.py`, and its anchor `if self._add_days_to:` is gone — `add_days_to` is now deprecated and inert because `daysTo` is always set.

So I'd like to close this as superseded rather than ask you to rebase onto a design that's since been decided differently. Please do take a look at the 3.0.0 attribute set and shout if it doesn't cover your use case — if there's a gap (say, a `timestamp` device-class entity for the next collection so HA's own relative-time rendering kicks in), that would be a great, and properly localisable, follow-up, and one I'd be glad to review.

Separately, thank you for the `tests/test_sensor.py` instinct — it's the right one. Worth knowing for next time: `pytest.ini` is an explicit allowlist, so a new test module has to be added to `python_files` or it silently never runs. That's a real gap and not your fault.

bbr111:
Superseded by #7496 on release/3.0.0.

PR #6735, opened 2026-06-30
Update Sunderland scraper for new portal (#6619)

### LINKED ISSUE #6619: [Source Defect]: Sunderland City Council source stopped working
### Source Name

sunderland_gov_uk

### What's happening?

Hi, the Sunderland City Council source stopped working on 6th June returning calendar.sunderland_city_council as unavailable and sensor.upcoming_waste as unknown. I have attempted some AI (claude) assisted debugging but got stuck trying to run the updated .py file.

In case it helps I have attached the AI's summary of what was found/done and a copy of the 'fixed' sunderland_gov_uk.py

[github_issue.pdf](https://github.com/user-attachments/files/28872223/github_issue.pdf)

[sunderland_gov_uk.py](https://github.com/user-attachments/files/28872442/sunderland_gov_uk.py)

Image
Image
Image

Let me know if you need anything else - thanks

### Error Logs

```Shell
AI response:
There are no error log entries at all for the waste collection integration. The scraper fails silently because the old ASP.NET endpoint returns HTTP 200 with empty HTML rather than an error, so nothing gets logged.
However, let me grab the current logs now that we've been running debug-level fetches, in case anything useful was captured:
Get LogsGet Logs
Confirmed — nothing at all. You can note that in the GitHub issue: no entries appear in the HA error log, which is itself diagnostic — it demonstrates the silent failure mode caused by the old endpoint returning HTTP 200 with empty content rather than an error code. That's actually useful information for the maintainer because it explains why users may not immediately realise the scraper is broken.
```

### Source Configuration

```YAML
see attached screenshots
```

### Checklist

Coderzz69:
## Summary

Fixes #6619

The Sunderland City Council recently migrated their waste collection schedule to a new `.NET` API endpoint structure (`BINCOLLECTIONCHECKERNEWV3`). This PR updates the `sunderland_gov_uk` source to use the new multi-step POST sequence required to resolve a UPRN and extract the collection dates. 

It handles the new form data payloads, base64-decoded variables, and the updated HTML structure of the result panels.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [ ] `doc/source/<name>.md` created for new sources (N/A)


bbr111 (approved):
Thanks for updating this for the new Sunderland portal! I made a couple of small tweaks to address the review-bot feedback: removed the unused imports/SUBMIT_URL constant and the dead full_details line, fixed the _get_hidden_fields docstring, switched the bad-postcode / unknown-address paths to SourceArgumentNotFound / SourceArgumentNotFoundWithSuggestions, and added a guard so we raise (rather than silently return an empty list) when no dates parse — which is the exact silent-failure mode from #6619. ruff and tests/test_source_components.py (9 passed) are green. One thing I couldn't do from here is a live fetch against the portal, so I'd suggest a quick live-test before merge.

PR #6737, opened 2026-07-01
Add Gladstone Regional Council waste collection source

kodie-arch:
## Summary

Adds a new waste collection source for Gladstone Regional Council, Queensland, Australia.

The source retrieves collection information from the council's public ArcGIS REST service.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Notes

The council API provides:
- Weekly general waste
- Fortnightly recycling (Week 1 / Week 2)
- Garden Organics (GO) service availability

The API currently does not expose the recycling/GO anchor date directly, so the implementation derives collection dates from the council's published schedule. This can be updated if a more reliable source becomes available in the future.

## Checklist

- [x] Tested with a real address
- [x] Source follows project structure
- [x] New source filename is unique

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/gladstone_qld_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/gladstone_qld_gov_au.py
@@ -1,0 +1,144 @@
+from __future__ import annotations
+
+from datetime import date, timedelta
+
+import requests
+
+from waste_collection_schedule import Collection, Icons
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
+
+TITLE = "Gladstone Regional Council"
+DESCRIPTION = "Waste collection schedule for Gladstone Regional Council, Queensland."
+URL = "https://www.gladstone.qld.gov.au/Living-Here/Services/Waste-and-Recycling/Bin-Collection-Day"
+COUNTRY = "au"
+SOURCE_CODEOWNERS = ["@kodie-arch"]
+
+TEST_CASES = {
+    "85 J Hickey Avenue, Clinton": {
+        "address": "85 J Hickey Avenue",
+        "suburb": "CLINTON",
+    },
+}
+
+PARAM_DESCRIPTIONS = {
+    "en": {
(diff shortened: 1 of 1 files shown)


bbr111 (changes requested):
Thanks for the contribution, and for wiring up the ArcGIS lookup cleanly! Unfortunately I can't merge this as-is because of one blocking design issue plus a few smaller items.

Blocking: the recycling and garden-organics dates are derived from a hardcoded reference date (WEEK_2_ANCHOR = date(2026, 7, 3)) rather than fetched live. The project doesn't accept sources that rely on hardcoded dates or schedule rules — this will drift out of sync over time and needs manual maintenance. Your PR notes the API doesn't expose the anchor directly; before this can be merged we need the fortnightly parity to come from upstream data (a nextRecycleDate/anchor field in the ArcGIS layer, a related table, or the council's ICS/feed). If the council genuinely publishes no machine-readable anchor, the source isn't viable in its current form.

Smaller items to fix in the same pass:
- Add the required doc/source/gladstone_qld_gov_au.md doc page (every new source needs one).
- SourceArgumentNotFound("address", ...) — pass the failing address as the value and put the explanation in message_addition; right now the message string is passed as the value.
- Escape single quotes in the address/suburb before building the ArcGIS where clause (e.g. "O'Connell" breaks the query). See red_bank_tn_us.py.
- Use Icons.GARDEN for the "Garden Organics" stream instead of Icons.ORGANIC.
- Consider adding PARAM_TRANSLATIONS alongside PARAM_DESCRIPTIONS.

Happy to re-review once the live-data anchor is sorted.

bbr111:
Thanks again for the contribution.

The ArcGIS address lookup is a good foundation, but the source still derives the recycling and garden-organics schedule from a hardcoded calendar anchor. Without a live upstream reference date, the generated dates may silently drift from the council's actual schedule.

Since the blocking issue and the other requested changes have not been addressed, I am closing this draft for now to keep the pull request queue manageable.

Please feel free to reopen it or submit a new pull request if a reliable upstream anchor, next-collection date, related ArcGIS table, or other live schedule source can be found.

PR #6754, opened 2026-07-04
fix: birmingham city council

### LINKED ISSUE #6745: [Source Defect]: Birmingham City Council
### Source Name

birmingham_gov_uk

### What's happening?

Birmingham City Council have finally after 18 months started recycling and food waste collection!!!

I think the source needs to be updated to present these new collections.
Currently on the web site:

Friday 10 July | Food
Friday 10 July | Mixed recycling
Friday 10 July | Rubbish

Friday 17 July | Food
Friday 17 July | Rubbish

Friday 24 July | Food
Friday 24 July | Mixed recycling

Friday 31 July | Food
Friday 31 July | Rubbish

Friday 07 August | Food
Friday 07 August | Mixed recycling

### Error Logs

```Shell
No issues found for search term 'waste_collection_schedule'
```

### Source Configuration

```YAML
Not configured by YAML
```

### Checklist

jamesonuk:
## Summary

Fixes: https://github.com/mampfes/hacs_waste_collection_schedule/issues/6745

Updates following changes to website

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/birmingham_gov_uk.py
@@ -29,30 +29,24 @@ ICON_MAP = {
     "Mixed recycling": Icons.RECYCLING,
 }
 
-YEAR_ROLLOVER_THRESHOLD_MONTHS = 6
 
-
-def _parse_collection_date(raw_date: str, today: datetime) -> date | None:
+def _parse_collection_date(raw_date: str, today: datetime) -> date:
     """
     Parse a 'Weekday DD Month' table entry (year omitted).
 
-    Schedules only ever contain current/future dates plus one just-completed
-    row from the past week, so year rollover only happens at the Dec -> Jan
-    boundary - never mid-year. A month-gap > 6 is unambiguous evidence of a
-    genuine wraparound (e.g. today=Dec, parsed=Jan); anything smaller is just
-    "earlier this month/last month" and is left alone (the caller filters it
-    out via the cutoff if it's genuinely in the past).
+    If we are in the last quarter of the year and the parsed month is in the first quarter, assume the date is in the next year.
+    if we are in the first quarter of the year and the parsed month is in the last quarter, assume the date is in the previous year.
     """
     try:
         parsed = dateutil_parse(raw_date, default=today, fuzzy=True)
-    except (ParserError, ValueError, OverflowError):
-        return None
+    except (ParserError, ValueError, OverflowError) as exc:
+        raise ValueError(f"Could not parse collection date '{raw_date}'") from exc
 
-    if (
-        parsed.month < today.month
-        and today.month - parsed.month > YEAR_ROLLOVER_THRESHOLD_MONTHS
-    ):
+    # handle year crossover
+    if today.month >= 10 and parsed.month <= 3:
         parsed = parsed.replace(year=parsed.year + 1)
+    elif today.month <= 3 and parsed.month >= 10:
+        parsed = parsed.replace(year=parsed.year - 1)
 
     return parsed.date()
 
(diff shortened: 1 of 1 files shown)


bbr111 (changes requested):
Thanks for this — the switch to the new collection-day checker matches what Birmingham's rolled out. One regression risk: this drops the old /xfp/form/619 endpoint entirely, but I tested one of the existing TEST_CASES ("Ludgate Loft Apartments", B3 1DW / UPRN 10033389698) against the new endpoint directly and it returns "Unfortunately we have been unable to find your rubbish collection schedule" — so this address, which works today via the legacy endpoint, would break with this change. Could you keep the old endpoint as a fallback for addresses the new checker can't resolve? Also a small nit: ICON_MAP.get(collection_type, "mdi:help") still falls back to a raw mdi string — worth defaulting to None instead, or a canonical Icons member.

jamesonuk:
> Thanks for this — the switch to the new collection-day checker matches what Birmingham's rolled out. One regression risk: this drops the old /xfp/form/619 endpoint entirely, but I tested one of the existing TEST_CASES ("Ludgate Loft Apartments", B3 1DW / UPRN 10033389698) against the new endpoint directly and it returns "Unfortunately we have been unable to find your rubbish collection schedule" — so this address, which works today via the legacy endpoint, would break with this change. Could you keep the old endpoint as a fallback for addresses the new checker can't resolve? Also a small nit: ICON_MAP.get(collection_type, "mdi:help") still falls back to a raw mdi string — worth defaulting to None instead, or a canonical Icons member.

I have done a little cleaning and sorted out the hard coded icon (I hadn't touched that bit).

Interestingly the old endpoint it no longer linked from anywhere on the council website so that address that doesn't work cannot check unless they happen to know that URL.    I don't have time just now to build in a failover right now but will see if I get tine during the week maybe

bbr111 (changes requested):
Thanks for updating this to the new collection-day checker, and for cleaning up the icon fallback (now Icons.GENERAL_WASTE) — that resolves the raw-mdi nit nicely.

The one thing still holding this back is the regression flagged before: dropping the /xfp/form/619 endpoint means the existing "Ludgate Loft Apartments" TEST_CASE (UPRN 10033389698, B3 1DW) no longer resolves — the new checker returns "Unfortunately we have been unable to find your rubbish collection schedule" for it, so fetch() would raise for an address that works today. Could you either (a) reinstate the legacy endpoint as a fallback for addresses the new checker can't resolve, or (b) if that address genuinely can no longer be looked up anywhere, replace that TEST_CASE with a currently-resolvable one and note the coverage change? Happy to re-review as soon as that's sorted.

Everything else looks good — ruff is clean and the structural test suite passes locally.

billmaterial:
FYI: I initially posted the issue, I can confirm this does work now with my real address. Nice work thank you :)

image

jamesonuk:
> The one thing still holding this back is the regression flagged before: dropping the /xfp/form/619 endpoint means the existing "Ludgate Loft Apartments" TEST_CASE (UPRN 10033389698, B3 1DW) no longer resolves — the new checker returns "Unfortunately we have been unable to find your rubbish collection schedule" for it, so fetch() would raise for an address that works today. Could you either (a) reinstate the legacy endpoint as a fallback for addresses the new checker can't resolve, or (b) if that address genuinely can no longer be looked up anywhere, replace that TEST_CASE with a currently-resolvable one and note the coverage change? Happy to re-review as soon as that's sorted.

I have updated the address to one that works.   I have tried addresses all across the city and I have only found a few that don't work.

The updated URL is the only available link via the council website so whilst the old endpoint still exists there is no way to navigate to it for normal users.    I can only see that the couple of addresses that do not resolve are just glitches in the system or awaiting some sort of final migration.    I am sure the old endpoint will get removed at some point.

jamesonuk:
however I will add fallback I think as it should be easy enough to remove

bbr111 (approved):
Thanks for adding the legacy fallback and swapping in a resolvable TEST_CASE — that fully addresses the regression from before.

One small tweak while re-reviewing: the fallback branch did `return self.legacy_api` (returning the bound method) rather than `return self.legacy_api()`, so fetch() would have handed a method object back instead of a list[Collection] whenever the new checker can't resolve an address — i.e. the exact fallback path. I pushed a one-line fix to call it.

Verified locally: ruff clean, tests/test_source_components.py passes (9), and a live run resolves all five TEST_CASES — "Ludgate Loft Apartments" now returns via the legacy fallback as intended. Approving.

PR #6770, opened 2026-07-06
Fix memotri_agglo_pau_fr: site switched from BunkerWeb to Django CSRF form + Akamai bot protection

vchatela:
## Problem

The source has been returning `404 Client Error: Not Found for url: https://memotri.agglo-pau.fr/` for all users.

Two things changed on the site:

1. **BunkerWeb removed**: The site replaced its BunkerWeb proof-of-work challenge with a standard Django CSRF form submission. The old code tried a direct `GET /consignes-tri/{ban_id}/{label}/` which now returns 404 — the server only serves that URL after a valid form `POST` has been submitted from the home page.

2. **Akamai Bot Manager added**: The site is now behind Akamai CDN with bot protection that blocks non-browser TLS fingerprints (JA3/JA4). Plain `requests.Session()` — even with Chrome headers — is detected and served a cached 404 at the CDN edge. The existing `curl_cffi` dependency (already in `manifest.json`) resolves this by impersonating Chrome's TLS stack.

## Fix

- Remove `_solve_challenge()` (BunkerWeb PoW — no longer present on the site)
- Remove unused `hashlib` and `urllib.parse.quote` imports  
- Add `_get_csrf_token()`: uses `curl_cffi` to GET `/` and extract the Django CSRF token
- Update `fetch()`: POST the address form (`csrfmiddlewaretoken` + `select_adresse` in `banId&label` format as expected by the select2 widget), then follow the redirect to `/consignes-tri/…` to parse the schedule HTML

The new flow mirrors exactly what a browser does:
```
GET /  →  extract CSRF token  →  POST / (form)  →  redirect  →  parse /consignes-tri/{id}/{label}/
```

## Testing

Validated live against both test cases:
- `39 avenue larribau 64000 Pau` → 133 entries returned ✓  
- `2 rue de l'Industrie 64320 Idron` → 133 entries returned ✓

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/memotri_agglo_pau_fr.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/memotri_agglo_pau_fr.py
@@ -1,10 +1,9 @@
-import hashlib
 import re
 from datetime import date, timedelta
-from urllib.parse import quote
 
 import requests
 from bs4 import BeautifulSoup
+from curl_cffi import requests as curl_requests
 from dateutil.rrule import (
     FR,
     MO,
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for the thorough writeup and the live validation against both test cases. The switch to curl_cffi (impersonate="chrome131") for the Akamai-protected site is exactly right, the BunkerWeb PoW removal and the new Django CSRF GET -&gt; token -&gt; POST -&gt; redirect flow look correct, and the unused hashlib/quote imports were cleaned up. Structural test suite passes locally (9 passed) and lint/format are clean. Looks good to merge.

PR #6774, opened 2026-07-06
Add RITL waste collection schedule source

JeanPascalDC:
### Description
Add support for waste collection schedules managed by **Régie intermunicipale des Trois-Lacs (RITL)** in the Laurentides region of Quebec, Canada.

This integration covers the following 10 municipalities:
- Ivry-sur-le-Lac
- Lac-Supérieur (Secteur 1, Secteur 2)
- Lantier
- Montcalm (Secteur 1)
- Mont-Blanc (Secteur Est, Secteur Ouest)
- Sainte-Agathe-des-Monts (Secteur Centre-ville, Secteur Nord, Secteur Sud / Fatima)
- Sainte-Lucie-des-Laurentides
- Val-des-Lacs
- Val-David (Secteur 1, Secteur 2, Secteur 3)
- Val-Morin (Val-Morin Est, Val-Morin Ouest)

### Added Files
- `custom_components/waste_collection_schedule/waste_collection_schedule/source/ritl_ca.py`
- `doc/source/ritl_ca.md`

### Test Cases Verified
The category mapping was verified by extracting data from the live widgets on the RITL site. The ICS feeds successfully return event lists which are correctly parsed and mapped to the canonical `Icons` enum:
- `Compost` -> `Icons.ORGANIC` (mdi:leaf)
- `Recyclage` -> `Icons.RECYCLING` (mdi:recycle)
- `Déchets` -> `Icons.GENERAL_WASTE` (mdi:trash-can)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ritl_ca.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ritl_ca.py
@@ -1,13 +1,17 @@
-import logging
+from datetime import date as datetime_date
+from datetime import datetime as datetime_datetime
+from datetime import timedelta
 from typing import Dict, List
 
 import requests
+from icalendar import Calendar
 from waste_collection_schedule import Collection, Icons
 from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
-from waste_collection_schedule.service.ICS import ICS
 
 TITLE = "RITL"
-DESCRIPTION = "Source for Régie intermunicipale des Trois-Lacs (RITL) waste collection schedule."
+DESCRIPTION = (
+    "Source for Régie intermunicipale des Trois-Lacs (RITL) waste collection schedule."
+)
 URL = "https://ritl.ca/"
 COUNTRY = "ca"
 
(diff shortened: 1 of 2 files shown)


bbr111 (changes requested):
Thanks for putting this together, and for documenting the category/icon mapping so clearly!

Before this can be merged there's one blocking problem: when I run the four TEST_CASES against the live site, every one returns 0 entries. Digging in, the RITL "All-in-One Event Calendar" ICS export encodes all of the future collection dates as RDATE properties (I counted 493 RDATE values and 0 RRULE values in one category feed), and our shared ICS service (service/ICS.py, via icalevents) only expands RRULE-based recurrences — it does not expand RDATE-only events. So self._ics.convert(r.text) comes back empty even though there are 26 valid upcoming collection dates in the feed.

To fix it you'll need to expand the RDATE entries yourself rather than relying on ICS.convert — e.g. iterate the VEVENTs, read each CATEGORIES value to determine the waste type, and emit a Collection for every RDATE (and the DTSTART) that falls in range. Once a live test shows real dates for each TEST_CASE, please push again and I'll re-review.

A couple of smaller things while you're in there (non-blocking): drop the unused import logging, add a timeout= to the requests.get, and consider adding a PARAM_TRANSLATIONS dict alongside PARAM_DESCRIPTIONS.

bbr111 (approved):
Thanks for the quick turnaround! The RDATE handling is exactly what was needed — you're now parsing the feed with icalendar directly and expanding each RDATE (plus DTSTART) into a Collection. I re-ran all four TEST_CASES against the live site and every one returns real dates now (44 / 43 / 45 / 65 entries), correctly mapped to Compost / Recyclage / Déchets with the canonical Icons. The smaller notes are all addressed too: unused import logging gone, timeout=30 added to the request, and a PARAM_TRANSLATIONS dict is in place. Structural tests, ruff lint and format all pass. Looks great — approving.

PR #6775, opened 2026-07-06
Add Syndicat Emeraude (France-95) source

NightFlowed:
## Summary

Add Syndicat Emeraude (French) source
- Add rule-based source for Syndicat Emeraude (Val-d'Oise)
- Support all 17 communes and their available sectors
- Add automatic test cases for every commune/sector combination
- Include source owner

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `doc/source/syndicat_emeraude_fr.md` created for new sources


bbr111 (changes requested):
Thanks for the detailed work here — the rule engine and per-commune tables are clearly a lot of effort, and the doc + full TEST_CASES coverage are appreciated.

Unfortunately I can't merge this in its current form. The source doesn't fetch any live data: fetch() generates all dates from the hardcoded RULES tables, and DATA_URL is never actually requested. The project doesn't accept hardcoded schedules (see the "What NOT to do" section of CONTRIBUTING) — sources must pull from a live API, ICS feed, or webpage. The concern is real-world drift: as soon as the syndicate shifts a round for a holiday or changes the seasonal green-waste window, a static table is silently wrong with no way to self-correct.

Could you look at whether https://www.syndicat-emeraude.fr/collecte exposes the schedule data dynamically (an API/JSON/ICS endpoint, or HTML that can be parsed per commune/sector)? A quick check shows the page returns 403 to a plain request, so you'll likely need curl_cffi with impersonate="chrome" (see east_renfrewshire_gov_uk.py / south_ayrshire_gov_uk.py for the Cloudflare pattern). If the underlying data endpoint can be found and consumed at fetch time, I'd be happy to re-review.

Everything else — structure, COUNTRY, icons enum, translations, doc file, lint — looks good, so once the data is fetched live this should be close.

markvp (changes requested):
Thanks for keeping this moving, and for the care that went into the per-commune tables. I went digging into the provider to check whether @bbr111's request is actually achievable. There is good news, plus one thing that changes the plan.

**The good news: the real data is fetchable.** The `/collecte` page itself sits behind a JavaScript anti-DDoS challenge that returns 418 even with `curl_cffi` and `impersonate="chrome"`, and the challenge payload varies per request, so scraping that page is a dead end. The annual PDF calendars it links to, though, are served straight from the file system with no challenge at all:

```
https://www.syndicat-emeraude.fr/sites/default/files/media/downloads/Calendrier-2026-ANDILLY.pdf
```

That pattern (`Calendrier-<year>-<COMMUNE>.pdf`, commune uppercased with hyphens) resolves for the communes in your list, and page 1 carries the rules as extractable text, per sector. For Andilly 2026:

```
ORDURES MÉNAGÈRES   Collecte : Vendredi soir
PAPIERS/EMBALLAGES  Collecte : Mercredi matin
VERRE               Collecte : 2e lundi du mois
ENCOMBRANTS         Collecte : Mercredi 29 avril, Mercredi 30 septembre, Mercredi 30 décembre
VÉGÉTAUX            Collecte : Mercredi soir, du 1er au 31 janvier, du 1er mars au 30 novembre,
                    Mercredi 18 février et Mercredi 16 décembre
```

**The thing that changes the plan: the static tables are already out of date.** Andilly bulky waste is the clearest example:

* your rule `MonthlyRule("ME", 4, months=[4, 9, 12])` yields 22 April, 23 September and 23 December 2026
* the 2026 calendar says 29 April, 30 September and 30 December
* the 2025 calendar says 29 January, 30 July and 29 October

So both the months and the position within the month move each year. That is exactly the drift the "no hardcoded schedules" rule exists to prevent, and here it is already live in the diff rather than hypothetical.

Suggested direction, which keeps most of your work:

1. Keep the rule engine, but build the rules at fetch time by downloading and parsing page 1 of the commune's PDF. `pypdf` is already a dependency, and its `extract_text()` handles these files. The `Collecte :` lines map neatly onto your `WeeklyRule` / `MonthlyRule` / explicit-date cases.
2. Use `curl_cffi` for the PDF request, and fall back to the previous year's file when the new calendar is not published yet.
3. Read one-off dates such as the bulky-waste ones straight from the PDF instead of approximating them with an nth-weekday rule.

One small parsing gotcha: the calendar font renders digit zero as the letter "O", so you will see `3O septembre` and `2O26` in the extracted text and will need to normalise that.

Everything else (structure, `COUNTRY`, the `Icons` enum, translations, the doc page, lint and the structural test suite) is in good shape, so this really is just the data-sourcing layer. Happy to re-review once `fetch()` is pulling from the PDFs.

bbr111:
Following up on @markvp's review above — two things worth adding since they'd trip up a first implementation attempt:

**1. The commune-name-to-URL mapping isn't stable across years.** 2025 files use hyphens (`Calendrier-2025-SAINT-PRIX.pdf`, `ENGHIEN-LES-BAINS`, `SOISY-SOUS-MONTMORENCY`); the 2026 files for those *same* communes switched to URL-encoded spaces (`SAINT%20PRIX`, `ENGHIEN%20LES%20BAINS`, `SOISY%20SOUS%20MONTMORENCY`). A hyphen-only resolver would silently drop 5 of the 17 communes once 2026 becomes current. Two communes were also renamed outright: `montigny` is `MONTIGNY-LES-CORMEILLES` in 2025 but plain `MONTIGNY` in 2026; `le-plessis-bouchard` is `PLESSIS-BOUCHARD` in 2025 and `LE%20PLESSIS%20BOUCHARD` in 2026. Trying a small candidate list (hyphenated-upper, space-encoded-upper, plus a per-commune alias where needed) × (current year, previous year) resolves all 17.

**2. A missing PDF returns HTTP 418, not 404** — the same status the anti-DDoS challenge on /collecte returns. Treat any non-200 as "not found" rather than checking specifically for 404, or a typo'd/retired commune name will look like a protocol block instead of a bad URL.

@NightFlowed — are you still planning to pick this back up and rework `fetch()` to pull from the PDFs, or would you rather we close this and someone else takes a run at it? Just want to know where it stands.

bbr111:
@NightFlowed Thanks again for your work on this.

I'll close this PR for now, since we're moving development over to release/3.0.0.

Please feel free to reopen the discussion or submit a new PR against release/3.0.0 if you'd like to implement this as a fetching source instead of a rule-based one.

PR #6785, opened 2026-07-08
Add eThekwini Municipality (Durban) source

robtesch:
## Summary

- Adds a new source `durban_gov_za` for [eThekwini Municipality](https://www.durban.gov.za/page/refuse-collection-schedules) (Durban, South Africa).
- The municipality publishes refuse collection schedules as regional DOCX files (Outer West, North, North Central, South, South Central, Inner West) with suburbs listed under weekday columns. There is no public ICS/API feed, so the source uses a static area→weekday mapping (1,793 areas across all six regions) with provenance comments pointing at the official DOCX URLs.
- At runtime the source projects ~52 weeks locally: **General Waste** weekly on the mapped weekday; **Recycling** (orange bags) fortnightly on the same weekday. Orange-bag parity is not encoded in the DOCX tables, so users set `recycling_in_even_week` (same pattern as `harvey_wa_gov_au.py`). Recycling is skipped on South African public holidays; general waste is still collected.
- Adds `custom_components/waste_collection_schedule/waste_collection_schedule/source/durban_gov_za.py` and `doc/source/durban_gov_za.md`.
- Adds South Africa (`za`) to `COUNTRYCODES` in `update_docu_links.py` (required for the source to appear in generated docs).
- Declares `SOURCE_CODEOWNERS = ["@robtesch"]`. DOCX fetch/parse tooling to regenerate the mapping is maintained in my fork only and is **not** part of this PR.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `doc/source/durban_gov_za.md` created for new sources

## Test plan

```text
$ python test_sources.py -s durban_gov_za -l
Testing source durban_gov_za ...
  found 79 entries for Outer West Hillcrest
  found 81 entries for Outer West Drummond
  found 79 entries for North Umdloti
  found 79 entries for North Central Overport
  found 79 entries for South Central Montclair
  found 79 entries for Inner West Westville
  found 79 entries for South Umlazi
```

- [x] `ruff check --fix` / `ruff format` — clean on `durban_gov_za.py`
- [x] `python test_sources.py -s durban_gov_za -l` — all 7 TEST_CASES return non-empty collections
- [x] `doc/source/durban_gov_za.md` added
- [x] `update_docu_links.py` — `za` added to `COUNTRYCODES`


bbr111 (changes requested):
Thanks for the thorough writeup and the clean structural work — COUNTRY/za, the Icons enum usage, PARAM_* dicts, doc page and codeowners are all in good shape, and the structural test suite passes.

The blocker is the data-sourcing approach. This project does not accept sources that embed a hardcoded/frozen schedule — sources must fetch and parse their data live (API, ICS, or webpage/document) so they stay correct when the provider updates. Right now fetch() never makes a network request; it projects ~52 weeks from a static 1,793-area mapping that was snapshotted from the six DOCX files. When eThekwini re-issues those DOCX files (the URLs are already date-stamped /2025/09/18/), the source silently goes stale and needs a code change to fix.

Could you move the DOCX fetch/parse step (the tooling you mention keeping in your fork) into the source itself, so the region→area→weekday mapping is derived at runtime from the live DOCX files rather than embedded? The orange-bag recycling_in_even_week parity and the public-holiday handling can stay as-is. A couple of small things to fold in while you're at it: _south_africa_public_holidays annotates holidays as set[date] but builds a list, and please double-check whether general waste really is collected on public holidays (the doc page asserts it is).

Happy to re-review once it fetches live.

robtesch:
@bbr111 I'm happy to do that, but honestly I didn't think that adding heavy DOCX parsing as part of the source here made much sense since I'm not aware of any other sources that do that. Additionally there's no guarantee that eThekwini keeps the same format on their next update, and that would potentially require a parser update anyway. I'm more than happy to supply PRs whenever an update occurs. Would you be OK with that as an alternative?

For what it's worth, the tooling sits on my fork at https://github.com/robtesch/hacs_waste_collection_schedule/tree/tooling/durban-updater. 

EDIT: I've also added a weekly GitHub Action there that re-fetches the council DOCX links, regenerates the mapping, and opens an issue if anything changed (fork-only; not for upstream). That way I'm alerted when eThekwini publishes new date-stamped files rather than relying on someone noticing stale data. Happy to walk through it if useful.

bbr111:
@markvp @5ila5 could I get your take on this one? It's a borderline case for our "no hardcoded schedules" policy and I'd like a second opinion before deciding.

@robtesch — regarding your proposal to keep the static table and submit manual update PRs when eThekwini republishes: there's an alternative that avoids both the "heavy DOCX parsing in the source" concern and the manual-PR-per-update cycle. Since you already have the DOCX→schedule tooling working in your fork, you could run it yourself and publish the output as a hosted calendar (e.g. an ICS file on GitHub Pages, a Gist, or similar) that you regenerate on your own schedule. The `durban_gov_za` source would then just do a live fetch of that URL — no DOCX parsing inside this repo, no PR needed on every update, and you stay in control of when/how it's regenerated. You'd be the one responsible for keeping it current, which seems fair given you're also the codeowner.

That keeps the source itself simple and genuinely "live" (matching the project's fetch-from-a-live-source policy), while separating "who maintains the parsing logic" from "how the integration consumes it." Alternatively, for anyone who doesn't want to rely on your hosted feed, our generic [`static` source](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/static.md) already covers truly static/manual recurrence rules — but that's a fallback, not what I'd want baked into a dedicated per-municipality module.

Tagging the other maintainers since this pattern (community member owns the upstream parsing/hosting, module just fetches) comes up periodically and it'd help to have a consistent answer.

robtesch:
@bbr111 thanks very much for your consideration. I think that your proposal could definitely work, and may be the easiest path forward. Having said that, I'll wait for input from the other maintainers before starting work on that.

markvp:
Thanks for the careful escalation @bbr111. Agreed on the hosted-feed direction: a per-municipality module that ships a frozen 1,793-area table isn't something we can take, since it goes stale silently the next time eThekwini republishes (the date-stamped URLs tell us it will), and heavy DOCX parsing inside the source isn't a good fit either. Your ICS-feed suggestion is the right call and you've already laid it out for @robtesch, so I won't pile on. Happy to re-review once it fetches from a live feed.

robtesch:
@markvp thank you. I will withdraw this PR and work on that alternative implementation.

PR #6787, opened 2026-07-08
Expand ruff lint scope to source/, enable bugbear/comprehensions/pie/ruf rules

bbr111:
## Summary

`ruff.toml` previously only enabled `E, F, W, I` (mirroring the old flake8+isort setup), and the pre-commit `ruff` lint hook fully excluded the `~600`-file `source/` directory. This PR:

- Adds `B` (flake8-bugbear), `C4` (flake8-comprehensions), `PIE` (flake8-pie) and `RUF` (ruff-native) rule groups, ignoring `RUF001-003` (ambiguous-unicode) since they misfire heavily on non-English provider names/titles.
- Removes the `source/` exclude from the lint hook, and extends the hook's file pattern to also cover the two root-level scripts (`update_docu_links.py`, `default_translations.py`).
- Fixes everything the new rules flagged.

## Notable fixes

- **Real bugs**: several `source/` scrapers called `r.raise_for_status` **without parentheses** — the check never actually ran, so HTTP errors were silently swallowed (`B018`, caught in `eastdunbarton_gov_uk`, `melton_gov_uk`, `northlanarkshire_gov_uk`, `oldham_gov_uk`, `shellharbourwaste_com_au`, `sunderland_gov_uk`, `west_lindsey_gov_uk`).
- Two closures (`coventry_gov_uk`, `warrington_gov_uk`) read loop variables without binding them (`B023`); fixed by passing the values explicitly as parameters/lambda defaults instead of relying on late-binding closures.
- A first automated pass wrongly wanted to mark `config_flow.py`'s `_sources` as `ClassVar`, but it's actually rebound per-instance as a lazy-init cache — reverted that one to a plain mutable default with a `noqa: RUF012` and an explanatory comment.
- **Mechanical**: missing `raise ... from err/None` (`B904`), mutable class attributes annotated `ClassVar` where they're genuinely shared/read-only lookups (`RUF012`), mutable default arguments replaced with `None` sentinels (`B006`), unused loop/unpack variables prefixed with `_` (`B007`/`B020`/`RUF059`), comprehension/dict simplifications (`C4xx`, `RUF005`).

## Test plan

- [x] `python -m pytest tests/` — full suite passes
- [x] `ruff check` — clean across `custom_components/`, `tests/`, and the two root scripts
- [x] `mypy` (pre-commit hook scope) — before/after diff shows **zero newly introduced errors** (56 pre-existing errors, unchanged, confirmed via a side-by-side worktree comparison against `master`)
- [x] `pre-commit run --all-files` equivalent (ran via commit hooks) — all hooks pass

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -8,15 +8,13 @@ repos:
   - repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.15.17
     hooks:
-      # Linting (replaces flake8 + isort). Source modules carry pre-existing
-      # issues out of scope here, matching what flake8 excluded.
+      # Linting (replaces flake8 + isort).
       - id: ruff
         args: [--fix]
-        files: ^(custom_components|script|tests)/.+\.py$
-        exclude: ^custom_components/waste_collection_schedule/waste_collection_schedule/source/
+        files: ^(custom_components|script|tests)/.+\.py$|^[^/]+\.py$
       # Formatting (replaces black). Runs on the full tree, like black did.
       - id: ruff-format
-        files: ^(custom_components|script|tests)/.+\.py$
+        files: ^(custom_components|script|tests)/.+\.py$|^[^/]+\.py$
   - repo: https://github.com/codespell-project/codespell
     rev: v2.4.2
     hooks:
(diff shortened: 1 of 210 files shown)


markvp (approved):
Thanks for taking this on, it's a big and careful sweep. Expanding the lint scope to source/ and turning on bugbear/comprehensions/pie/ruf is well worth it, and it flushed out some real bugs in the process (the bare raise_for_status calls and the late-binding closures are nice catches). I ran the structural suite (tests/test_source_components.py, 9 passed), ruff check and ruff format --check across the full scope and everything is clean, and the config-flow ClassVar/noqa: RUF012 reasoning checks out. Approving.

PR #6802, opened 2026-07-09
Expand ruff to UP, ISC, RET rule sets; drop redundant pyupgrade hook

bbr111:
## Summary

Follow-up to #6787 (which enabled B/C4/PIE/RUF across `source/`). This adds three more rule sets that resolved **fully clean** across the whole tree with zero manual/behavior-risking fixes needed:

- **UP** (pyupgrade) — `typing.Dict/List/Optional/Tuple` → builtin generics / PEP 604 unions, `dict.keys()` iteration simplification, etc. Replaces the standalone `pyupgrade` pre-commit hook (now redundant — one tool, one target-version).
- **ISC** (implicit string concatenation)
- **RET** (flake8-return), except `RET503`/`RET504` — see below.

Also fixes a real config bug: `ruff.toml` declared `target-version = "py313"`, but CI (`validate.yaml`) actually runs the lint/test suite on Python 3.10. Since UP's fixes are gated on `target-version`, this could have suggested 3.11+/3.12+-only syntax that breaks on the Python version CI actually runs. Set to `py310` to match CI.

### Evaluated but deliberately **not** included this round

`SIM`, `PERF`, and `T20` (print) were tested too. Almost every hit under those rule sets has **no automatic fix** and would mean hand-rewriting hundreds of loops/prints across ~600 `source/` files for what's mostly a stylistic gain (e.g. `PERF401` alone: 277 manual loop→comprehension rewrites; `T201`: 180 `print()` calls, several in `wizard/` helpers that likely want proper logging instead, not a search-and-replace). That's a much larger and riskier undertaking than this PR's scope — worth a separate, deliberate cleanup pass rather than bundling here.

`RET503`/`RET504` hit the same wall (64 combined instances, no auto-fix for either) and are ignored for the same reason.

`ERA` (commented-out code detector) was also evaluated separately and intentionally left out: most of its ~284 hits are genuine dead code, but it has a real false-positive rate on prose comments describing string formats (e.g. `# part is "f.x.y = 'z'"` gets misread as an assignment) — good candidate for a dedicated cleanup PR, not a standing lint rule bundled with this one.

## Test plan

- [x] `ruff check .` — all checks pass with the new `select`
- [x] `ruff format --check .` — all files already formatted
- [x] `python -m pytest tests/` — 9/9 pass
- [x] `mypy` over the non-source integration layer — same 56 pre-existing errors as before (no new ones introduced)
- [x] Spot-checked live `TEST_CASES` runs (`test_sources.py`) for a few of the more heavily-touched source files (`ritl_ca`, `ccc_tas_gov_au`, `amsa_it`) — behave identically to before


markvp (approved):
Nicely scoped follow-up, and thank you for the thorough writeup. The target-version correction to py310 to match CI is a good catch on its own. I checked the config change and spot-checked the mechanical fixes (the RET else/elif de-indentations and the UP typing rewrites) and everything is behaviour-preserving; ruff check, ruff format --check and the test suite all pass clean here too. Agree with deferring SIM/PERF/T20/ERA and RET503/504 to separate passes. Approving.

PR #6819, opened 2026-07-11
feat(calendar): expose collection description and location on calendar events

ItsMly:
## Summary

`Collection` carries optional `description` and `location` fields (populated by e.g. the `ics` source, `abfall_io`, and the new `mywastewatcher_de`), but the calendar entity dropped them when converting to `CalendarEvent`. This passes both through, so e.g. mobile hazardous-waste collection stops/times and holiday reschedule notes ("Vorverlegung") appear in the calendar event details.

Both fields default to `None` on `CollectionBase` (and `CollectionGroup` aggregates them), so events from sources without this data are unchanged. Verified live on a Home Assistant instance.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other

## Checklist

- [ ] `doc/source/<name>.md` created for new sources — n/a
- [ ] TEST_CASES use real, publicly accessible addresses (not my own) — n/a

### DIFF
--- a/custom_components/waste_collection_schedule/calendar.py
+++ b/custom_components/waste_collection_schedule/calendar.py
@@ -118,6 +118,8 @@ class WasteCollectionCalendar(CalendarEntity):
             summary=collection.type,
             start=collection.date,
             end=collection.date + timedelta(days=1),
+            description=collection.description,
+            location=collection.location,
             uid=uuid.uuid4(),
         )
 


bbr111 (approved):
Thanks Emily. Clean, well-scoped change — `description` and `location` already exist on `Collection` (defaulting to `None`) and are aggregated by `CollectionGroup`, so events from sources that don't set them are unaffected. `CalendarEvent` accepts both kwargs, structural tests pass (9 passed) and ruff is clean. Approving as-is.

cc @markvp — flagging this for the `BaseSource` architecture work: right now `BaseSource` only constructs `Collection(date=..., waste_type=...)` and doesn't populate `description`/`location`. Now that the calendar entity surfaces both fields, is this something worth incorporating into the `BaseSource` pipeline (e.g. exposing them via the parser/transformer output) so sources migrated to the new architecture aren't missing this compared to legacy sources?

PR #6860, opened 2026-07-15
fix failures when there is a collection entry with a blank description

### LINKED ISSUE #6859: [Source Defect]: 1Coast source fails when there is an entry that is just a dash (-)
### Source Name

1coast_com_au

### What's happening?

When adding integration to HA, I am getting a "list index out of range" error. 

Image

### Error Logs

```Shell
N/A
```

### Source Configuration

```YAML
address: 56 EVERGLADES CR, WOY WOY. CENTRAL COAST 2256
```

### Checklist

blueciii:
## Summary

This aims to fix the "list index out of range" error when setting up the integration, this [bug](https://github.com/mampfes/hacs_waste_collection_schedule/issues/6859).

## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [X] `python test_sources.py -s 1coast_com_au -l` returns entries

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/1coast_com_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/1coast_com_au.py
@@ -1,3 +1,4 @@
+import re
 from datetime import date, datetime
 
 import requests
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for the fix — pairing each bin name with its own date block is the right approach and cleanly resolves the "list index out of range" error from #6859 (and the Copilot alignment concern). I made one small tweak: restored icon=ICON_MAP.get(bin_name) on the HTML-parsed entries, which the rewrite had dropped (only matters on the ICS-404 fallback path). Structural tests pass. Looks good to merge.

PR #6825, opened 2026-07-13
Fix Sutherland Shire Council source: adapt to new services-map widget

bbr111:
## Summary
- Sutherland Shire Council replaced the ASP.NET suburb/street/house-number
  dropdown form the source relied on with a JS-driven "services map"
  address-search widget. The old scraper can no longer find the suburb
  dropdown at all (#6813).
- Reverse engineered the widget's production JS bundle to find its real
  data flow: address → lat/lng (Google Places, client-side only) → a
  public ArcGIS query on the council's own GIS server
  (`geoserver.ssc.nsw.gov.au/arcgis/rest/services/WebOnline/MapServer/4/query`)
  returning `DAY_OF_WEEK` and `ZONE_NO` for the point, then a
  fortnight-parity calculation (ISO week number vs. zone) to decide
  Recycling vs. Garden Waste.
- Replaced the unusable Google Places step with the Esri World geocoder
  (`geocode.arcgis.com`), the same no-API-key pattern already used by
  `cityofparramatta_nsw_gov_au.py` in this repo, then query the council's
  own zone layer and reproduce the exact parity rule extracted from the
  site's JS.
- Simplified `suburb`/`street`/`house_number` argument semantics — they no
  longer need to match an exact dropdown value/case, since they're now
  geocoded rather than pattern-matched against ASP.NET dropdown state.

## Test plan
- [x] `ruff check --fix` / `ruff format` — no changes needed
- [x] `python -m pytest tests/test_source_components.py -q` — 9 passed
- [x] `python test_sources.py -s sutherlandshire_nsw_gov_au -l` — geocoding
      step verified live and returns correct coordinates for the test
      address; the final council ArcGIS zone-query call could not be
      reached from the sandbox used to prepare this PR (network egress
      restricted to a small allowlist of cloud domains, this host wasn't
      on it) — **please re-run the live test from an unrestricted network
      before merging** to confirm the zone-query endpoint, field names
      (`DAY_OF_WEEK`, `ZONE_NO`) and parity logic end-to-end.

PR #6872, opened 2026-07-17
Fix sutherlandshire_nsw_gov_au: council replaced lookup form; query public GIS layer

chris-clearvue:
## Problem

The `waste-information-booklet` page this source scraped was removed from the council website (it now 301s to a map-based page with no ASP.NET form), and sutherlandshire.nsw.gov.au also now rejects plain library HTTP clients by TLS fingerprint — every fetch fails with `403 Client Error: Forbidden`.

## Fix

The council's new ["When is my bin collected?" page](https://www.sutherlandshire.nsw.gov.au/living-here/waste-and-recycling/bin-collection) is backed by a public ArcGIS layer (`geoserver.ssc.nsw.gov.au/arcgis/rest/services/WebOnline/MapServer/4`) with one feature per property carrying `DAY_OF_WEEK`, `ZONE_NO` and `RECYCLE_PICKUP`. This rewrite queries that layer by address attributes instead:

- exact `situ1`/`situ2` match, prefix fallback for unit numbers, and `SourceArgumentNotFoundWithSuggestions` with the street's known house numbers when the house number doesn't match
- schedule generation (weekly garbage; recycling/garden waste alternating fortnights by zone parity) is unchanged from the previous implementation
- properties flagged `RECYCLE_PICKUP = Weekly` (some unit blocks) additionally get weekly recycling
- args are unchanged, so existing configs keep working; second TEST_CASE added

## Verified

Both TEST_CASES fetch successfully (`5 Cleveland Place BONNET BAY` → Monday zone 2; `20 Waratah Street CRONULLA` → Wednesday zone 2, weekly recycling), and a bogus house number returns suggestions for the street.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sutherlandshire_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sutherlandshire_nsw_gov_au.py
@@ -1,13 +1,13 @@
 from __future__ import annotations
 
-import re
-import urllib.parse
 from datetime import date, timedelta
 
 import requests
-from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
-from waste_collection_schedule.exceptions import SourceArgumentNotFound
+from waste_collection_schedule.exceptions import (
+    SourceArgumentNotFound,
+    SourceArgumentNotFoundWithSuggestions,
+)
 
 TITLE = "Sutherland Shire Council"
 DESCRIPTION = "Source for Sutherland Shire Council, NSW, Australia."
@@ -17,6 +17,11 @@ TEST_CASES = {
         "suburb": "BONNET BAY",
         "street": "Cleveland Place",
         "house_number": "5",
+    },
+    "20 Waratah Street, CRONULLA": {
+        "suburb": "CRONULLA",
+        "street": "Waratah Street",
+        "house_number": "20",
     },
 }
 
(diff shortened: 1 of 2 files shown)


bbr111 (changes requested):
Thanks for this, and for the thorough writeup — the attribute-based GIS query, the house-number suggestions, and the weekly-recycling handling for unit blocks are all nice touches.

The blocker is timing: this branch was cut just before #6825 was merged (2026-07-17), and #6825 already fixes this exact breakage against the *same* ArcGIS layer (WebOnline/MapServer/4) — it geocodes the address to a point and does a spatial query, and it added COUNTRY = "au" plus a calendar-verified fortnight anchor (_REFERENCE_MONDAY = date(2026, 7, 13), checked against the council's 2026/27 Zone 1 & 2 calendars). As it stands this PR conflicts with / would revert that work, and it drops COUNTRY.

Could you please rebase onto current master and reconcile with #6825? Specifically:
- Decide whether your attribute-query approach should replace the merged geocoding approach, or whether the suggestions + weekly-recycling-for-units pieces can be layered on top of what's already there.
- Keep COUNTRY = "au".
- Reconcile the zone/fortnight logic: master uses _REFERENCE_MONDAY = date(2026, 7, 13) verified against the council calendars, whereas this PR uses date(2024, 1, 1) + _ZONE_OFFSETS. These can disagree on which weeks are Recycling vs Garden Waste, so please confirm the alternation against the published Zone 1/Zone 2 calendars.

Once rebased and the recycling/garden alternation is verified, happy to take another look.

chris-clearvue:
Closing in favour of #6890 — identical change re-raised from my main account (CRZTFR).

PR #6873, opened 2026-07-17
Improve wanneroo_wa_gov_au address matching for human-formatted input

chris-clearvue:
## Problem

The City of Wanneroo register stores addresses in a rigid canonical format — abbreviated street type + uppercase suburb, e.g. `23 Bakana LP LANDSDALE`. The register's search endpoint returns **zero results** for anything else: `23 Bakana Loop`, `23 Bakana Loop, Landsdale`, or any spelled-out street type all raised `SourceArgumentNotFound` with no suggestions. Only users who type the exact canonical string ever get a schedule.

## Fix

Two contained changes to this source:

1. **Abbreviation-aware matching** — expand street-type abbreviations on both sides before comparing, so `23 Bakana Loop, Landsdale` now matches the canonical `23 Bakana LP LANDSDALE` entry directly.
2. **Reduced-query retry** — when the search returns nothing, retry with a shortened query (`23 Bakana`), which the register answers with its candidate list. Those candidates either match after normalisation (direct success) or feed the existing `SourceArgumentNotFoundWithSuggestions` flow instead of a dead end.

## Verified

- All three canonical TEST_CASES unchanged and passing
- `23 Bakana Loop, Landsdale` and `23 Bakana Loop` (previously hard failures) now fetch the correct schedule

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/wanneroo_wa_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/wanneroo_wa_gov_au.py
@@ -57,6 +57,67 @@ WEEKDAYS = [
     "sunday",
 ]
 
+# The council's register stores abbreviated street types ("23 Bakana LP
+# LANDSDALE"). Users typically spell them out; expand both sides before
+# comparing so "23 Bakana Loop, Landsdale" matches the canonical entry.
+STREET_TYPE_ABBREVIATIONS = {
+    "ST": "STREET",
+    "RD": "ROAD",
+    "DR": "DRIVE",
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for this — clean, well-scoped fix. The abbreviation normalisation and the capped reduced-query retry are both self-contained and nicely commented. Structural test suite passes (10/10), ruff lint and format are clean, and no generated files are touched. Couldn't live-hit the Wanneroo API from my environment, but the logic checks out on trace. Approving.

markvp:
I wonder if this could be expanded into something reusable by other sources...

PR #6876, opened 2026-07-17
fix(config): handle YAML sexagesimal parsing of fetch_time/day_switch_time

### LINKED ISSUE #5065: [Bug]: fetch_time over 09:59 in configure.yaml does not work
### I Have A Problem With:

The integration in general

### What's Your Problem

If I configure a "fetch_time" over 09:59 in configure.yaml it does not work.
Example:
If I configure
   fetch_time: 10:00
I will get the following message (in Verify and restart):

Invalid config for 'waste_collection_schedule' at configuration.yaml, line 38: Invalid time specified: 600 for dictionary value 'waste_collection_schedule->fetch_time', got 600

### Source (if relevant)

_No response_

### Logs

```Shell

```

### Relevant Configuration

```YAML

```

### Checklist Source Error

### Checklist Sensor Error

### Required

bbr111:
## Summary
- Unquoted `fetch_time` / `day_switch_time` values in `configuration.yaml` whose hour is 10-23 (e.g. `fetch_time: 10:00`) are silently mis-parsed by PyYAML's YAML 1.1 "sexagesimal" (base-60) integer resolver into an int (`10:00` -> `600`), which then fails Home Assistant's `cv.time` validator with `Invalid time specified: 600`.
- Times with a leading zero (`00:xx`-`09:xx`) are unaffected, which matches the reporter's exact symptom.
- Added a small wrapper validator (`_time_value`) in `init_yaml.py` that detects this collision (int in range 60-1439 for `H:MM`, or 3600-86399 for `H:MM:SS` — these ranges never overlap, so the original format is always recoverable) and reconstructs the correct `HH:MM[:SS]` string before validating with `cv.time`.
- Quoted strings, and the existing defaults, continue to work exactly as before.

Fixes #5065

## Test plan
- [x] `ruff check --fix` / `ruff format` on the changed file
- [x] Manually validated the schema against `09:59`, `10:00`, `23:59`, `00:30`, `1:00`, `1:30:00`, `9:05`, a quoted `"10:00"`, and the no-value default — all resolve to the correct `datetime.time`
- [x] `python -m pytest tests/ -q` passes (no regressions)

### DIFF
--- a/custom_components/waste_collection_schedule/init_yaml.py
+++ b/custom_components/waste_collection_schedule/init_yaml.py
@@ -41,7 +41,7 @@ def _time_value(value):
     """
     if isinstance(value, int) and not isinstance(value, bool):
         if 60 <= value <= 1439:
-            value = f"{value // 60:02d}:{value % 60:02d}:00"
+            value = f"{value // 60:02d}:{value % 60:02d}"
         elif 3600 <= value <= 86399:
             hours, remainder = divmod(value, 3600)
             minutes, seconds = divmod(remainder, 60)
--- a/custom_components/waste_collection_schedule/init_yaml.py
+++ b/custom_components/waste_collection_schedule/init_yaml.py
@@ -21,6 +21,34 @@ from waste_collection_schedule import Customize  # type: ignore # isort:skip # n
 
 _LOGGER = logging.getLogger(__name__)
 
+
+def _time_value(value):
+    """Validate a time-of-day config value.
+
+    Works around a YAML 1.1 quirk: PyYAML's implicit "sexagesimal" (base60)
+    resolver turns an *unquoted* ``HH:MM`` value such as ``10:00`` into the
+    integer ``600`` (10 * 60 + 0) instead of the string ``"10:00"``, which
+    then fails Home Assistant's ``cv.time`` validator. Times with a leading
+    zero (e.g. ``09:59``) are unaffected because the resolver requires the
+    first digit to be 1-9, so this only bites hours 10-23.
+
+    If we receive such an integer, reconstruct the original ``HH:MM`` (or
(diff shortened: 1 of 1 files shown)


bbr111:
@markvp 
do we want to change things like this?
or ask the user just to use yaml correctly?

markvp:
Yeah it's an interesting one. Is it the incorrect usage of yaml or a design choice/flaw in pyyaml. Should it be resolved via quoting the times, changing/fixing pyyaml, or absorbing the cost of checking and fixing every time value here. I'm not sure. Let me think about it

bbr111:
@markvp You pointed it out exactly: the real question is where this should be handled.

After looking into it, this is not simply invalid YAML. PyYAML still follows YAML 1.1 here, where an unquoted value like 10:00 is parsed as the sexagesimal integer 600. YAML 1.2 removed this behavior, but changing the parser is outside the scope of this integration because Home Assistant has already parsed the configuration before it reaches our schema.

For that reason, I would prefer the documentation approach: clearly document that time values must be quoted, for example:

fetch_time: "10:00"
day_switch_time: "18:30"

This keeps the integration simple and avoids silently reconstructing values after the original YAML representation has already been lost. A more helpful validation message could still be added later, but I would not work around the parser behavior here.

markvp:
I agree, we should probably also see if we support a fix upstream. Is there a newer 1.2 compatible version of pyyaml that HA could upgrade to?

bbr111:
@markvp pyyaml is still on 1.1 and the PR for 1.2 is not merged right now.

markvp:
Can you link the original issue to the pyyaml 1.2 PR, I think that's about all we can do then.

bbr111:
https://github.com/yaml/pyyaml/pull/555
there are more than one.

PR #6890, opened 2026-07-18
Fix sutherlandshire_nsw_gov_au: council replaced lookup form; query public GIS layer

CRZTFR:
Re-raise of #6872 from my main account (same commit; #6872 closed).

## Problem

The `waste-information-booklet` page this source scraped was removed from the council website (it now 301s to a map-based page with no ASP.NET form), and sutherlandshire.nsw.gov.au also now rejects plain library HTTP clients by TLS fingerprint — every fetch fails with `403 Client Error: Forbidden`.

## Fix

The council's new ["When is my bin collected?" page](https://www.sutherlandshire.nsw.gov.au/living-here/waste-and-recycling/bin-collection) is backed by a public ArcGIS layer (`geoserver.ssc.nsw.gov.au/arcgis/rest/services/WebOnline/MapServer/4`) with one feature per property carrying `DAY_OF_WEEK`, `ZONE_NO` and `RECYCLE_PICKUP`. This rewrite queries that layer by address attributes instead:

- exact `situ1`/`situ2` match, prefix fallback for unit numbers, and `SourceArgumentNotFoundWithSuggestions` with the street's known house numbers when the house number doesn't match
- schedule generation (weekly garbage; recycling/garden waste alternating fortnights by zone parity) is unchanged from the previous implementation
- properties flagged `RECYCLE_PICKUP = Weekly` (some unit blocks) additionally get weekly recycling
- args are unchanged, so existing configs keep working; second TEST_CASE added

## Verified

Both TEST_CASES fetch successfully (`5 Cleveland Place BONNET BAY` → Monday zone 2; `20 Waratah Street CRONULLA` → Wednesday zone 2, weekly recycling), and a bogus house number returns suggestions for the street.


bbr111 (approved):
Thanks for the rewrite to the public GIS layer — clean approach and the fallback-to-suggestions logic is nice. I made one small tweak: the rewrite had dropped the COUNTRY = "au" constant, which would have quietly orphaned the source out of the generated README/sources.json listing, so I restored it. Structural tests pass. Approving.

PR #6893, opened 2026-07-18
fix(config_flow): stop duplicating new-style sources in the picker

bbr111:
## Summary
- `_async_setup_sources()` merged `_discover_new_style_sources()`'s runtime scan into the `sources.json`-derived list with a blind `.extend()`, with no check for ids already present.
- `sources.json` is already regenerated to include new-style (`PARAMS`-based `BaseSource`) sources, so every source that's been migrated to the new pipeline was appended a second time and showed up twice in the UI country → source picker (e.g. `abfall_neunkirchen_siegerland_de` under Germany).
- Verified against this branch: 242 sources currently use `PARAMS`, and all 242 were duplicated before this fix.

## Fix (updated per discussion below)
Per markvp: this runtime scan predates `sources.json` being regenerated for new-style sources through CI, and is now fully redundant with it. Removed `_discover_new_style_sources()` and `_SOURCE_DIR` entirely — `sources.json` is now the sole source of truth for `_async_setup_sources()`. Simpler than a dedupe fallback, and avoids importing every source module under `source/` on every config-flow init.

## Test plan
- [x] `ruff check` / `ruff format` clean
- [x] `python -m pytest tests/test_source_components.py` passes (10/10)
- [x] `mypy` — no new errors vs. unmodified branch (same 22 pre-existing errors, unrelated to this file's touched lines)
- [x] Reproduced the original bug and the fix locally by replicating `_load_sources()` + `_discover_new_style_sources()` + the merge logic outside HA: confirmed `abfall_neunkirchen_siegerland_de` (and all 242 currently-migrated new-style sources) go from 2 picker entries to 1 with sources.json alone


bbr111:
One thing worth deciding explicitly before this merges: should `_discover_new_style_sources()` go away entirely, with the picker relying on `sources.json` alone?

Arguments for removing it now rather than leaving it as alpha scaffolding:
- It's currently 100% redundant — all 242 `PARAMS`-based sources are already in `sources.json`, so the fallback path never actually contributes an entry in practice.
- It costs real startup time: importing every `.py` file under `source/` on each config-flow init just to check for `PARAMS`, only to find nothing new.
- Less code/logic to keep in sync = one less place for this exact class of bug to recur.

Argument for keeping it (why I left it in for now): it's the only thing that makes a source in an *unmerged* PR selectable in a live HA instance's UI, since `sources.json` is only regenerated by CI post-merge and contributors/maintainers are asked not to run that generator themselves in a branch. Pulling it would mean testing a new source's config flow end-to-end has to wait until after merge — a real loss during active migration to the new pipeline.

I left it in behind a dedupe + a comment flagging it as alpha-only, to be dropped once the pipeline migration settles down and pre-merge UI testing matters less. But if the team would rather cut it now and accept "regenerate `sources.json` locally to UI-test a new source" as the workflow, I can simplify this PR to just delete `_discover_new_style_sources()` and the merge step outright — happy to go either way, wanted to flag the tradeoff rather than decide unilaterally.

markvp:
Yeah I think remove. I think I built it before I was merging them through ci

bbr111:
Makes sense, thanks — removed `_discover_new_style_sources()` (and the now-unused `_SOURCE_DIR`) entirely in the latest commit. `_async_setup_sources()` now relies on `sources.json` alone.

PR #6920, opened 2026-07-19
refactor Logan City Council to use OpenCities API (same as MyLogan tool)

CpuID:
## Summary

Refactors Logan City Council to use the same backend as https://www.logan.qld.gov.au/MyLogan uses today (OpenCities)

Uses `curl_cffi` with Chrome impersonation to reliably pass Akamai's checks

Caches the geolocation-id in-memory so its not called every day (less API calls), only queried once on restart + initial setup.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [x] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/logan_qld_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/logan_qld_gov_au.py
@@ -1,127 +1,159 @@
-import json
-import urllib.parse
-from datetime import date, timedelta
+from datetime import datetime
 
-import requests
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for the thorough refactor — this is a clean migration to the OpenCities backend. curl_cffi with Chrome impersonation, in-memory caching of the geolocation id, Icons enum usage, and English-only PARAM_TRANSLATIONS/PARAM_DESCRIPTIONS all look good. Structural tests pass (10 passed) and a live run returns correct schedules for all three TEST_CASES, including green waste for Rochedale South. Approving as-is.

PR #6921, opened 2026-07-19
Fix 7 wrong council website domains in impactapps_com_au source

deal0f49-21:
## Summary
- Brisbane City Council's \`website\` pointed to a \`.nsw.gov.au\` domain — Brisbane is in Queensland, real domain is \`brisbane.qld.gov.au\`
- Campbelltown, Canada Bay, Cumberland, Port Stephens, and Wollongong councils' \`website\` all pointed to \`.vic.gov.au\` domains — all five are actually in NSW
- Lithgow City Council's \`website\` (\`lithgow.nsw.gov.au\`) doesn't resolve at all — the council's real site is \`council.lithgow.com\`

The \`url\` field (used for the actual waste-info.com.au API calls this source makes) is unaffected — only \`website\`, the reference link to each council's own official site, was wrong.

## How found
Building a per-scan "data sourced from X" attribution feature in a downstream app that surfaces \`website\` to end users. Verified every corrected domain resolves and is the right council (direct HTTP check, cross-referenced against Wikipedia's infobox for a couple that didn't respond directly from my network) before fixing.

## Test plan
- [x] File still parses (valid Python)
- [x] Diff is minimal — exactly the 7 wrong \`website\` values changed, nothing else touched
- [x] Confirmed all 7 corrected domains resolve and match the real council

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/impactapps_com_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/impactapps_com_au.py
@@ -217,7 +217,7 @@ SERVICE_MAP = [  # supported calendars can be found at https://calendars.impacta
     {
         "name": "Brisbane City Council",
         "url": "https://brisbane.waste-info.com.au",
-        "website": "https://www.brisbane.nsw.gov.au",
+        "website": "https://www.brisbane.qld.gov.au",
     },
     {
         "name": "Burwood City Council",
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for the careful data fix — all seven corrected domains check out (Brisbane is QLD, the five .vic.gov.au entries are all NSW councils, and Lithgow's real site is council.lithgow.com). The change is minimal and only touches the website reference links in SERVICE_MAP, which feed EXTRA_INFO(); the url values used for the live API calls are untouched. Lint, format, and the structural test suite all pass. Approving as-is.

PR #6946, opened 2026-07-19
refactor: migrate 6 OpenCities sources onto shared service/OpenCities.py (batch 1/5)

bbr111:
## Summary

Batch 1 of the phased OpenCities/MyArea consolidation started in #6934: migrates the 6 easiest sources onto `OpenCitiesClient` instead of each hand-rolling the search → geolocation-id → wasteservices flow.

- `kiama_nsw_gov_au.py`, `shoalhaven_nsw_gov_au.py` — geolocation-id-only councils (no search step), exercise `fetch(geolocation_id=...)` in isolation.
- `ballina_nsw_gov_au.py`, `lanecove_nsw_gov_au.py`, `willoughby_nsw_gov_au.py`, `cassowarycoast_qld_gov_au.py` — address-search councils, introduce `search_fuzzy`, `max_results`, `page_link`, custom `headers`, and curl_cffi one at a time.

### Client fixes/extensions found while migrating these six

- **Nested-block double-counting bug**: some council HTML wraps a matching `<article>` inside a matching `div.waste-services-result` (confirmed live on Ballina and Cassowary Coast) — the client's original `"article, div.waste-services-result"` selector matched both, double-counting every entry. Fixed by only keeping the outermost match of a nested pair (`OpenCitiesClient._select_top_level`), with a regression test.
- **`OpenCitiesConfig.require_date_precise`** (new field): Shoalhaven's wasteservices response mixes precise-date blocks with vague/recurring ones in the same payload and only wants the former — a genuine behavioral difference from every other migrated council, not just style, so it's a config flag rather than hardcoded logic.
- **lanecove/willoughby now need curl_cffi**: confirmed via a raw HTTP replay of each source's *pre-migration* code that both now 403 under plain `requests` (Akamai/Incapsula), so this migration bundles the `curl_cffi` fix per the project's existing "403 → curl_cffi" convention rather than leaving them broken.

### Note on kiama

Both of `kiama_nsw_gov_au.py`'s hardcoded `TEST_CASES` geolocation IDs return `"No collection services found for this address"` from the live API today — verified by replaying the exact pre-migration HTTP request, so this is pre-existing stale test-fixture data, not something this migration caused. Not fixed here (would require a fresh real address, out of scope for a refactor PR).

## Type of change

- [ ] New source
- [x] Bug fix / source fix (nested-block double-counting; lanecove/willoughby curl_cffi 403 fix)
- [ ] Documentation update
- [x] Other (migration onto shared service/OpenCities.py client)

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (27 passed)
- [x] `ruff check --fix` / `ruff format` / `mypy --ignore-missing-imports --explicit-package-bases` clean (pre-commit hooks pass)
- [x] Live-tested each of the 6 sources via `test_sources.py -s <name> -l` against the real API
- [x] `TITLE`/`URL`/`COUNTRY`/`TEST_CASES`/`HOW_TO_GET_ARGUMENTS_DESCRIPTION`/`PARAM_*` unchanged; `doc/source/*.md` untouched

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
@@ -88,6 +88,14 @@ class OpenCitiesConfig:
     all, where the first result is the only option.
     """
 
+    require_date_precise: bool = False
+    """
+    If True, only wasteservices blocks additionally carrying a
+    ``date-precise`` CSS class are parsed, skipping vague/recurring entries
+    with no concrete next date (some council deployments, e.g. Shoalhaven,
+    mix both kinds of block in the same response).
+    """
+
 
 class OpenCitiesClient:
     def __init__(self, config: OpenCitiesConfig) -> None:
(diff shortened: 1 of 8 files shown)


bbr111:
kiama testcases now has output.

PR #6951, opened 2026-07-19
refactor: migrate 6 more OpenCities sources onto shared client (batch 2/5)

bbr111:
## Summary

Batch 2 of the phased OpenCities/MyArea consolidation (batch 0: #6934, batch 1: #6946): migrates `hobartcity_com_au.py`, `moorabool_vic_gov_au.py`, `banyule_vic_gov_au.py`, `cambridge_wa_gov_au.py`, `mildura_vic_gov_au.py`, and `logan_qld_gov_au.py` (itself only recently migrated to OpenCities in #6920) onto `OpenCitiesClient`.

These six were picked because between them they exercise the client-side variance not yet covered by batch 1: curl_cffi, the `geolocation_id` bypass kwarg alongside address search, a non-default `date_format`, and hand-rolled caching/ambiguous-address-match logic (hobartcity, moorabool, logan) that the shared client's defaults now subsume.

### Client extensions added while migrating (found via live-testing, not assumed upfront)

- **`OpenCitiesConfig.strip_type_suffixes`** — moorabool strips a `"collection"` suffix some labels carry.
- **`OpenCitiesConfig.warm_up_before`** (`"search"` or `"wasteservices"`) — banyule and cambridge fire their warm-up GET *after* resolving a geolocation id (via search or the direct bypass) and *before* the wasteservices call. The client originally always warmed up before search; this needed to be configurable, not hardcoded, since different councils need different ordering.
- **Explicit `Accept: application/json` header for banyule/cambridge** — their search endpoints silently return XML instead of JSON without it (confirmed by replaying the raw HTTP request both with and without the header). Added via the existing `headers` config field rather than extending `search_response_format`, since it's an Accept-header quirk rather than a deployment that only emits XML.

### Note on Cambridge's "duplicate" entry

`Cambridge Library` (`99 The Boulevard, FLOREAT 6014`) returns 4 entries with two identical "General Waste / Wed 22/7/2026" rows. Verified via a raw HTTP replay that the API's own `wasteservices` response genuinely contains two identical `<article>` blocks for that address — this is upstream duplicate data, not a client regression (pre-migration code, which also just iterated every `<article>`, would have produced the same duplicate).

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other (migration onto shared service/OpenCities.py client + client extensions)

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (29 passed)
- [x] `ruff check --fix` / `ruff format` / `mypy --ignore-missing-imports --explicit-package-bases` clean (pre-commit hooks pass)
- [x] Live-tested each of the 6 sources via `test_sources.py -s <name> -l` against the real API
- [x] `TITLE`/`URL`/`COUNTRY`/`TEST_CASES`/`HOW_TO_GET_ARGUMENTS_DESCRIPTION`/`PARAM_*` unchanged; `doc/source/*.md` untouched

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
@@ -67,6 +67,14 @@ class OpenCitiesConfig:
     warm_up_url: str | None = None
     """A page fetched once, lazily, before the first API request (session/cookie priming)."""
 
+    warm_up_before: Literal["search", "wasteservices"] = "search"
+    """
+    When the (one-off) warm-up GET fires: before the address search
+    (default), or after it — right before the wasteservices call, which
+    also covers the direct ``geolocation_id`` bypass path where no search
+    ever happens.
+    """
+
     date_format: str = DEFAULT_DATE_FORMAT
 
     icon_keywords: dict[str, Any] | None = None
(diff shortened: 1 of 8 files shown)


bbr111:
local tests all positiv.

Issue #6981, opened 2026-07-20
[Bug]: Fetch failed for polish users


DeveloperSzkodziu:
### Describe the bug

Hello,
Since tester I can observe this issue in logs. Other users from Poland also raising this issue. Could you please help?

Rejestrator: custom_components.waste_collection_schedule.waste_collection_schedule.source_shell
Źródło: custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py:189
Integration: Waste Collection Schedule (dokumentacja)
Pierwsze zdarzenie: 19 lipca 2026 o 22:13:33 (3 wystąpienia)
Ostatnio zalogowany: 07:15:12

[...]

PR #6983, opened 2026-07-20
Retry setup after transient source fetch failures

PrzemyslawKlys:
Addresses #6981.

## Summary

- make `SourceShell.fetch()` report success without discarding cached entries or refresh time on failure
- let Home Assistant retry a config entry when its first source fetch fails instead of loading sensors as `unknown`
- advance fetch-interval bookkeeping only after successful UI and YAML fetches, while still attempting every configured YAML source

This targets the stable `master` branch and is intentionally independent of the v3 source architecture.

## Validation

- 38 tests passed after merging current `master`
- 38 tests passed in a fresh Python 3.12 CI-compatible dependency environment
- the existing focused formatting, lint, type, and recovery review evidence remains applicable


markvp:
Thanks @PrzemyslawKlys — the retry approach looks solid: returning success from `SourceShell.fetch()`, only advancing `_last_fetch_date` on success, and raising `UpdateFailed` on first-refresh failure is exactly the right HA pattern, and the added tests are welcome.

The failing check is a CI-environment issue, not your logic: the new tests import `wcs_coordinator`, which pulls in `DeviceInfo` from `homeassistant.helpers.device_registry`, and that import fails on the old Python 3.10 / older-HA runner CI used to use. Master now runs CI on Python 3.12 + HA 2024.4 (via #6987), where that import resolves — so merging master into this branch should clear it. Could you merge master in? Then I'll get it merged.

PrzemyslawKlys:
Merged current `master` as requested. The new head is `c8a790770a7ffb20f7f5a278ed9572ceb8c5a367`; the full 38-test suite passes in a fresh Python 3.12 CI-compatible environment.

PrzemyslawKlys:
The remaining hassfest failure is inherited from current `master`, not introduced by this PR. The [master run after #6987](https://github.com/mampfes/hacs_waste_collection_schedule/actions/runs/30011442678) fails on the same generated KVA Thurgau translation placeholders; this PR's repository validation is green.

markvp (approved):
Approving. The retry approach is sound — keeping cached collections on failure, reporting first-refresh failures via `UpdateFailed`, and only advancing the fetch interval on success are all the right patterns, and the tests are a welcome addition.

The remaining hassfest red here is the inherited-master issue you flagged (invalid translation placeholders), now fixed on master via #7060 and #7061. Your branch merged master just before those landed, so it clears once this merges onto current master. Thanks again for surfacing that master breakage.

PR #6984, opened 2026-07-20
Handle BOM-prefixed EcoHarmonogram responses in v3

PrzemyslawKlys:
## Summary

- decode EcoHarmonogram response text as UTF-8 with BOM support before parsing JSON
- retry the provider request once when a response is not valid JSON
- add focused coverage for normal JSON, BOM-prefixed JSON, and a transient invalid response

The integration-wide transient-fetch recovery previously included here has been removed at maintainer request. That work is preserved separately for a future v3.1 PR; this PR now contains only the provider-specific parser fix suitable for `release/3.0.0`.

## Validation

- 3 focused EcoHarmonogram response tests passed
- 3,070 passed, 4 skipped, and 2,139 deselected in the complete offline suite
- Ruff, Ruff format, codespell, Bandit, and mypy hooks passed

Related to #6981 and the v3 architecture in #6562.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/EcoHarmonogramPL.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/EcoHarmonogramPL.py
@@ -1,4 +1,5 @@
 import datetime
+import json
 import logging
 import re
 from collections.abc import Mapping
@@ -21,6 +22,7 @@ if TYPE_CHECKING:
 _LOGGER = logging.getLogger(__name__)
 
 API_URL = "https://ecoharmonogram.pl/api/api.php"
+JSON_RESPONSE_ATTEMPTS = 2
 
 SUPPORTED_APPS_LITERAL = Literal[
     None,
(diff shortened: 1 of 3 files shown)


markvp (changes requested):
Thanks for this, and for the thorough diagnosis and tests. The EcoHarmonogram parsing fix is spot on: I confirmed it's a genuine v3 regression. On `master` the client used `requests` with `response.encoding = "utf-8-sig"`, which handled the BOM; the pipeline migration moved it onto the curl_cffi session, whose `.json()` parses raw bytes and chokes on that same payload. Decoding `utf-8-sig` explicitly is the right fix, and I like that the error message doesn't leak the response body. This half is a must-have for 3.0.

Could I ask you to split the PR, though? The transient-recovery hardening (the `SourceShell.fetch` -> `bool` change, `UpdateFailed` on first refresh, and the YAML "advance only when all sources succeed" behaviour) isn't actually v3-specific: the same "a failed fetch still marks the day as fetched" behaviour is present on `master` today, so it's a long-standing robustness improvement rather than a v3 regression. It also changes coordinator and YAML fetch semantics that deserve their own review, and it would widen the blast radius of the release branch.

So my suggestion:
- Keep this PR to the EcoHarmonogram fix, targeting `release/3.0.0`.
- Move the recovery hardening to a separate PR against 3.1 (it's an additive change, so a minor). Given the v2 counterpart in #6983, that one may even be a candidate for a 2.x patch.

Happy to help line up the 3.1 PR once this one is scoped down. Thanks again.

PrzemyslawKlys:
Applied the requested scope split. This branch is now rebased on the current `release/3.0.0` and contains only the EcoHarmonogram BOM/JSON parser fix plus its three response tests. The generic transient-fetch recovery is no longer in this PR and is preserved separately for a future v3.1 PR. The focused tests pass, and the complete offline suite reports 3,070 passed and 4 skipped.

markvp (approved):
Thanks for splitting this down to just the EcoHarmonogram fix — exactly the 3.0 scope. The `utf-8-sig` BOM decode plus the single retry reads cleanly, the regression test is now wired into `pytest.ini` so it actually runs, and CI is green. Good for 3.0.

PR #6985, opened 2026-07-20
Add opt-in device-backed sensor controls

PrzemyslawKlys:
Successor to #6052, built on the v3 collection model.

Adds opt-in configuration controls on the integration service device, with a child device for each configured sensor. The existing bulk options editor remains available. Create/delete actions and per-sensor configuration preserve sensor identities and history; creation actions are idempotent and appear for types discovered by later fetches.

New sensors store stable `WasteType.id` filters. Existing display-name and provider-label filters keep working without rewriting saved configuration. Pipeline collections retain their original provider label so source-specific filters and custom aliases survive canonical type mapping. Coordinator timers and delayed callbacks are cancelled on unload.

Rebased on the current `release/3.0.0` branch. This remains a draft for 3.1 and should be retargeted to `master` after the 3.0 tag, as requested. No generated documentation or translation output is included.

Validation: 7,718 offline tests passed, 8 skipped; focused migration tests cover source parsing, old filters, aliases, identity, and duplicate creation. All pre-commit checks pass in the Python 3.14 validation environment. An independent compatibility review found no actionable issues.

PR #6987, opened 2026-07-20
Align validation with the supported Home Assistant version

PrzemyslawKlys:
## Summary

- raise the declared Home Assistant minimum to 2024.4.0
- run repository validation on Python 3.12
- allow urllib3 1.26.18 and newer in the development requirements

## Why

The current `master` config flow imports `ConfigFlowResult`, which is available from Home Assistant 2024.4. Keeping 2023.12 as the declared minimum therefore understates the integration's real runtime requirement.

This remains separate from #6983 so the transient-fetch recovery PR stays product-focused.

## Validation

- Python 3.12 with Home Assistant 2024.4.0 successfully imported `ConfigFlowResult` and the integration config flow
- 34 base tests passed
- dependency consistency check passed
- workflow YAML formatting and actionlint passed

### DIFF
--- a/.github/workflows/validate.yaml
+++ b/.github/workflows/validate.yaml
@@ -18,10 +18,10 @@ jobs:
       - uses: actions/checkout@v5
         with:
           fetch-depth: 0
-      - name: Set up Python 3.10
+      - name: Set up Python 3.12
         uses: actions/setup-python@v6
         with:
-          python-version: '3.10'
+          python-version: '3.12'
           cache: pip
           cache-dependency-path: requirements.txt
       - name: Install dependencies
--- a/hacs.json
+++ b/hacs.json
@@ -2,5 +2,5 @@
   "name": "Waste Collection Schedule",
   "domains": ["sensor"],
   "iot_class": "cloud_polling",
-  "homeassistant": "2023.12.0"
+  "homeassistant": "2024.4.0"
 }
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,11 +7,11 @@ python-dateutil>=2.8.2
 pytz>=2021.3
 PyYAML>=6.0.1
 requests>=2.31.0
-urllib3>=2.0.7
+urllib3>=1.26.18
 jinja2>=3.1.2
 lxml>=4.9.4
 pycryptodome>=3.20.0
 typing_extensions>=4.12.2
 pypdf>=5.1.0
 pdfminer.six>=20221105
-homeassistant
+homeassistant>=2024.4.0


bbr111:
@markvp @PrzemyslawKlys 
would we like to bump python version direct to 3.12 or 3.13?

I've created a table for us as discussion base.

## Home Assistant installations affected by Python version changes

Reported installations: **655.295**

| Python upgrade | Minimum HA version | Excluded systems | Remaining systems | Excluded percentage | Transition period |
|---|---:|---:|---:|---:|---|
| 3.10 -> 3.11 | 2023.8+ | 7.486 | 647.809 | 1,14% | Home Assistant 2023.7 supports Python 3.10 and 3.11 |
| 3.11 -> 3.12 | 2024.2+ | 11.772 | 643.523 | 1,80% | - |
| 3.12 -> 3.13 | 2025.2+ | 35.650 | 619.645 | 5,44% | Home Assistant 2024.12 to 2025.1 supports Python 3.12 and 3.13 |
| 3.13 -> 3.14 | 2026.3+ | 142.715 | 512.580 | 21,78% | Home Assistant 2026.2 supports Python 3.13 and 3.14 |

### Notes

- The data is based on installations that voluntarily report analytics to Home Assistant.
- The analytics data contains the Home Assistant version, but not the exact Python version used by each installation.
- The excluded installation counts use the first Home Assistant release that requires the newer Python version.
- Installations in transition releases may already use the newer Python version.

PrzemyslawKlys:
@bbr111 Thanks for raising this. I tested the current runtime before turning the idea into a proposal:

- Python 3.11.15 with the exact advertised floor, Home Assistant 2023.12.0: dependency check clean and 34 tests passed.
- Python 3.14.2 with the current Home Assistant release, 2026.7.2: dependency check clean and 34 tests passed.
- The same Python 3.14.2/current-HA environment also passed the product PR suites for #6983, #6984, and #6985.

My suggestion is therefore not to replace the minimum lane with only 3.12 or 3.13. Neither represents an endpoint we specifically need to protect. Instead:

- keep Python 3.11 + HA 2023.12.0 as the minimum compatibility lane;
- add Python 3.14.2 + latest HA as the forward-compatibility lane.

This leaves the user-facing minimum in `hacs.json` unchanged, so it does not exclude older installations. It adds one parallel CI job and gives us early warning when current Home Assistant APIs, Python, or dependencies change.

I opened #6999 as an optional stacked follow-up so #6987 can remain the small CI-floor repair. If the direction is useful, the same shape can be ported to `release/3.0.0` after #6988. If the additional CI cost or floating-current canary does not make sense for this repository, please feel free to close #6999; #6987 remains useful and independent.

markvp:
Thanks for this, and for the runtime testing. One correction before it lands: `config_flow.py` on master uses `ConfigFlowResult`, which Home Assistant only added in 2024.4, so 2023.12 isn't actually a viable floor. It passes CI today only because `pytest.ini` collects just `test_source_components.py`, which never imports `config_flow`. The same fix is going onto `release/3.0.0` in #7017.

Could you update this PR to the true floor? I'd have pushed it myself, but maintainer edits 403'd on the renamed fork. Three small changes:

```diff
# hacs.json
-  "homeassistant": "2023.12.0"
+  "homeassistant": "2024.4.0"

# requirements.txt
-urllib3>=2.0.7
+urllib3>=1.26.18
-homeassistant>=2023.12.0
+homeassistant>=2024.4.0

# .github/workflows/validate.yaml
-          python-version: '3.11'
+          python-version: '3.12'
```

`urllib3` relaxes to `>=1.26.18` (still CVE-patched) so it stays compatible with HA 2024.4's `urllib3<2`; it is dev/CI-only, not in `manifest.json`. Impact is ~1.8% of installations per the analytics table above. Thanks!

PrzemyslawKlys:
Applied, thank you. The branch is now rebased on current `master` and uses Python 3.12, `homeassistant: 2024.4.0` in `hacs.json`, `urllib3>=1.26.18`, and `homeassistant>=2024.4.0`. I verified the integration config-flow import against the exact Home Assistant 2024.4.0 floor and all 34 base tests pass. #6999 has also been restacked on this corrected floor.

markvp (approved):
Thanks for splitting this out. The alignment is correct: master already imports `ConfigFlowResult` in `config_flow.py`, which only exists from HA 2024.4, so the `hacs.json` floor of 2023.12 was understating the real requirement. Bumping to 2024.4.0 and validating on 3.12 both make sense.

Flagging for the record (no change needed): since this raises the declared HA floor it counts as an HA-support drop under `doc/versioning.md`, so we will batch it into a breaking/major release rather than a patch. Pairs naturally with #6999, which adds the CI matrix to prove the new floor.

PR #7023, opened 2026-07-21
Retry v3 setup after transient source fetch failures

PrzemyslawKlys:
## Summary

- keep cached collections when a source fetch fails instead of replacing them with an apparent successful empty result
- report first-refresh failures through `UpdateFailed` so Home Assistant can retry setup
- leave the fetch interval open after a transient failure instead of marking the day as successfully fetched
- for YAML configurations, attempt every configured source and advance the interval only when all sources succeed

## Release targeting

This is the generic recovery work split out of #6984 at maintainer request. It remains a draft for 3.1 and is rebased on the current `release/3.0.0` architecture. After 3.0 is tagged, it can be rebased or retargeted to `master`, following the same holding pattern as #6985.

The EcoHarmonogram BOM/JSON parser fix remains separate in merged #6984 and is not included here. The equivalent stable-branch recovery is #6983.

## Validation

- 4 focused recovery tests passed; 7 focused recovery/provider regression tests passed together
- 3,120 passed, 4 skipped, and 2,127 deselected in the complete Python 3.12 offline suite
- Ruff, Ruff format, codespell, Bandit, mypy, and pyright passed on the preserved recovery patch before this base-only rebase
- the source-and-test patch matches the reviewed recovery contract; only the current pytest collection list was reconciled during rebase

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_api.py
+++ b/custom_components/waste_collection_schedule/waste_collection_api.py
@@ -109,9 +109,13 @@ class WasteCollectionApi:
             self._update_sensors_callback()
             return
 
+        fetch_succeeded = True
         for shell in self._source_shells:
-            shell.fetch()
-        self._last_fetch_date = today
+            if not shell.fetch():
+                fetch_succeeded = False
+
+        if fetch_succeeded:
+            self._last_fetch_date = today
 
         self._update_sensors_callback()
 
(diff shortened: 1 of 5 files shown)


PrzemyslawKlys:
Rebased this holding draft on the current `release/3.0.0` branch without changing recovery scope. The only conflict was the shared `pytest.ini` collection list, which now retains the merged EcoHarmonogram regression tests and this PR's recovery tests. Current validation is 7 focused tests plus 3,120 passed / 4 skipped in the complete offline suite.

markvp (approved):
Approving for 3.0.0. This is the v3 counterpart of #6983, adapted to the BaseSource pipeline codebase, and it's fully current with `release/3.0.0` with both CI lanes (including hassfest) green. Retagged from 3.1.0 to 3.0.0 so the recovery behaviour ships with v3. Thanks!

PR #7059, opened 2026-07-23
Add support for additional waste types to lisburn_castlereagh source

JDickson835:
## Summary
Certain regions from inside the lisburn_castlereagh_gov_uk.py source have different configurations/types of waste collection.
A number of unsupported types being returned by the API resulted in the transformation logic breaking.

To fix this I have:
- Added support for `FoodWaste` & `WheelieBox` waste to the lisburn_castlereagh_gov_uk source.
- Updated the tests to target areas with both known waste configurations.

## Type of change
- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/lisburn_castlereagh_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/lisburn_castlereagh_gov_uk.py
@@ -20,19 +20,23 @@ TEST_CASES = {
     "Test_001": {"postcode": "BT28 1AG", "house_number": "19A"},
     "Test_002": {"postcode": "BT26 6AB", "house_number": "31"},
     "Test_003": {"postcode": "BT26 6AB", "house_number": 15},
-    "Test_004": {"property_id": "DYYSm8Ls8XxGi3Nq"},
-    "Test_005": {"property_id": "ZJat6DWG1Lp95xp1"},
+    "Test_004": {"property_id": "RTTKBJY77wmTr5wP"},
+    "Test_005": {"property_id": "6opWbT5hi9odCQ2u"},
 }
 
 API_URL = "https://lisburn.isl-fusion.com"
 ICON_MAP = {
     "ResidualBin": Icons.GENERAL_WASTE,
     "RecycleBin": Icons.RECYCLING,
-    "BrownBin": Icons.BIO_KITCHEN,
+    "WheelieBox": Icons.RECYCLING,
+    "BrownBin": Icons.GARDEN,
+    "FoodWaste": Icons.BIO_KITCHEN,
 }
 NICE_NAMES = {
     "ResidualBin": "Refuse",
     "RecycleBin": "Recycling",
+    "WheelieBox": "Wheelie Box",
+    "FoodWaste": "Food Waste",
     "BrownBin": "Garden",
 }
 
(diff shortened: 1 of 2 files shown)


JDickson835:
Output from source test
image

Output from pytest
image

markvp (changes requested):
Thanks @JDickson835, nicely diagnosed. I confirmed the two property IDs you replaced are dead (the API returns an empty nextCollections for both), and your new Test_005 exercises the FoodWaste and WheelieBox streams properly. Moving BrownBin to the garden icon also lines it up with its existing "Garden" label, so that reads better than before.

Two things to sort before this can go in:

1. `t=NICE_NAMES.get(bin["name"])` still returns None for any bin name that isn't in the map, which is the same failure mode you're fixing here. A fallback to the raw name lets the next unknown stream degrade gracefully:

```python
bin_type = bin["name"]
...
t=NICE_NAMES.get(bin_type, bin_type),
icon=ICON_MAP.get(bin_type),
```

2. `doc/source/lisburn_castlereagh_gov_uk.md` documents the argument as `post_code`, but the constructor kwarg is `postcode`, so anyone copying the example hits a `SourceArgumentExceptionMultiple`. While you're in there, the `property_id` in the example is one of the IDs you have just retired, so it is worth swapping for `RTTKBJY77wmTr5wP`.

One other thing that would help: this PR has "Allow edits by maintainers" turned off, so we cannot push small fixes for you. If you enable it (checkbox in the right-hand column of the PR page) we can save you a round trip on things like this in future.

JDickson835:
> Thanks @JDickson835, nicely diagnosed. I confirmed the two property IDs you replaced are dead (the API returns an empty nextCollections for both), and your new Test_005 exercises the FoodWaste and WheelieBox streams properly. Moving BrownBin to the garden icon also lines it up with its existing "Garden" label, so that reads better than before.
> 
> Two things to sort before this can go in:
> 
> 1. `t=NICE_NAMES.get(bin["name"])` still returns None for any bin name that isn't in the map, which is the same failure mode you're fixing here. A fallback to the raw name lets the next unknown stream degrade gracefully:
> 
> ```python
> bin_type = bin["name"]
> ...
> t=NICE_NAMES.get(bin_type, bin_type),
> icon=ICON_MAP.get(bin_type),
> ```
> 
> 2. `doc/source/lisburn_castlereagh_gov_uk.md` documents the argument as `post_code`, but the constructor kwarg is `postcode`, so anyone copying the example hits a `SourceArgumentExceptionMultiple`. While you're in there, the `property_id` in the example is one of the IDs you have just retired, so it is worth swapping for `RTTKBJY77wmTr5wP`.
> 
> One other thing that would help: this PR has "Allow edits by maintainers" turned off, so we cannot push small fixes for you. If you enable it (checkbox in the right-hand column of the PR page) we can save you a round trip on things like this in future.

Excellent suggestions for improvements; thanks for the review @markvp 
I have addressed the comments in a pair of new commits and enabled the `Allow edits by maintainers` setting.

bbr111 (approved):
Thanks for the quick turnaround @JDickson835 — both points are addressed.

I re-ran the live test against master and reproduced the failure for the two retired property IDs, then confirmed all five cases pass on your branch. Test_005 returns 9 entries spanning Food Waste, Wheelie Box, Refuse and Garden, so the new streams are properly exercised.

I pushed one small follow-up: the Configuration Variables section of the doc still listed the argument as `post_code` even though the YAML example was fixed, so anyone reading the reference list would still have hit the same error. That's now consistent throughout.

Structural tests pass (34 passed). Good fix — thanks for enabling maintainer edits too, it made this a one-liner rather than another round trip.

PR #7063, opened 2026-07-24
Fix centralbedfordshire new site

Liam-Whiteside:
## Summary
Central Bedfordshire Council migrated their waste collection page to the LocalGov Drupal Waste Collection module, which uses a different URL flow (postcode search -> UPRN lookup -> direct view by UPRN) and a different page structure than the old scraper expected.

- Replace the old multi-step POST/session flow with two plain GETs
- Parse dates from the <time datetime> attribute instead of text
- Split the combined 'X, Y and Z collections' sentence into individual Collection entries per bin type
- Deduplicate bin types per date, since some properties (e.g. those with a separate garden waste subscription) list the same date across two <li> rows

## Type of change
- [x] Bug fix / source fix

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/centralbedfordshire_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/centralbedfordshire_gov_uk.py
@@ -1,8 +1,40 @@
-import time
+"""
+Central Bedfordshire Council waste collection source.
+
+Rewritten for the council's new July 20216 site
+
+    Postcode search : {base}/find?postcode={postcode}
+    Direct schedule : {base}/view/{uprn}
+
+On this council's site {base} = /waste-and-recycling/waste-collection-schedule
+
+Address lookup:
+    The /find?postcode=... page has a <select id="edit-uprn" name="uprn">
+    with one <option value="{uprn}">{full address}</option> per property.
+
+Results page:
+    Each collection day is a <li class="waste-collection__day"> containing:
+        <span class="waste-collection__day--day"><time datetime="DD-MM-YYYY">
+        <span class="waste-collection__day--type">Recycling, garden waste
+            and food waste collections</span>
+    Note the type is a single natural-language sentence combining every bin
+    collected that day (e.g. "Refuse (black bin) and food waste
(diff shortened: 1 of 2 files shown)


McMoobud:
Nice work @Liam-Whiteside — same diagnosis I landed on. Heads-up for reviewers / a possible follow-up: the new site also exposes a structured **per-UPRN iCal feed** at

```
https://www.centralbedfordshire.gov.uk/waste-and-recycling/waste-collection-schedule/download/{UPRN}
```

(response is `text/calendar`, parseable via this repo's `service.ICS` helper). It's the same data the on-page "Add to iCalendar" button uses, so it's a stable public endpoint and more robust than HTML scraping against future Drupal frontend tweaks. If Liam's PR is the one that merges, we could migrate the internals to that endpoint later without changing the user-facing config.

I opened #7074 with that variant + an optional `uprn` arg (skips the postcode form entirely) before spotting this PR — closing mine as a duplicate since yours is first. Feel free to cherry-pick anything useful from #7074 if it helps.

markvp (approved):
Thanks @Liam-Whiteside, this is a really thorough rewrite. The docstring explaining the LocalGov Drupal URL flow and the per-date deduplication is genuinely useful for the next person who touches this, and moving from the POST/session dance to two plain GETs is a big simplification.

I verified both test cases live (65 entries each) and confirmed the dedup path is real: 1 Chestnut Avenue returns 39 rows with 13 duplicated dates, which collapse correctly.

I have pushed a couple of small tweaks on top:

- An unrecognised postcode returns HTTP 200 with the dropdown simply missing, so that case was hitting the "page structure may have changed" error. It now raises `SourceArgumentNotFound("postcode", ...)` so the UI points at the right field.
- Restored the old "postcode without space" coverage on the second test case (the site accepts both forms).
- Refreshed `doc/source/centralbedfordshire_gov_uk.md`, which still pointed at the old `/info/163/` URL and described the previous address-option format.

Also worth noting for later: @McMoobud spotted a per-UPRN iCal feed at `/waste-and-recycling/waste-collection-schedule/download/{UPRN}` (confirmed `text/calendar`). That would be a more robust internal implementation down the track, but it is a separate change and does not need to hold this up.

PR #7067, opened 2026-07-25
Fix up the UserAgent and URL/domain so Auckland Council works again

nicwise:
## Summary

* Change the URL as "new.aucklandcouncil.govt.nz" just redirects to www anyway
* Update the UserAgent as they have a WAF and are checking for old UAs

## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [X] Documentation update
- [ ] Other

## Checklist

- [ ] `ruff check --fix` and `ruff format` run on changed source files

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
@@ -6,8 +6,8 @@ from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection
 
 TITLE = "Auckland Council"
-DESCRIPTION = "Source for Auckland council."
-URL = "https://new.aucklandcouncil.govt.nz"
+DESCRIPTION = "Source for Auckland Council."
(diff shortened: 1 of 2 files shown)


nicwise:
Sorry, my Mac doesn't have a fully functioning python setup - so I couldn't run ruff.

Works in tests + live on my HA instance.

Changes are very minor tho :)

markvp:
Thanks @nicwise, and thanks for chasing this down and for nudging on the issue to get one of these merged. Your diagnosis was right and the `new.` to `www.` domain change is real, I verified it 301s.

I'm going to land #7078 instead, mainly because it also fixes the silent failure behind #7070: without a `raise_for_status()`, a future block gives people an empty sensor and nothing in the log, which is exactly what got us here. It carries your domain fix and both doc links, and I've folded your `DESCRIPTION` capitalisation in on top.

The other reason is the comment you added, which is honest and correct: a pinned UA string will go stale again. `curl_cffi` with `impersonate="chrome"` avoids that by tracking a current profile.

Closing this as superseded by #7078, with credit. Appreciate you taking the time.

PR #7071, opened 2026-07-26
fix(swale_gov_uk): update form field IDs

jameswgm87:
## Summary

Swale Borough Council has updated the form field IDs used by the "Check Your Bin Day" form.

The current source still submits the previous field names, causing configuration to fail with:

> Could not find next collection date – the page may have changed or returned an error.

This updates both POST payloads to use the current field identifiers.

## Testing

- Tested against a live Swale Borough Council property.
- Configuration now completes successfully.
- Waste collection calendar is created successfully.
- Collection data is returned correctly.

The field names were verified against the current live Swale website.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/swale_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/swale_gov_uk.py
@@ -61,10 +61,10 @@ class Source:
 
         # mimic postcode search
         payload: dict = {
-            "SQ_FORM_499078_PAGE": "1",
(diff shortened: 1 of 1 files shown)


markvp (approved):
Thanks for tracking this down. I verified the new field IDs against the live Swale form: page 1 now uses `SQ_FORM_535662` / `q535679:q1` and page 2 uses `q535685:q1`, exactly as in your change. All three test cases fetch real collection dates again (the occasional 403 is just Swale's rate limiting, which is already noted in the source).

Structural tests and ruff are clean.

One thought for later, not for this PR: the form IDs are hardcoded, so this will break again the next time the council republishes the form. Reading them from an initial GET would make the source self-healing. Happy to open a follow-up issue for that if you are not keen to take it on.

PR #7073, opened 2026-07-27
Adds garden waste to herefordshire

Jab2870:
## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other

## Checklist

   - Ruff found issues, but not in the files I have changed

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/herefordshire_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/herefordshire_gov_uk.py
@@ -45,6 +45,7 @@ HEADER = {"user-agent": "Mozilla/5.0"}
 ICON_MAP = {
     "General": Icons.GENERAL_WASTE,
     "Recycling": Icons.RECYCLING,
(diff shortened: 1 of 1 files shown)


markvp (approved):
Thanks for adding garden waste, and for tidying the icons up to the Icons enum while you were in there.

I pushed a couple of small tweaks. `ICON_MAP` is a dict, so `ICON_MAP.General` was raising `AttributeError` and both test cases were failing. I also found that the "Garden waste - brown bin" heading is rendered for every property, but properties without a garden subscription get a calendar download link under it rather than dates, which crashed `strptime` and took general rubbish and recycling down with it. Non-date list items are now skipped, so those addresses keep working.

Both test cases pass live again and `tests/test_source_components.py` is green.

One thing I could not check: I have no publicly identifiable Herefordshire address with an active garden subscription, so the garden dates themselves are untested. The code assumes the same `%A %d %B %Y` format as the other two bins, which looks right from the page markup. If you have a subscribed property to hand, a quick sanity check would be great.

PR #7075, opened 2026-07-27
Fix Esch schedule fetch with identity encoding

framesfree:
The Esch endpoint's compressed response advertises Transfer-Encoding: chunked twice. Some urllib3 versions reject it as multiple Transfer-Encoding headers, preventing schedule refreshes.

Use the canonical trailing-slash endpoint and request identity encoding. This avoids the malformed compressed response without changing user configuration. Add a regression test covering the request and parsed schedule.


markvp (approved):
Thanks for the clear diagnosis and the regression test.

I reproduced it against the live endpoint: the compressed response really does come back with `Transfer-Encoding: chunked, chunked`, and requesting identity encoding drops it to a single header. Nice catch. Live test passes for both zones (123 and 121 entries), and the structural suite is green.

Good spot on the trailing slash too, that also puts the previously unused `API_URL` constant back to work.

PR #7072, opened 2026-07-26
Fix Auckland Council (aucklandcouncil_govt_nz): HTTP 406 from council WAF broke all lookups

bterps:
## Summary

Service provider: **Auckland Council** (`aucklandcouncil_govt_nz`, New Zealand)

The Auckland Council WAF now rejects requests that only send a `User-Agent` header with **HTTP 406 Not Acceptable**, so the source returned an empty response for every `area_number` (config flow fails with "The source returned an empty response"). Sending a complete browser-like header set — `Accept`, `Accept-Language`, the `Sec-Fetch-*` fetch-metadata headers and `sec-ch-ua` client hints — restores normal operation.

Only `HEADER` in the existing source file is changed; no user-facing arguments changed, so `doc/source/aucklandcouncil_govt_nz.md` needs no update. `ruff check` and `ruff format` pass.

## Test output

Before (current master):

```
Testing source aucklandcouncil_govt_nz ...
  found 0 entries for 429 Sea View Road
  found 0 entries for 8 Dickson Road
  found 0 entries for with Food Scraps
  found 0 entries for 3 Andrew Road
```

After (this PR):

```
Testing source aucklandcouncil_govt_nz ...
  found 2 entries for 429 Sea View Road
    2026-07-27 : rubbish
    2026-08-03 : recycling
  found 2 entries for 8 Dickson Road
    2026-07-30 : rubbish
    2026-08-06 : recycling
  found 3 entries for with Food Scraps
    2026-07-27 : rubbish
    2026-07-27 : food scraps
    2026-08-03 : recycling
  found 3 entries for 3 Andrew Road
    2026-07-31 : rubbish
    2026-07-31 : food scraps
    2026-08-07 : recycling
```

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #7078, opened 2026-07-28
fix(aucklandcouncil_govt_nz): bypass WAF 406 and fail loudly on empty results

### LINKED ISSUE #7070: [Source Defect]: aucklandcouncil_govt_nz
### Source Name

aucklandcouncil_govt_nz

### What's happening?

The Rubbish Days and Recycling Days sensors update their timestamp on every fetch (and the coordinator logs success: True), but the rubbish / recycling attributes always come back as an empty string. No exceptions or warnings appear in the log even with custom_components.waste_collection_schedule set to debug.

Steps to reproduce:

Configure the integration with source aucklandcouncil_govt_nz and a valid area_number.
Wait for a scheduled fetch (or reload the config entry).
Check sensor.<name>_rubbish_days / _recycling_days — state is unknown, rubbish/recycling attributes are "".

Expected behavior: Sensors populate with the upcoming collection dates for the configured address.

Actual behavior: Sensors stay empty indefinitely, with no logged error.

Root cause:

https://www.aucklandcouncil.govt.nz/en/rubbish-recycling/rubbish-recycling-collections/rubbish-recycling-collection-days/12340303405.html — returns HTTP 406 Not Acceptable for automated requests (I reproduced this independently outside of Home Assistant, multiple times). The generic (non-address) collection-days page loads fine, so this looks like bot/WAF protection specifically on the personalized per-address pages.

Since the source code doesn't check the response status before parsing with BeautifulSoup, it silently parses the 406 error page, finds no acpl-schedule-card elements, and returns an empty list — no exception is raised, so the coordinator reports success with no data.

### Error Logs

```Shell
Nothing is logged as 406 errors are not being parsed and empty return string is not checked.
```

### Source Configuration

```YAML
Environment:

Home Assistant: 2026.7.4
Waste Collection Schedule integration: v2.31.1 (latest)
Source: aucklandcouncil_govt_nz
Config: area_number: 12340303405
```

### Checklist

jtbnz:
Fixes #7070

## Service provider

[Auckland Council](https://www.aucklandcouncil.govt.nz) (`aucklandcouncil_govt_nz`), New Zealand.

## Problem
A hard-coded `Chrome/58.0.3029.110` User-Agent (2017). Auckland Council's WAF now answers **HTTP 406 Not Acceptable** to it on the per-address collection page.

`fetch()` never checked the response status, so the 293-byte 406 error page was handed to BeautifulSoup, no `acpl-schedule-card` was found, and an empty list was returned. The coordinator logged a successful fetch with no data, and nothing appeared in the log even at debug level — exactly the silent failure described in #7070.

The page markup itself has **not** changed — the existing parser works unmodified once the request is accepted.

| Request | Result |
|---|---|
| `Chrome/58` UA (as shipped), any other headers | `406` |
| no UA | `406` |
| `python-requests/2.32.3` | `406` |
| `Chrome/128` UA, no other headers | `200` |
| `curl_cffi` `impersonate="chrome"` | `200` |

## Changes

`custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py`

- Use `curl_cffi` with `impersonate="chrome"` instead of a hand-written `User-Agent`. A hardcoded UA string is what broke here, and `impersonate` also matches on TLS fingerprint. `curl_cffi` is already in `manifest.json` / `requirements.txt` and used by other sources, so this adds no dependency.
- Call `raise_for_status()` so a future block surfaces as an error rather than an empty schedule.
- Raise `SourceArgumentNotFound("area_number", ...)` when no collections parse, covering the HTTP-200-but-empty invalid-area case.
- Extract parsing into a module-level `parse_page()` and move waste-type normalisation into a `COLLECTION_TYPES` map. The `try` block now wraps only `toDate()`, the one call that can raise — previously it also enclosed `Collection(...)` and the `append`.
- `new.aucklandcouncil.govt.nz` now 301s to `www.aucklandcouncil.govt.nz`; `URL` and the fetch URL updated.

`doc/source/aucklandcouncil_govt_nz.md`

- Same `new.` → `www.` link update (2 links, both verified 200).

Partial emptiness still parses correctly and does **not** raise — `12342453293` has no food scraps collection and returns rubbish + recycling as before.

No generated files are touched, and `update_docu_links.py` was not run.

## Test output

```
$ python test_sources.py -s aucklandcouncil_govt_nz -l
Testing source aucklandcouncil_govt_nz ...
  found 2 entries for 429 Sea View Road
    2026-08-03 : rubbish
    2026-08-03 : recycling
  found 2 entries for 8 Dickson Road
    2026-07-30 : rubbish
    2026-08-06 : recycling
  found 3 entries for with Food Scraps
    2026-08-03 : rubbish
    2026-08-03 : food scraps
    2026-08-03 : recycling
  found 3 entries for 3 Andrew Road
    2026-07-31 : rubbish
    2026-07-31 : food scraps
    2026-08-07 : recycling
```

All four `TEST_CASES` return an empty list on `master` today.

Also verified:

- `python -m pytest tests/` — 38 passed
- `ruff check` / `ruff format --check` — clean
- `mypy --ignore-missing-imports --explicit-package-bases` — clean for this file (2 pre-existing errors on `master` fixed in passing, from `icon.get("class", [])` → `icon.get("class") or []`)
- An out-of-tree regression suite covering the broken/fixed behaviour: a frozen snapshot of the `master` source asserting the 406 and the silent empty list, live checks of all four test cases plus a real address, an invalid area number raising, and offline parser tests against captured HTML fixtures. 3 passed / 7 failed before the fix, 10 passed after.

🤖 Assisted by [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
@@ -7,7 +7,7 @@ from waste_collection_schedule import Collection
 from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Auckland Council"
-DESCRIPTION = "Source for Auckland council."
+DESCRIPTION = "Source for Auckland Council."
 URL = "https://www.aucklandcouncil.govt.nz"
 
 TEST_CASES = {
@@ -125,7 +125,7 @@ class Source:
         # The site's WAF answers HTTP 406 to plain requests/urllib clients, so
         # impersonate a real browser (TLS fingerprint included).
         session = requests.Session(impersonate="chrome")
-        r = session.get(API_URL.format(self._area_number))
+        r = session.get(API_URL.format(self._area_number), timeout=30)
         r.raise_for_status()
 
         entries = parse_page(r.text)
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/aucklandcouncil_govt_nz.py
@@ -1,13 +1,14 @@
 import datetime
 import re
 
-import requests
 from bs4 import BeautifulSoup
+from curl_cffi import requests
 from waste_collection_schedule import Collection
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Auckland Council"
 DESCRIPTION = "Source for Auckland council."
-URL = "https://new.aucklandcouncil.govt.nz"
+URL = "https://www.aucklandcouncil.govt.nz"
 
 TEST_CASES = {
     "429 Sea View Road": {"area_number": "12342453293"},  # Monday
@@ -15,6 +16,8 @@ TEST_CASES = {
     "with Food Scraps": {"area_number": 12341998652},
     "3 Andrew Road": {"area_number": "12345375455"},  # friday with foodscraps
 }
+
+API_URL = "https://www.aucklandcouncil.govt.nz/en/rubbish-recycling/rubbish-recycling-collections/rubbish-recycling-collection-days/{}.html"
 
 MONTH = {
     "January": 1,
@@ -29,6 +32,13 @@ MONTH = {
     "October": 10,
     "November": 11,
     "December": 12,
+}
+
+# icon class used on the page -> waste type reported by this source
+COLLECTION_TYPES = {
+    "rubbish": "rubbish",
+    "recycle": "recycling",
+    "food-waste": "food scraps",
 }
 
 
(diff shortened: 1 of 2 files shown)


jtbnz:
there is also an existing PR https://github.com/mampfes/hacs_waste_collection_schedule/pull/7072 My bad I missed this one and this one https://github.com/mampfes/hacs_waste_collection_schedule/pull/7067

markvp (approved):
Thanks @jtbnz, this is the one I'd like to land. Three PRs turned up for this breakage and yours is the only one that fixes the actual defect in #7070: the source silently reported success with no data. `raise_for_status()` plus the `SourceArgumentNotFound` on an empty parse means the next time the council changes something, people see an error instead of an empty sensor.

Using `curl_cffi` with `impersonate="chrome"` is also the right call over a hand-written User-Agent. It's the documented approach for WAF-protected sites here, it's already a dependency, and it tracks a current browser profile rather than a version string that goes stale. Narrowing the `try` to just `toDate()` is a nice catch too, the old block was swallowing errors from the `Collection` construction as well.

I've made a couple of small tweaks: a `timeout=30` on the request so a stalled response can't hang an update cycle, and I've folded in the `DESCRIPTION` capitalisation from #7067 so nothing is lost when I close that one.

One note for anyone else testing this: I'm in Australia and the per-address path returns 406 for me no matter what client I use, including `curl_cffi`, while the parent page returns 200. It looks like a geo-scoped Fastly rule on that path, so a 406 from outside NZ isn't a sign the fix has failed.

PR #7129, opened 2026-08-04
fix(swale_gov_uk): preserve form state during lookup

jameswgm87:
## Summary Swale's bin-day lookup is now handled as a stateful form flow. 

The previous direct POST implementation hard-coded numeric form field names and did not load or preserve the page's hidden state. This became unreliable when Swale returned a non-results page after submission. 

This change: 
- Fetches the public bin-day form through the existing `curl_cffi` session. 
- Preserves hidden fields, including the page token, across both submissions. 
- Discovers the live postcode, address/UPRN, submit-control, method, and action fields from returned forms. 
- Retains the existing collection-result selectors. 
- Raises descriptive errors for Cloudflare challenges, validation errors, returned input forms, and missing controls. 
- Adds mocked coverage for successful stateful lookup and error paths. 

- ## Verification 
- A live Swale refresh completed successfully after reloading the integration. 
- The calendar and configured Recycling and Refuse sensors populated correctly. 

- ## Local Test Status 
- Python syntax compilation passed for both changed files. Pytest and Ruff are unavailable in the current container, so no dependencies were installed into Home Assistant. 
- GitHub Actions should run the repository's Ruff, pre-commit, and pytest validation after the branch is pushed.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/swale_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/swale_gov_uk.py
@@ -1,6 +1,8 @@
 import logging
+import re
 from datetime import date, datetime, timedelta
 from time import sleep
+from urllib.parse import urljoin
 
 from bs4 import BeautifulSoup
 from curl_cffi import requests
@@ -12,7 +14,7 @@ TITLE = "Swale Borough Council"
 DESCRIPTION = "Source for swale.gov.uk services for Swale, UK."
 URL = "https://swale.gov.uk"
 API_URL = (
-    "https://swale.gov.uk/bins-littering-and-the-environment/bins/my-collection-day"
+    "https://swale.gov.uk/bins-littering-and-the-environment/bins/check-your-bin-day"
 )
 ICON_MAP = {
     "Refuse": Icons.GENERAL_WASTE,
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for this — replacing the hardcoded `SQ_FORM_535662_*` / `q535679:q1` field names with live discovery from the returned form, and carrying the Squiz page token across both submissions, is a solid hardening of this source. The mocked test coverage is a nice bonus.

- `_raise_for_unexpected_page` now only treats an error container as an error when it actually contains text. `[role='alert']` matches the empty aria-live region that council pages keep permanently in the DOM, so as written it could have aborted an otherwise successful lookup on the results page.
- Dropped the duplicate `soup: BeautifulSoup` annotation in `fetch()`.
- Swapped the `"[redacted]"` placeholders in the test fixtures for the public UPRN that's already in `TEST_CASES`, so the fixtures read cleanly.

Verified after the tweaks: `ruff check`/`ruff format` clean, `tests/test_source_components.py` 35 passed, `tests/test_swale_gov_uk.py` 6 passed, and a live run of all three TEST_CASES returns entries.

One note for the record: I also ran the pre-PR version live and it currently works too, so I've treated this as a robustness refactor rather than a fix for a presently-broken source. The hardcoded numeric form IDs were a ticking time bomb either way, so this is still a clear improvement.

PR #7163, opened 2026-08-08
Fix available collection types in options flow

ahstax:
## Summary

Fix the options flow so that all collection types currently provided by the source are available when creating or editing sensors.

Previously, `get_types_of_sensors_and_customizations()` only collected types from existing customizations and sensor configuration. This could hide collection types that were returned by the source but had not yet been configured as a sensor.

The sensor configuration was also checked using `CONF_TYPE`, while the current sensor configuration stores its selected collection types in `CONF_COLLECTION_TYPES`.

This change:

- includes the collection types currently known by the coordinator/aggregator
- reads existing sensor types from `CONF_COLLECTION_TYPES`
- keeps customization types included
- removes duplicates and returns the result sorted

This was reproduced with the `awido_de` source, where additional collection types returned by the provider were not available in the sensor type selector until this change was applied.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### Test result

```text
35 passed in 6.72s
All checks passed!

### DIFF
--- a/custom_components/waste_collection_schedule/config_flow.py
+++ b/custom_components/waste_collection_schedule/config_flow.py
@@ -1169,8 +1169,8 @@ class WasteCollectionOptionsFlow(OptionsFlow):
 
         return self.async_show_form(step_id="init", data_schema=SCHEMA, errors=errors)
 
-    def get_types_of_sensors_and_customizations(self):
-        fetched_types = []
+    def get_types_of_sensors_and_customizations(self) -> list[str]:
+        fetched_types: list[str] = []
 
         coordinator: WCSCoordinator = self.hass.data.get(DOMAIN, {}).get(
             self._entry.entry_id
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Nice catch — this is a genuine bug. Sensor entries are persisted under CONF_COLLECTION_TYPES ("types") by finish() and read back that way in sensor.py, so the old `if CONF_TYPE in c` branch could never match. I checked the history and it has been wrong since the sensor reconfiguration feature landed, so there is no legacy config format to preserve here. Pulling the currently-known types off the coordinator's aggregator is also consistent with what async_step_init already does a few lines above.

I made a couple of small tweaks on top:

- Added `-> list[str]` and `fetched_types: list[str]` annotations. mypy skips the bodies of untyped functions, so the new logic was not actually being type-checked.
- Changed `sorted(set(fetched_types))` to `sorted({t for t in fetched_types if t})`. Sorting makes element comparability matter in a way the old unordered `list(set(...))` did not, so a falsy/None type would raise a TypeError. finish() already filters the same aggregator data with `if t`, so this just matches that.

pytest tests/ passes (39 passed) and ruff is clean. Thanks!

PR #7173, opened 2026-08-11
fix(ics): preserve event-local collection dates

phoenixaus:
## Summary

Preserve the collection date declared by each ICS event, independently of the host timezone. Midnight events in Brisbane, Sydney, Auckland and Berlin must stay on their intended calendar day instead of shifting to the previous UTC date.

Both `convert()` and `convert_events()` now call `icalevents.events(strict=True)`, which retains event-local date/time values instead of normalising every event to the calendar-wide timezone (UTC by default). The shared date helper extracts the occurrence's own calendar date without a second timezone conversion. The option is available in the supported minimum version, icalevents 0.2.1.

### Windows-TZID regression

Converting an expanded occurrence back to the original DTSTART tzinfo can shift it across midnight when the recurrence timezone and the embedded VTIMEZONE have different daylight-saving rules. The supplied `AUS Eastern Standard Time` fixture reproduced this on both Windows and Linux: 26 of 52 Thursdays became Wednesdays, for example 8 October 2026 became 7 October.

Keeping the occurrence's wall-clock date avoids that additional shift. “Windows timezone” refers to the name in the calendar, not the host operating system.

## Tests

- Include `test_ics_timezone.py` in `pytest.ini`; previously normal CI collected only the 39 existing tests.
- Retain the nine calendar shapes, five Linux host timezones, both conversion paths, DST transition and all-day regression cases.
- Add 20 cases covering nonrecurring events and explicit UTC events whose date differs from the calendar-wide timezone.
- On Windows, where `time.tzset()` is unavailable, run every calendar shape under the native host timezone. Linux continues to exercise the full five-zone matrix.
- Restore the process timezone even if a test fails.

### Validation

| Runtime | icalevents | icalendar | Timezone tests |
| --- | --- | --- | --- |
| Windows 11 / Python 3.13 | 0.3.1 | 7.3.0 | 40 passed |
| Debian / WSL2 / Python 3.11 | 0.3.1 | 7.3.0 | 112 passed |
| Windows 11 / Python 3.13 | 0.2.1 | 7.3.0 | 40 passed |
| Debian / WSL2 / Python 3.11 | 0.2.1 | 7.3.0 | 112 passed |

The tests above were run with the ICS module loaded in isolation from Home Assistant. The old PR implementation still fails the Windows-TZID regression in the same harness. Ruff check and format passed using the repository configuration. For final commit eb9399d, the full PR CI passed on both Python 3.12 and Python 3.14: **158 tests passed in each job**, including the timezone tests. Ruff, all applicable pre-commit hooks (including mypy), and hassfest passed. CI run: https://github.com/mampfes/hacs_waste_collection_schedule/actions/runs/34720179250

With strict parsing, custom title templates also receive the original event date/time types rather than calendar-normalised datetimes.

## Release branch

`release/3.0.0` at `2d6f2ceb53843ee57183896b8ca263d1de60640c` still uses the old date extraction in both conversion paths. Its newer parsers call those same methods, so the final fix and tests are needed there as well. This PR does not modify the release branch.

## Type of change

- [x] Bug fix
- [x] Regression tests and CI test discovery
- [x] No generated files changed

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
@@ -30,6 +30,50 @@ def _event_location_description(e: Any) -> tuple[str | None, str | None]:
     return loc, desc
 
 
+def _event_start_date(e: Any) -> datetime.date | None:
+    """Return the calendar date an event starts on.
+
+    icalevents normalises event start times to UTC. Reducing those straight to a
+    date shifts the day for any feed whose DTSTART sits close to midnight in a
+    timezone other than UTC: ``DTSTART;TZID=Australia/Brisbane:20251002T000000``
+    arrives as ``14:00Z`` on the *previous* day, and even
+    ``DTSTART;TZID=Europe/Berlin`` is reported a day early.
+
+    The date the feed intended is the one in the timezone the feed declared, so
+    convert back into that timezone before taking the date. Deliberately not
+    ``astimezone()`` without an argument: the host/Home Assistant timezone is
+    unrelated to the calendar, and using it both fails to fix the bug when the
+    host runs in UTC and breaks all-day events when the host is behind UTC.
+    """
+    start = getattr(e, "start", None)
+
+    if not isinstance(start, datetime.datetime):
+        # Plain date (all-day) - already the value we want.
+        return start if isinstance(start, datetime.date) else None
+
+    if start.tzinfo is None:
+        # Floating time: the wall clock date is authoritative.
+        return start.date()
+
+    # Recover the timezone declared by the original DTSTART property. icalendar
+    # has already resolved TZID to a tzinfo object, which also handles Windows
+    # style zone names such as "AUS Eastern Standard Time" that ZoneInfo would
+    # reject.
+    component = getattr(e, "component", None)
+    dtstart_prop = component.get("DTSTART") if component is not None else None
+    original = getattr(dtstart_prop, "dt", None)
+
+    if isinstance(original, datetime.datetime):
+        if original.tzinfo is not None:
+            return start.astimezone(original.tzinfo).date()
+    elif isinstance(original, datetime.date):
+        # VALUE=DATE all-day event. icalevents still hands back a UTC midnight
+        # datetime, so its UTC date is the literal date from the feed.
+        return start.date()
+
+    return start.date()
+
+
 class ICS:
     def __init__(
         self,
@@ -106,16 +150,7 @@ class ICS:
 
         for e in events:
             # calculate date
-            dtstart: datetime.date | None = None
-
-            if isinstance(e.start, datetime.datetime):
-                # icalevents normalises event times to UTC. Reducing straight to
-                # a date loses a day for timezones ahead of UTC: a DTSTART of
-                # midnight Australia/Brisbane arrives as 14:00Z on the previous
-                # day. Convert back to local time before taking the date.
-                dtstart = e.start.astimezone().date()
-            elif isinstance(e.start, datetime.date):
-                dtstart = e.start
+            dtstart: datetime.date | None = _event_start_date(e)
 
             # Only continue if a start date can be found in the entry
             if dtstart is not None:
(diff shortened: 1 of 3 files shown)


bbr111:
I agree with the review comment here. For waste collection schedules, the important value is the calendar date intended by the ICS source, not the local date in the Home Assistant/host timezone.

So I think the conversion should ideally be independent of the HA timezone:

VALUE=DATE: keep the date unchanged
DATETIME with TZID: convert the parsed datetime back to that original TZID before calling .date()
datetime without a source TZID: don't run it through the host timezone

Something along these lines, assuming e.component["DTSTART"] exposes the original TZID:

```python
if isinstance(e.start, datetime.datetime):
    dtstart_prop = e.component.get("DTSTART")
    tzid = (
        dtstart_prop.params.get("TZID")
        if dtstart_prop is not None
        else None
    )

    if tzid and e.start.tzinfo is not None:
        dtstart = e.start.astimezone(ZoneInfo(str(tzid))).date()
    else:
        dtstart = e.start.date()
elif isinstance(e.start, datetime.date):
    dtstart = e.start
```
This would also mean that an Australian waste calendar gives the same collection date whether Home Assistant itself runs in UTC, Europe/Berlin, or another timezone.

bbr111 (approved):
Thanks for this — the diagnosis is spot on and it's a genuinely nasty one to find, since there's no error in the log and the only symptom is quietly wrong dates. Confirmed and reproduced.

I've reworked the fix itself, because `astimezone()` without an argument turned out to have two problems:

1. It doesn't fix the bug when Home Assistant runs in UTC, which is the container default. `astimezone()` uses the *host* zone, so under `TZ=UTC` it's a no-op and your Brisbane example still comes back as Wednesday. It worked in your testing because your host is `Australia/Brisbane`.

2. It regresses all-day events on any host behind UTC. The `isinstance(e.start, datetime.datetime)` check runs first, and icalevents hands back `VALUE=DATE` events as UTC-midnight datetimes — so all-day events go down the datetime branch. Under `TZ=America/Los_Angeles` every one of my test cases shifted a day early, including plain all-day feeds, which is most of the ~600 sources.

The waste collection date we want is the one the feed intended, which shouldn't depend on where the HA instance happens to run. So the date is now resolved against the timezone declared by the feed's own `DTSTART;TZID`, pulled from `e.component`. Both call sites now share a single `_event_start_date()` helper.

One detail worth mentioning: I used the tzinfo icalendar already resolved from `DTSTART` rather than `ZoneInfo(TZID)`, because `ZoneInfo` raises on Windows style zone names like `AUS Eastern Standard Time` — which real Australian council feeds do emit.

Also: this isn't AU/NZ specific. `DTSTART;TZID=Europe/Berlin:...T000000` was reported a day early too, so this was affecting German feeds that use TZID datetimes rather than `VALUE=DATE`.

Added `tests/test_ics_timezone.py` to lock it in: 9 calendar shapes x 5 host timezones, plus DST crossing and all-day cases. It fails on master and on the original version of this PR, and passes now.

Thanks again for tracking down the root cause and for the self-contained reproduction — that made this much quicker to verify.

bbr111:
@markvp could you please double check?

bbr111 (approved):
Re-reviewed after the rework. Confirming the final state, since the approach changed twice on this branch.

The `strict=True` route is the right call. It resolves the Copilot finding — host timezone no longer participates at all — and it avoids the problem with my earlier `ZoneInfo(TZID)` suggestion, which shifts recurring Windows-TZID events across midnight when the recurrence zone and the embedded `VTIMEZONE` disagree on DST. Good catch on `ZoneInfo` raising on names like `AUS Eastern Standard Time`; real council feeds do emit those.

Verified locally:
- `strict` is present in icalevents 0.2.1, our pinned minimum — checked the sdist directly, so no dependency bump is needed.
- No regression on live feeds: captured every date emitted by `test_sources.py -s ics -l` under this branch and under master. Byte-identical.
- `pytest.ini` picking up `test_ics_timezone.py` takes the gating suite from 39 to 151 tests.
- ruff clean, structural tests 35 passed, CI green on all three jobs.

I also dry-ran the backport onto `release/3.0.0`: `ICS.py` applies cleanly (the divergence there is additive and the target lines are untouched), only `pytest.ini` needs a one-line manual merge into its longer allowlist. On that branch the timezone tests give 112 passed and the full gating suite 7671 passed, including the offline ICS fixtures. I'll open the backport PR once this lands.

Thanks again to @phoenixaus for the original diagnosis and the self-contained reproduction — quietly-wrong dates with no log error are the worst kind of bug to find.

PR #7175, opened 2026-08-12
add Bloomington IL and removing dead https://api-city.recyclecoach.co…

1mckenna:
…m/geo endpoint to fix adding new

## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,9 +72,9 @@ If you already have enough information for your municipality/region, you are ver
 - [Burwood City Council](/doc/source/impactapps_com_au.md) / burwood.nsw.gov.au
 - [Byron Shire Council](/doc/source/byron_nsw_gov_au.md) / byron.nsw.gov.au/Residential-Services/Waste-Recycling/Bin-Collection-Services/Bin-Collection-Schedules
 - [Cairns Regional Council](/doc/source/cairns_qld_gov_au.md) / cairns.qld.gov.au
+- [Campbelltown City Council](/doc/source/impactapps_com_au.md) / campbelltown.nsw.gov.au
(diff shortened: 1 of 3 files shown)


bbr111:
sources.json and README File are autogen files.
you need to edit the source files

bbr111 (approved):
Thanks for tracking this down — I confirmed the api-city.recyclecoach.com host is genuinely dead (it 301s to us-api-city.recyclecoach.com, which refuses connections on 443), and the us-web/ca-web endpoints return byte-identical payloads. Good catch. The city/search endpoint you left on api-city is still alive, so that's correctly untouched.

I made a couple of small tweaks:

- Reverted README.md and sources.json — those are generated by update_docu_links.py and CI regenerates them post-merge, so they shouldn't be in the diff.
- The Bloomington test case was failing live. zone-z11214 is a valid zone id from the address lookup, but the backend has no data for it: /zone/collections returns an empty types dict and /zone/schedules returns the bare string "This pattern is not exist in database", which then crashes on coll_root["types"]. Switched to zone-z11025, which returns a full schedule (351 entries), and quoted project_id to match the other test cases.

Full live run is now 26/27 green. The one remaining failure, "Richmond Hill, Ontario, Canada", is pre-existing on master and unrelated — the provider no longer matches a bare street name ("MOONLIGHT LANE"), only numbered addresses. Both the old and new endpoints behave identically there, so it's not a regression from this PR; worth a separate one-liner.

tests/test_source_components.py passes (35 passed). Approving.

PR #7185, opened 2026-08-14
Fix Ipswich City Council WhatBinDay API integration

compiled-somehow:
## Description

Fix the Ipswich City Council source following changes to the WhatBinDay API.

The existing integration relied on the legacy GET `/api/search` interface with a manually constructed address. This no longer resolves correctly for some Ipswich addresses.

This change:

- Uses ArcGIS World Geocoder to resolve the configured street/suburb.
- Obtains the postcode and normalized address components automatically.
- Sends the address in the format expected by the current WhatBinDay `/api/search` endpoint.
- Preserves the existing `street` and `suburb` configuration interface.
- Supports the current `GO Bin` collection name.
- Handles both the existing ISO date format and the current WhatBinDay agenda date format.

## Testing

Validated against the existing `TEST_CASES`:

- Camira State School
- 50 Brisbane Road, Redbank

Both returned six collection entries.

Automated test suite:

39 passed

The updated source was also validated in a live Home Assistant installation and returned the expected collection schedule.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ipswich_qld_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ipswich_qld_gov_au.py
@@ -4,6 +4,7 @@ from html.parser import HTMLParser
 
 import requests
 from waste_collection_schedule import Collection, Icons
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Ipswich City Council"
 DESCRIPTION = "Source for Ipswich City Council rubbish collection."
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for tracking down the WhatBinDay API change — the ArcGIS geocode step plus the POST payload is the right approach, and I verified it live against both TEST_CASES (6 entries each, correct GO Bin / Waste Bin / Recycle Bin mapping).

I pushed one small tweak on top: the two geocoder failure paths returned an empty list, which Home Assistant shows as "no upcoming collections" rather than telling the user their street/suburb is wrong. Both now raise SourceArgumentNotFound, consistent with bassendean_wa_gov_au.py. That resolves the Copilot review comment.

tests/test_source_components.py passes (35 passed). Approving.

PR #7187, opened 2026-08-15
Add source for Central Otago District Council

soasmileynz:
Adds support for Central Otago District Council (New Zealand) rubbish and recycling collection schedules.

Uses the same Environz API backend as the existing Dunedin District Council source (`dunedin_govt_nz.py`), with CODC's own API key and endpoint path (`/api/codc/nextservicedate`).

**Testing**
- `pytest tests/test_source_components.py`: 35 passed
- `test_sources.py -s codc_govt_nz`: 117 entries found for the Alexandra test case (5 Campbell Street Alexandra), correctly split across all four bin types (yellow/red/blue/green)
- `ruff`, `mypy`, `bandit`, `codespell`: clean on the new files

- [x] `codc_govt_nz.py` added
- [x] `doc/source/codc_govt_nz.md` added
- [x] `test_sources.py -s codc_govt_nz` passes locally
- [x] `pytest tests/test_source_components.py` passes locally

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/codc_govt_nz.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/codc_govt_nz.py
@@ -12,6 +12,23 @@ URL = "https://www.codc.govt.nz/"
 TEST_CASES = {
     "Alexandra": {"address": "5 Campbell Street Alexandra"},
 }
+
+HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
+    "en": "Use the CODC Bin App (https://play.google.com/store/apps/details?id=nz.co.environz.codc) and search for your address. The address argument should match how the app displays your address alongside your next collection details.",
+}
+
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for this — nice clean source, and good call on using SourceArgumentNotFound rather than the bare Exception the sibling Dunedin source still has.

Verified locally: 117 entries for the Alexandra test case, correctly split across all four bin types, and tests/test_source_components.py passes (35 passed).

- Added PARAM_TRANSLATIONS / PARAM_DESCRIPTIONS / HOW_TO_GET_ARGUMENTS_DESCRIPTION (en) so the config-flow UI shows a proper label and help text for the address argument instead of a bare field.
- Added timeout=30 to the API request so a hung provider can't stall the HA update loop.
- Removed a stray trailing "---" at the end of doc/source/codc_govt_nz.md.

Nothing for you to do. Merging.

PR #7180, opened 2026-08-12
Fix City of Yarra v3 port and release discovery

yeaaaaaahh:
## What changed

- Added an opt-in `require_all` mode to `ArcGisMultiFeatureRetriever`; its default remains tolerant for existing sources.
- Made City of Yarra require both waste and glass layers and keep only the first polygon match per layer.
- Restored the legacy waste-before-glass layer order.
- Fail closed when required Yarra weekday, anchor, frequency, or layer-label data is absent or invalid.
- Parse numeric ArcGIS timestamps in UTC and reject booleans/non-integral frequencies.
- Preserve collections shifted onto today after Good Friday or Christmas.
- Apply Yarra's published cascading Christmas delay through the remaining weekday rounds (including Thursday 25 -> Friday 26 and Friday 26 -> Saturday 27 in 2025).
- Added exact cassette assertions for all four canonical waste types, 143-event totals, first dates, and 7/14/28-day cadence for both recorded addresses.
- Run the supported documentation/catalogue generator on `release/**` pushes so v3 release branches generate runtime discovery artefacts after merge.

## Why

The initial v3 Yarra port inherited tolerant multi-layer defaults. A failed/empty layer or malformed provider field could therefore produce a plausible but incomplete calendar, and overlapping polygons could duplicate schedules. The release branch also contains the source but not its generated `sources.json` entry, so a HACS install cannot select City of Yarra in config flow.

The holiday correction follows City of Yarra's [general collection rule](https://www.yarracity.vic.gov.au/residents/bins-waste-recycling-and-cleansing/recycling-and-rubbish/bin-collection) and its [published 2025 Christmas timetable](https://www.yarracity.vic.gov.au/about-us/news-and-media/council-services-christmas-and-holiday-opening-hours-2025).

## User impact

- Provider failures no longer silently hide Glass, General Waste, Organic Waste, or Recycling.
- Boundary overlaps no longer create duplicate schedules.
- Christmas/Good Friday collections remain visible when shifted onto the current day.
- After merge and the generator bot commit, City of Yarra becomes discoverable from a clean v3 HACS install.

This does not change v3's intentional canonical labels: `General Waste` and `Organic Waste` remain the v3 display names.

## Validation

- Focused ArcGIS/Yarra unit suite: 37 passed.
- Yarra cassette suite: 6 passed.
- Release-workflow regression: 1 passed.
[...]

PR #7194, opened 2026-08-17
fix(yarracity_vic_gov_au): handle joined ArcGIS layer attribute names

tengunmit:
## Problem

Every City of Yarra install currently fails on fetch with:

> We could not find values for the argument 'address' with the value '…', the address must be within the City of Yarra.

The council rebuilt their waste MapServer. The service's layers are now named **"Collection Zones"** (layer 0, glass) and **"Collection Day"** (layer 1, waste/recycling/FOGO), and layer 0 has been rebuilt as a **joined** ArcGIS layer (`Glass_Collection_Zones` joined with `Waste_Recycle_FOGO_Collection_Zones`), so its attributes now carry fully-qualified names:

```text
internal.gdo.Glass_Collection_Zones.anchor_date
internal.gdo.Glass_Collection_Zones.frequency_days
internal.gdo.Waste_Recycle_FOGO_Collection_Zones.recycling_anchor_date
…
```

(visible at <https://yccgis-prd.esriaustraliaonline.com.au/arcgis/rest/services/Waste_Services/CW_FC_PRD_Waste_Collection/MapServer/0?f=json>)

The source's bare-name glass query, `out_fields="anchor_date,frequency_days"`, is therefore rejected:

```text
$ curl ".../MapServer/0/query?f=json&where=1%3D1&outFields=anchor_date,frequency_days&resultRecordCount=1"
{"error":{"code":400,"message":"Failed to execute query.","details":[]}}
```

`query_feature_layer` surfaces the 400 as an `ArcGisError`, which the source maps to "address must be within the City of Yarra" — hence the misleading error for every valid address.

## Fix

- Query with `out_fields="*"` and resolve bare field names through a small `_attr` helper that accepts either bare or dot-qualified attribute keys (`key == name or key.endswith("." + name)`), so the source works under both naming schemes.
- The waste layer (layer 1, "Collection Day") still uses bare attribute names today, but it gets the same treatment so the source survives an equivalent rework of that layer. The cost is negligible — the query returns a single feature with ~14 scalar fields.
- The required `"."` separator means `…recycling_anchor_date` can never falsely satisfy a lookup for `anchor_date`.

With `out_fields="*"` the joined layer responds normally; the qualified keys resolve via `_attr`:

```text
"internal.gdo.Glass_Collection_Zones.anchor_date": "2026-07-10",
"internal.gdo.Glass_Collection_Zones.frequency_days": 28
```

## Test output

```text
$ test_sources.py -s yarracity_vic_gov_au -i -l
Testing source yarracity_vic_gov_au ...
  found 143 entries for Fitzroy Town Hall
    2026-08-19 : Rubbish [mdi:trash-can]
    2026-08-19 : FOGO [mdi:food-apple]
    2026-08-26 : Rubbish [mdi:trash-can]
    …
  found 143 entries for Richmond Town Hall
```

(143 entries per address matches the totals asserted in the cassette tests of #7180.)

Also run: `pytest tests/test_source_components.py` — 35 passed; `ruff check` / `ruff format --check` — clean. No doc changes needed: the source's arguments and usage are unchanged.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/yarracity_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/yarracity_vic_gov_au.py
@@ -70,6 +70,15 @@ _WEEKDAYS = {
 }
 
 
+def _attr(attributes: dict, name: str):
+    # Joined ArcGIS layers qualify attribute names ("<join>.<field>");
+    # match a bare field name against either scheme.
+    for key, value in attributes.items():
+        if key == name or key.endswith("." + name):
+            return value
+    raise KeyError(name)
+
+
 def _holiday_shift(d: date) -> date:
     # Council rule: collections run on all public holidays except Good Friday
     # and Christmas Day, when the collection happens one day later.
@@ -89,15 +98,21 @@ class Source:
             raise SourceArgumentNotFound("address", self._address) from e
 
         try:
+            # 2026-08: the council rebuilt the glass layer as a joined layer
+            # whose attributes carry fully-qualified names (e.g.
+            # "internal.gdo.Glass_Collection_Zones.anchor_date"), and a
+            # bare-name out_fields now fails with HTTP 400. Fetch all fields
+            # and resolve names via _attr so either scheme works on either
+            # layer.
             waste = query_feature_layer(
                 WASTE_LAYER_URL,
                 geometry=location,
-                out_fields="collection_day,recycling_anchor_date",
+                out_fields="*",
             )[0]
             glass = query_feature_layer(
                 GLASS_LAYER_URL,
                 geometry=location,
-                out_fields="anchor_date,frequency_days",
+                out_fields="*",
             )[0]
         except ArcGisError as e:
             raise SourceArgumentNotFound(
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for tracking this down — the diagnosis is spot on, and the write-up with the layer metadata and the failing bare-name query made it easy to verify.

I made one small tweak to `_attr`: it now checks for an exact key match before scanning for a `".<field>"` suffix. Behaviour is identical on the payloads the council returns today, but it removes the dependence on attribute ordering if a joined layer ever exposes both a bare `anchor_date` and a qualified `<join>.anchor_date`.

Note that `yccgis-prd.esriaustraliaonline.com.au` returns 403 to non-AU traffic (plain nginx block, not Cloudflare — `curl_cffi` impersonation doesn't help), so I couldn't reproduce the live fetch from my environment. Going with your test output, which matches the expected 143 entries per address.

`pytest tests/test_source_components.py` passes (35 passed), ruff clean.

tengunmit:
oh, I was only vaguely aware Claude had submitted this. The pull had nothing to do with the stupid human trying to remember to take his out his trash. Oh well. What a good bot

PR #7208, opened 2026-08-20
Add source: City of Greater Shepparton (greatershepparton_com_au)

### LINKED ISSUE #3812: [Source Request]: City of Greater Shepparton
### Municipality / Region

City of Greater Shepparton, Victoria Australia

https://greatershepparton.com.au/animals-environment-and-waste/waste-and-recycling/kerbside-collection

### Example Address

15 Main Rd, Arcadia 3631

### Collection Data Format

As html on a webpage

### Additional Information

Search by address or pickup zone

RedPandaDoge:
## Summary

Adds a new source for the City of Greater Shepparton (VIC, Australia). Closes #3812.

The council's bin-lookup widget isn't backed by any of the shared platforms this project already supports (OpenCities, IntraMaps, ArcGIS proxy, Recollect, etc.) - it's a bespoke in-house build calling three plain unauthenticated endpoints under `greatershepparton.com.au/external/gis-api/`:

- `address?search=...` - address geocoding/autocomplete
- `bin-zone-name-from-location?latitude=&longitude=` - resolves an address's collection zone
- `bin-zone-schedules` - serves the zone→reference-date data as a JS module (not JSON)

Since the third endpoint isn't a clean data API, `fetch()` fetches it live on every call and extracts just the resolved zone's reference-date literals with a narrow, anchored regex (never memorized into the source file), then reproduces the same weekly/fortnightly/every-28-days cadence math the council's own widget uses. This mirrors the approach already used by `boroondara_vic_gov_au.py` for a near-identical situation (another Victorian council whose schedule only exists inside a live JS file) rather than hardcoding a schedule snapshot.

Known limitation (documented in the doc page): the council also applies a temporary one-day collection shift for some zones around Christmas/New Year, whose exact scope depends on which weekday the holidays fall on that year. Modelling that would require parsing the JS's conditional/control-flow structure rather than just its data literals, which has no precedent in this codebase, so it isn't modelled - collection dates may be off by up to a day for ~2 weeks spanning the holiday period.

## How to get the arguments

Enter your full street address including suburb (e.g. `161 Welsford Street, Shepparton`), using full street-type words (`Street`, `Road`, `Square`) rather than abbreviations - the council's own address lookup only matches on the full word.

## Test cases used

All verified live against the real endpoints, spanning three different zones/days:
- `161 Welsford Street, Shepparton` (Tuesday D)
- `15 Main Road, Arcadia` (Tuesday C - the example address from #3812)
- `1-16 Lonsdale Square, Kialla` (Thursday C)

## Type of change

- [x] New source

## Checklist

- [x] No generated files in diff (README.md, info.md, sources.json, translations/*.json - CI regenerates these post-merge)
- [x] `doc/source/greatershepparton_com_au.md` created

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/greatershepparton_com_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/greatershepparton_com_au.py
@@ -50,7 +50,6 @@ PARAM_TRANSLATIONS = {
     },
 }
 
-# Add your own handle here if you're picking this source up:
 SOURCE_CODEOWNERS = ["@RedPandaDoge"]
 
 ADDRESS_API_URL = "https://greatershepparton.com.au/external/gis-api/address"
@@ -73,12 +72,10 @@ _WEEKDAYS = [
     "Sunday",
 ]
 
-# Every one of the council's 20 zones (Monday A .. Friday D) follows the same
-# collection cadence per bin colour - only the per-zone reference dates differ,
+# Every one of the council's 20 zones (Monday A .. Friday D) uses the same
+# collection cadence per bin colour; only the per-zone reference dates differ,
 # and those are fetched live from ZONE_SCHEDULES_URL on every call (see
-# _extract_reference_dates below). This mapping is stable rostering structure,
-# not schedule data, the same way every other source in this repo encodes its
-# own day-of-week/interval rules in Python while only dates come from upstream.
+# _extract_reference_dates below).
 _BIN_INTERVAL_DAYS = {
     "General Waste": 14,
     "Recycling": 14,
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Nice work - this is a well-built source. The address -> zone -> live schedule lookup is clean, the reference dates are genuinely fetched on every call rather than snapshotted, and I appreciate that you documented the Christmas/New Year one-day shift instead of quietly approximating it. Your comparison to boroondara_vic_gov_au.py is fair; if anything this is stricter, since Boroondara hardcodes its reference date and yours doesn't.

I checked for overlap with the shared platforms (OpenCities/MyArea, Recollect, RecycleCoach, IntraMaps, the ICS YAML providers) and found none - the /external/gis-api/ endpoints really are bespoke, so a standalone source is right here.

I pushed a few small tweaks:

- removed a leftover "add your own handle here" template comment above SOURCE_CODEOWNERS
- trimmed the two long rationale comment blocks, including an in-source reference to CLAUDE.md (that file is contributor/assistant guidance, and no other source module points at it). The Christmas/New Year caveat itself is unchanged and still documented in doc/source/.
- small fix in _geocode: when several candidates matched your normalised input, the ambiguity error listed every candidate the API returned rather than the narrowed matches, which made the suggestion list less useful. It now suggests the narrowed set when there is one.

tests/test_source_components.py passes (35 passed) and ruff is clean. I couldn't reach the council's endpoints from my environment to re-run the live test, so I'm trusting your verification plus CI on that. Thanks for the thorough PR description and for creating the doc page.

PR #7211, opened 2026-08-20
Fix Broxtowe Firmstep workflow

Superfluxus:
## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [X] Documentation update
- [ ] Other

## Checklist

---

Fixes Broxtowe Borough Council collection lookup after the self-service form moved from the legacy ASP flow to the new Firmstep address lookup/submit flow.

Validation:
- `python -m pytest tests/test_source_components.py -q`
- `ruff check --fix` / `ruff format`
- `test_sources.py -s broxtowe_gov_uk -l`

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/broxtowe_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/broxtowe_gov_uk.py
@@ -30,7 +30,7 @@ ICON_MAP = {
 }
 
 
-API_URL = "https://selfservice.broxtowe.gov.uk/renderform?t=217&k=9D2EF214E144EE796430597FB475C3892C43C528"
+API_URL = "https://selfservice.broxtowe.gov.uk/renderform?t=217&amp;k=9D2EF214E144EE796430597FB475C3892C43C528"
 ADDRESS_LOOKUP_URL = "https://selfservice.broxtowe.gov.uk/core/addresslookup"
 SUBMIT_URL = "https://selfservice.broxtowe.gov.uk/RenderForm"
 
@@ -80,6 +80,7 @@ class Source:
                 "FF5683lbltxt": address_label,
                 "FF5683-text": self._postcode,
             },
+            timeout=30,
         )
         r.raise_for_status()
         soup = BeautifulSoup(r.text, "html.parser")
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for tracking down the Firmstep migration — I confirmed the old ASP flow is dead and that this restores collections. Live-tested all three TEST_CASES against the provider and got 6, 7 and 7 entries respectively.

Nice reuse of the existing `service/FirmstepSelfService.py` helpers rather than reimplementing the lookup, and the doc update accurately describes the new `core/addresslookup` JSON response.

I pushed two small tweaks:

- Added `timeout=30` to the final submit POST, addressing Copilot's comment. It was a fair catch — both helper calls already default to a 30s timeout internally, and `derbyshiredales_gov_uk` / `southkesteven_gov_uk` bound their submit requests the same way.
- Fixed two leftover error strings that still said `ashford.gov.uk`. Pre-existing copy-paste, but you'd already removed the third instance so it seemed worth finishing the job.

`tests/test_source_components.py` passes (35 passed).

One thing for a possible follow-up, not blocking: the live response includes a `BROWN 240L` collection type that has no `ICON_MAP` entry, so it falls back to the default icon. I left it alone rather than guess which waste stream Broxtowe's brown bin is — if you know, a one-line addition would be a nice polish.

custom_components/waste_collection_schedule/waste_collection_schedule/source/broxtowe_gov_uk.py
```diff
@@ -26,24 +30,9 @@
...
-UPRN_ARGS = {
-    "ctl00$ScriptManager1": "ctl00$ContentPlaceHolder1$APUP_5683|ctl00$ContentPlaceHolder1$FF5683DDL",
-    "__EVENTTARGET": "ctl00$ContentPlaceHolder1$FF5683DDL",
-    "__ASYNCPOST": "true",
-}
-
-SUBMIT_ARGS = {
-    "__EVENTTARGET": "ctl00$ContentPlaceHolder1$btnSubmit",
-}
+API_URL = "https://selfservice.broxtowe.gov.uk/renderform?t=217&amp;k=9D2EF214E144EE796430597FB475C3892C43C528"
```
bbr111:
```suggestion
API_URL = "https://selfservice.broxtowe.gov.uk/renderform?t=217&k=9D2EF214E144EE796430597FB475C3892C43C528"
```
Use & instead of &amp; here. &amp; is the HTML-escaped representation and should not be used in the actual request URL.

PR #7212, opened 2026-08-20
Cancel coordinator timers on shutdown

Superfluxus:
## Type of change

- [ ] New source
- [X] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

My local council moving from ASP.NET and breaking their endpoint exposed a generic coordinator issue: failed initial setup could leave registered timers behind. Home Assistant retries setup, accumulating callbacks that later all fire together. My HA tried to reach out approx 20k times at 1am to resolve the broken endpoint.

This defers timer registration until the first successful fetch and cancels all coordinator timers, including randomized delayed fetch callbacks, on shutdown.

Validation:
- `python -m pytest tests/ -q`
- `ruff check --fix`
- `ruff format`

### DIFF
--- a/custom_components/waste_collection_schedule/wcs_coordinator.py
+++ b/custom_components/waste_collection_schedule/wcs_coordinator.py
@@ -4,7 +4,7 @@ from random import randrange
 from typing import Any
 
 import homeassistant.util.dt as dt_util
-from homeassistant.core import HomeAssistant, callback
+from homeassistant.core import CALLBACK_TYPE, HomeAssistant, callback
 from homeassistant.helpers.device_registry import DeviceEntryType, DeviceInfo
 from homeassistant.helpers.dispatcher import dispatcher_send
 from homeassistant.helpers.event import (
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Nice catch, and a clean fix. Deferring timer registration until the first successful fetch is the right call — it also clears the two long-standing `# TODO: cancel on unload` comments. I confirmed the new `async_shutdown()` override actually runs: `DataUpdateCoordinator.__init__` registers `config_entry.async_on_unload(self.async_shutdown)`, and `WCSCoordinator` is only constructed inside `async_setup_entry`, so the config entry is picked up from the context var.

Worth noting for anyone reading later: a failed first fetch no longer leaves a "retry tomorrow" tracker behind, it now falls through to `ConfigEntryNotReady` and HA's own setup backoff. That's the canonical HA pattern and better than the old once-a-day retry, so no concerns.

I pushed one small tweak: `_fetch_callback` was overwriting `_fetch_later_unsub` without cancelling the previous handle, so it now cancels first. In practice the old handle has already fired and cancelling it is a no-op, but it's the same handle-stacking pattern this PR is fixing, so it seemed worth closing off.

`pytest tests/` (41 passed), ruff and mypy all clean.

PR #7213, opened 2026-08-20
Add Municipium (it) source

fedfus:
Adds a new dedicated source for the **Municipium** app (Maggioli S.p.A.), the platform many Italian municipalities use to publish their door-to-door (*porta a porta*) waste collection calendar. Provider website: https://www.municipiumapp.it

**Why a dedicated source** — Municipium offers no ICS/webcal subscription or stable static file, so the generic ICS route doesn't apply. The data is only available through the app's JSON API (`/api/v2`), which this source consumes.

**How it works**
- Resolves the municipality by name against the platform directory (`cloud.municipiumapp.it/api/v2/municipalities`) at runtime → not limited to a fixed list of comuni.
- Reads events from the comune's subdomain (`<comune>-api.municipiumapp.it/api/v2/calendars/<id>`), filtered by a `[start, end]` epoch-second window.
- Zone selected by full name, unique partial name (`Zona 2`, `Migliuso`), or numeric calendar id; single-zone comuni need no `area`.
- Duplicate comune name across provinces, unknown comune, or multi-zone with missing/ambiguous `area` → `SourceArgument*WithSuggestions`, so the HA UI shows valid options.

**Files added**
- `custom_components/waste_collection_schedule/waste_collection_schedule/source/municipium_it.py`
- `doc/source/municipium_it.md`

No generated files included (left to post-merge CI).

**Test output** — `python test_sources.py -s municipium_it`:
```
Testing source municipium_it ...
  found 140 entries for Serrastretta (CZ) - Zona 2 Migliuso
  found 138 entries for Serrastretta (CZ) - by calendar id
  found 186 entries for Acate (RG)
```
Structural suite: `pytest tests/test_source_components.py -q` → `35 passed`.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/municipium_it.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/municipium_it.py
@@ -44,6 +44,7 @@ TEST_CASES = {
 # comuni work, because the source resolves the comune against the platform
 # directory at runtime. More comuni can be added here as they are verified.
 COMUNI = [
+    ("Acate", "RG"),
     ("Serrastretta", "CZ"),
 ]
 
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/municipium_it.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/municipium_it.py
@@ -20,6 +20,10 @@ SOURCE_CODEOWNERS = ["@fedfus"]
 # Central directory of every municipality on the platform.
 CLOUD_URL = "https://cloud.municipiumapp.it"
 
+# Network timeout (seconds) for every outbound request, so an unresponsive
+# endpoint cannot occupy a Home Assistant executor worker indefinitely.
+TIMEOUT = 30
+
 TEST_CASES = {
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Looks good — thanks for the thorough work here, and for turning around all of the Copilot feedback (timeouts, selectable suggestions, ambiguity exceptions, Italian accents, and the stray scanner-tool references) before this review.

Verified on my side:
- COUNTRY "it" valid, ICON_MAP fully on the Icons enum, PARAM_TRANSLATIONS/PARAM_DESCRIPTIONS limited to en/it (both allowlisted), exceptions all SourceArgument*.
- No generated files in the diff, and doc/source/municipium_it.md is present.
- Not covered by an existing shared platform; Municipium exposes no ICS feed, so a dedicated source is the right call.
- Live test passes: 132 / 132 / 176 entries across the three test cases.
- Structural suite: 35 passed.

I pushed one tiny tweak: added Acate (RG) to the COMUNI list, since it is already a verified TEST_CASE but wasn't being advertised via EXTRA_INFO, and updated the matching line in the doc.

Nice touch resolving the comune against the platform directory at runtime rather than pinning a fixed list. Approving.

PR #7214, opened 2026-08-20
Add source for Gmina Bochnia (Poland) - bochnia_gmina_pl

Sairento-92:
## Summary

Adds a new dedicated source provider for **Gmina Bochnia** (Poland), covering all 31 villages in the municipality (including Baczków, Damienice, Proszówki, Łapczyca, etc.).

- **Provider website**: http://bochnia-gmina.pl/p,3,harmonogram-wywozu-odpadow-komunalnych-i-zasady-segregacji
- **Waste types**: Mixed & segregated waste (`Odpady zmieszane i segregowane`), bulky and hazardous waste (`Gabaryty i niebezpieczne`).
- **Parsing**: Fetches and parses collection dates directly from the official municipal PDF schedules with fallback.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (35 passed)
- [x] `doc/source/bochnia_gmina_pl.md` created for new sources

## Validation

- `test_sources.py -s bochnia_gmina_pl -i -l` passes (23 entries found for Baczkow, 23 for Proszowki)
- `python -m pytest tests/test_source_components.py -q` passes (35 passed)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/bochnia_gmina_pl.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/bochnia_gmina_pl.py
@@ -86,11 +86,12 @@ class Source:
         self._town = town
 
     def fetch(self) -> list[Collection]:
+        from waste_collection_schedule.exceptions import SourceArgumentNotFoundWithSuggestions
+
         norm_town = normalize(self._town)
         pdf_file = TOWNS_PDF_MAP.get(norm_town)
         if not pdf_file:
-            pdf_file = f"{norm_town.capitalize()}.pdf"
-
+            raise SourceArgumentNotFoundWithSuggestions("town", self._town, sorted(TOWNS_PDF_MAP))
         url = f"http://bochnia-gmina.pl/container/{urllib.parse.quote(pdf_file)}"
         req = urllib.request.Request(url, headers={"User-Agent": "Mozilla/5.0"})
         with urllib.request.urlopen(req, timeout=15) as resp:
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for this — nicely done. The PDF parser is solid: I live-tested it against all 31 villages in the gmina (not just the three TEST_CASES) and every single one returns a full 12 months of mixed/segregated collections plus both bulky/hazardous dates. Good call replacing the hardcoded dates with the dynamic parser, and thanks for already addressing the two bot comments.

I've pushed a few small tweaks on top:

- Swapped `urllib.request` for `requests` with `raise_for_status()` — that's the convention across the rest of the sources here, and it means an upstream 404 surfaces as a clear error rather than an opaque one.
- Widened the year regex from `202\d` to `20\d{2}` so it doesn't quietly fall back to `today().year` in 2030. I checked it doesn't collide with the postal codes or phone numbers in the PDF header.
- Anchored the bulky-waste `dd.mm.yyyy` scan to the text below the `Odpady wielkogabarytowe` heading. Previously it scanned the whole page, so any stray date elsewhere would have become a phantom collection.
- Re-keyed `TOWNS_PDF_MAP` on the real Polish names, so `SourceArgumentNotFoundWithSuggestions` now suggests `Łapczyca` / `Ostrów Szlachecki` rather than the normalised `lapczyca` / `ostrow szlachecki`. Input is still diacritic- and case-insensitive, so nothing changes for users.

One optional follow-up if you fancy it: the provider also publishes `Lapczyca-wielolokalowe.pdf` and `Stanislawice-wielolokalowe.pdf` for apartment blocks, which aren't covered yet — happy to take that as a separate PR.

`tests/test_source_components.py` passes (35) and the live test still returns 23/23/22. Approving.

bbr111:
Pushed one more small commit: added `EXTRA_INFO` (derived from `TOWNS_PDF_MAP`) so all 31 villages get their own entry in the generated source listing, not just the "Supported Towns" bullet list on this doc page. Tests still pass (35).

PR #7218, opened 2026-08-22
Add Charleston, SC waste collection sources

dmkjr:
## Summary

- add Charleston County residential curbside recycling using the county's live ArcGIS route layer
- add City of Charleston garbage and trash collection using the city's live ArcGIS collection-area layer
- cover City crews, Trident Waste service areas, and the Berkeley County portion of Daniel Island
- add configuration documentation for both sources

## Data behavior

The county source returns the exact next biweekly recycling date published by Charleston County, avoiding guesses about holiday adjustments. The city source derives the weekly collection dates from the official route weekday; temporary holiday or emergency delays are documented as a limitation.

## Validation

- ruff check --fix on both source modules
- ruff format on both source modules
- live source tests: 5 representative addresses across downtown Charleston, Johns Island, and Daniel Island
- tests/test_source_components.py: 35 passed
- full pytest suite: 39 passed

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/charleston_sc_gov.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/charleston_sc_gov.py
@@ -9,7 +9,9 @@ from waste_collection_schedule.service.ArcGis import (
 )
 
 TITLE = "Charleston, SC"
-DESCRIPTION = "Source for City of Charleston, SC garbage and trash collection."
+DESCRIPTION = (
+    "Source for City of Charleston, SC garbage and trash/yard-waste collection."
+)
 URL = "https://www.charleston-sc.gov/345/Environmental-Services"
 COUNTRY = "us"
 
@@ -37,7 +39,15 @@ PARAM_TRANSLATIONS = {
     },
 }
 
-FEATURE_URL = "https://gis.charleston-sc.gov/arcgis2/rest/services/External/mapnetExternal/MapServer/10"
+MAP_SERVER = "https://gis.charleston-sc.gov/arcgis2/rest/services/External/mapnetExternal/MapServer"
+
+# Garbage and trash/yard-waste are collected on independent routes and can
+# fall on different weekdays for the same address, so each layer is queried
+# separately and surfaced as its own stream instead of being folded together.
+LAYERS = [
+    (f"{MAP_SERVER}/10", "Garbage", Icons.GENERAL_WASTE),
+    (f"{MAP_SERVER}/11", "Trash & Yard Waste", Icons.GARDEN),
+]
 
 WEEKDAYS = {
     "Monday": 0,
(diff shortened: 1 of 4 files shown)


bbr111 (changes requested):
Thanks for this — two well-structured sources, and the groundwork is solid: live ArcGIS queries throughout, `Icons` enum used correctly, `SOURCE_CODEOWNERS` set, both `doc/source/*.md` files present, `COUNTRY = "us"` valid, clean ruff, and `tests/test_source_components.py` passes (35 passed). No generated files in the diff either. One blocker before I can merge.

**Copilot's comment on `charlestoncounty_org.py` is a false positive — please disregard it.** I checked the layer metadata directly and `PickupDate` is `esriFieldTypeString`, not `esriFieldTypeDate`:

```
field: PickupDate      | esriFieldTypeString
field: created_date    | esriFieldTypeDate    <- the only genuine date fields
field: last_edited_date| esriFieldTypeDate
```

Live values are yearless strings (`'SEPTEMBER 2'`, `'SEPTEMBER 9'`), your parser handles them correctly, and a live run returns `2026-09-02 : Recycling` / `2026-09-09 : Recycling`. Do **not** switch to `epoch_ms_to_date()` — it would break a working source. Your code is right as-is.

**Copilot's comment on `charleston_sc_gov.py` is correct, and I've confirmed it.** The MapServer exposes garbage and trash as two separate layers, each with its own `DAY`:

```
10 | Garbage Collection Areas
11 | Trash & Yard Waste Collection Areas
```

Your own primary test case diverges between them:

```
=== 123 Coming St, Charleston, SC 29403
   layer 10 (garbage): DAY = TUESDAY
   layer 11 (trash):   DAY = MONDAY
```

I grid-sampled the city to check how widespread this is: of 56 sample points inside both layers, 8 (~14%) had a different garbage day than trash day (`TUESDAY→WEDNESDAY`, `MONDAY→THURSDAY`, `WEDNESDAY→TUESDAY`, `MONDAY→TUESDAY`, `TUESDAY→MONDAY`). So downtown users currently get Tuesday entries labelled `Garbage & Trash` when their trash is actually collected Monday.

Either resolution is fine by me:

1. **Query both layers** and emit two streams — e.g. `Garbage` (`Icons.GENERAL_WASTE`) from layer 10 and `Trash & Yard Waste` (`Icons.GARDEN_WASTE` or similar) from layer 11. Best coverage.
2. **Scope to garbage only** — change `t` to `"Garbage"` and update `doc/source/charleston_sc_gov.md` to say trash/yard waste is not covered.

If you go with option 1, watch out for a trap: the collector field is `Collector` on layer 10 but `COLLECTOR` on layer 11, so a shared parsing helper will silently get `None` on one of them.

Two optional nits, neither blocking:

- `charlestoncounty_org` returns a single `Collection`. The layer already gives you `PickupDay` and `Week_Day` (`WEEK 1`/`WEEK 2`), which is enough to project a full biweekly series — `ArcGis.get_next_n_dates()` in the shared service does exactly this. Much more useful calendar.
- `_parse_next_pickup` uses yearless `strptime`, which warns on modern Python (behaviour changes in 3.15) and raises on a `FEBRUARY 29` value. Anchoring with an explicit year — `datetime.strptime(f"{value} {today.year}", "%B %d %Y")` — clears both.

Please update `TEST_CASES` and the doc alongside whichever option you pick, and re-request review — happy to merge once the garbage/trash split is sorted.

bbr111 (approved):
I pushed a fix directly to this branch (maintainer edits are enabled) addressing the blocking issue from my earlier review:

**`charleston_sc_gov.py`** now queries both ArcGIS layers (10 "Garbage Collection Areas" and 11 "Trash & Yard Waste Collection Areas") separately and returns them as two distinct streams — `Garbage` and `Trash & Yard Waste` — instead of folding them into one label. An address that only falls inside one of the two layers still gets that single stream. Live-verified against all three `TEST_CASES`, including the downtown case that previously diverged (garbage Tuesday, trash Monday).

While I was in there, I also picked up the two non-blocking nits:

- **`charlestoncounty_org.py`** now projects a full biweekly series of future pickups from the published next-pickup date (via the shared `ArcGis.get_next_n_dates()` helper) instead of returning a single occurrence.

Both docs updated to match. `ruff`, `tests/test_source_components.py` (35 passed), and a live run of both sources' `TEST_CASES` all pass. Approving.

PR #7226, opened 2026-08-25
Fix Ipswich WhatBinDay lookup

CRZTFR:
Summary:`n- replace the retired console HTML GET lookup with the current Ipswich WhatBinDay V3 app API`n- resolve addresses through the council app bounded geocoder configuration`n- preserve the existing street and suburb arguments and return explicit not-found errors`n`nVerification:`n- Ruff check and format passed`n- live source cases returned 208 Camira events and 210 Redbank events`n- invalid-address control raises SourceArgumentNotFound

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ipswich_qld_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ipswich_qld_gov_au.py
@@ -28,7 +28,7 @@ ICON_MAP = {
 BIN_NAMES = {
     "WasteBin": "Waste Bin",
     "RecycleBin": "Recycle Bin",
-    "GreenBin": "GO Bin",
+    "GreenBin": "Garden Organics (GO)",
     "GlassBin": "Glass Bin",
 }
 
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for migrating this off the retired console endpoint — pulling the geocoder URL from the app config instead of embedding an API key is a nice touch, and reusing the shared WhatBinDay service is exactly right.

Verified live: both TEST_CASES return 208 entries each, and tests/test_source_components.py passes (35 passed).

I pushed two small tweaks:
- Copilot's nit on the config-failure message — the `info` fallback now yields "Unknown error" instead of an empty suffix (using `or` so an empty-string `info` is handled too, not just a missing key).
- Added HOW_TO_GET_ARGUMENTS_DESCRIPTION / PARAM_DESCRIPTIONS / PARAM_TRANSLATIONS (en), matching the other WhatBinDay-backed sources so the config-flow UI gets proper labels for `street` and `suburb`.

doc/source/ipswich_qld_gov_au.md needed no changes since the street/suburb arguments are unchanged. Good to merge.

bbr111:
Rebased this branch onto current `master` — #7185 (a different fix for the same file) was merged in the meantime and caused a conflict here. The conflict was resolved by keeping this PR's version of `ipswich_qld_gov_au.py` in full, since it migrates to the shared `WhatBinDay` service and the V3 device/config API, which supersedes the approach in #7185. File content is unchanged from what was already reviewed and approved — only the base commit moved. Lint and `tests/test_source_components.py` (35 passed) re-verified after the rebase.

PR #7234, opened 2026-08-27
adding source deponia.hu (Cursor-generated)

tothi:
## Type of change

- [X] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/deponia_hu.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/deponia_hu.py
@@ -1,0 +1,542 @@
+"""Depónia Nonprofit Kft. (Hungary) waste collection source.
+
+Schedules are published as annual PowerPoint-exported PDFs behind the Drupal
+settlement search form at https://deponia.hu/telepuleskereso. Smaller towns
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for this — nice work, and thanks for taking codeownership via SOURCE_CODEOWNERS.

Reviewed and live-tested; all five TEST_CASES fetch successfully and the results look sane (no duplicate date/type pairs, plausible weekly/fortnightly cadences, full-year coverage). Structural tests pass, ruff is clean, no generated files in the diff, doc/source/deponia_hu.md is present, and ICON_MAP uses the Icons enum throughout.

I double-checked the ssl_verify parameter since it is unusual, and it is justified: deponia.hu serves only the leaf certificate without the intermediate, so "Verify return code: 21 (unable to verify the first certificate)". Keeping the default at True with a documented opt-out is the right call, and there is plenty of precedent in the repo for the urllib3.disable_warnings pattern.

One heads-up rather than a blocker: the PDF layout parser leans on positional column matching and several regex dialects, so it may need a touch-up when Deponia publishes the next annual calendar. Since you are listed as codeowner you will get pinged automatically if a bug report names this source.

Merging as-is.

PR #7236, opened 2026-08-27
Fix castlepoint_gov_uk returning the wrong street when a name is not unique

CRZTFR:
## Problem

`_get_road_id` returns the **first** `roadID=` link on the search results page, whatever it happens to be. Several street names occur more than once in the borough, on different collection rounds, so residents of the other one silently get another street's schedule.

Live results today:

```
searchterm=High Street  -> 3 results
    BENFLEET      / HIGH STREET            / 4907
    BENFLEET      / HIGH STREET - HADLEIGH / 4909
    CANVEY ISLAND / HIGH STREET            / 4908

searchterm=Ash Road    -> 2 results
    BENFLEET      / ASH ROAD - HADLEIGH    / 4481
    CANVEY ISLAND / ASH ROAD               / 4480
```

**The existing `Ash Road` TEST_CASE is one of these**, and the two are genuinely different rounds — fetched through the live site just now:

| `street_name` | roadID | next black sack | next pink sack |
|---|---|---|---|
| `Ash Road` (before this PR) | 4481 | 28 Aug | 5 Sep |
| `Ash Road` (after this PR) | 4480 | 1 Sep | 7 Sep |

So Canvey Island's Ash Road residents have been getting Benfleet's dates, and the test case passed throughout because it only ever checked that *some* dates came back.

Two smaller problems in the same path:

- The council's search matches bare street names and returns **nothing at all** for anything else — including a leading house number. `12 Ash Road`, which is what a resident naturally types, found no street whatsoever. (It also rejects the council's own `ASH ROAD - HADLEIGH` label.)
- A failed lookup raised bare `SourceArgumentNotFound`, so the config flow had no candidates to offer even when the search had returned some.

And one unrelated latent bug found while verifying the calendar parse:

- The month headers carry a month but no year (`August`, `September`), and the year fell back to `datetime.now().year` unconditionally. The page shows the current month *and the next*, so in December the second table is January of the following year — every collection in it was dated eleven months into the past.

## Change

Results are grouped per town in the markup (`div.flex-item-town` > `h3`), which is the only thing separating two streets of the same name. The town is now kept alongside the label, and candidates are matched in tiers:

0. town-qualified form, `HIGH STREET (CANVEY ISLAND)`
1. the council's exact label, `ASH ROAD`
2. the label without its district suffix, `GIFFORD ROAD` from `GIFFORD ROAD - SOUTH BENFLEET`

The tier order is what keeps `Ash Road` deterministic: it is both the full name of a Canvey Island street and the un-suffixed form of Benfleet's `ASH ROAD - HADLEIGH`, and tier 1 settles it on the one actually called Ash Road instead of declaring it ambiguous.

Where tiers cannot separate the candidates — `High Street` matches two byte-identical `HIGH STREET` labels — it now raises `SourceArgAmbiguousWithSuggestions` listing `HIGH STREET (BENFLEET)` and `HIGH STREET (CANVEY ISLAND)`. **Those suggested values are accepted verbatim as `street_name`**, so the config flow round-trips.

The search term is progressively simplified (house number, then town suffix, then district suffix) until the register recognises it, while the original value is kept for matching. `roadID` is untouched and still bypasses the search entirely.

## Behaviour change worth flagging

Anyone configured with `street_name: Ash Road` moves from roadID 4481 to 4480. The previous answer was positional, not chosen. Benfleet's street stays reachable as `ASH ROAD - HADLEIGH` or by `roadID: 4481`. I think that is the right way round, but happy to reverse the tie-break if you disagree.

## Verification

Everything below was run against the live site. Note that `apps.castlepoint.gov.uk` sits behind a firewall that only accepts UK IP addresses and drops everything else without a reply, so this source cannot be tested from outside the UK — I used a UK vantage point, and have added a note about it to the doc page since it also affects non-UK Home Assistant instances.

- Search-results and calendar HTML captured live, and the parser exercised against those pages: exact-label vs suffix-stripped tie-break, house-number stripping, suggestion round-trip, ambiguity, and the December→January year rollover.
- Parsed dates for `GIFFORD ROAD` (4802) cross-checked against the council's own calendar page — all nine entries, including the Thu 3 Sep slip for the bank-holiday week commencing 31 August.
- All TEST_CASES resolve live, `roadID` included.
- `python -m pytest tests/test_source_components.py` — 35 passed.
- `ruff check` / `ruff format --check` — clean.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/castlepoint_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/castlepoint_gov_uk.py
@@ -5,8 +5,9 @@ import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons
 from waste_collection_schedule.exceptions import (
+    SourceArgAmbiguousWithSuggestions,
     SourceArgumentExceptionMultiple,
-    SourceArgumentNotFound,
+    SourceArgumentNotFoundWithSuggestions,
 )
 
 TITLE = "Castle Point Borough Council"
@@ -18,6 +19,16 @@ TEST_CASES = {
     "ABBOTSWOOD": {"roadID": "4448"},
     "Ash Road": {"street_name": "Ash Road"},
     "St Marys Road": {"street_name": "St Marys Road"},
+    # "Ash Road" also matches "ASH ROAD - HADLEIGH" once the district suffix is
+    # dropped; naming the suffixed street picks that one instead.
+    "Ash Road, Hadleigh": {"street_name": "ASH ROAD - HADLEIGH"},
+    # HIGH STREET exists in both Benfleet and Canvey Island, on different
+    # rounds, so bare "High Street" is genuinely ambiguous. This covers the
+    # town-qualified form the ambiguity error hands back.
+    "High Street (Canvey Island)": {"street_name": "HIGH STREET (CANVEY ISLAND)"},
+    # A house number in front of the street is what a resident naturally types;
+    # the council's own search returns nothing at all for it.
+    "12 Ash Road": {"street_name": "12 Ash Road"},
 }
 
 API_URLS = {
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Nice piece of work — the diagnosis is careful and the writeup made this easy to review. Picking the first `roadID=` link on the page was always going to be a coin toss for the streets that occur twice, and grouping by `div.flex-item-town` is the right handle since the town is genuinely the only discriminator in the markup. The tier ordering settling `Ash Road` on the street actually called Ash Road, rather than declaring the borough ambiguous, is the right call — agreed on the direction of the tie-break. The December→January year rollover is a good catch too.

I pushed one small follow-up. `_match_tiers` had a gap for streets that carry a district suffix: `PARAM_DESCRIPTIONS` and the doc page both tell people to "add the town in brackets", but for Benfleet's Ash Road the tier-0 form is `ASH ROAD - HADLEIGH (BENFLEET)`, so the natural `ASH ROAD (BENFLEET)` missed it, fell through to the un-qualified tier and silently resolved to Canvey Island's 4480 — the same wrong-street failure mode this PR fixes. I added a tier for the town-qualified un-suffixed form between tiers 0 and 1, and made the loop bound derive from the tier list instead of a hardcoded `range(3)`.

The new tier only fires when the input itself carries a bracketed town, so nothing about bare-name resolution changes. I re-ran your scenarios offline against the candidate table from your description and all of them are unchanged: `Ash Road`→4480, `ASH ROAD - HADLEIGH`→4481, `12 Ash Road`→4480, `HIGH STREET (CANVEY ISLAND)`→4908, bare `High Street` still ambiguous with the same two suggestions. `ASH ROAD (BENFLEET)` now returns 4481.

I can confirm the firewall note independently — the host times out with no reply from here, so I could not live-test, and I have relied on your verification for the HTML shapes. `tests/test_source_components.py` passes (35). Thanks also for the doc page update; the note for non-UK instances will save someone a long debugging session.

I will call out the `Ash Road` roadID change in the release notes so existing users know why their dates shifted.

PR #7241, opened 2026-08-28
adding KOSITWEST and fixing KOSITEAST in SLOVAKIA region

mynameisdominik:
## Type of change

- [x] New source
- [x] Bug fix / source fix
- [x] Documentation update
- [x] Other

## Checklist


bbr111 (approved):
Thanks for this — nice piece of work. Generalising the KOSIT EAST PDF parser into a subclassable base rather than copy-pasting it for WEST is exactly the right call.

I verified the refactor is behaviour-preserving for EAST: Adidovce still yields 49 entries and Andrejová 27, identical to master. WEST resolves all three test cases live (Vieska 26, Sap 61, Michal na Ostrove 49) with sensible dates and waste types.

- Reverted the generated files (README.md, info.md, sources.json, source_metadata.json, translations/*.json). These are produced by update_docu_links.py, which CI runs automatically after merge — they should never be in a PR diff. No harm done, just noise.
- Restored the `_extract_year` docstring in kositeast_sk.py. The base implementation still matches `ROK\s+(\d{4})` specifically; it's only your WEST override that does the generic four-digit scan, so the loosened wording was misleading.
- The old code fell back to `page_h / 2` when the y-clustering produced fewer than two clusters. `_nearest_cluster_idx(rcy, y_clusters)` has no such fallback — if clustering ever collapses to one row, every entry silently lands in the bottom half. I added a warning mirroring the existing x-cluster one so that shows up in the log rather than as quietly wrong dates.
- Added a trailing newline and an example block to doc/source/kositwest_sk.md.

One optional suggestion: consider adding `SOURCE_CODEOWNERS = ["@mynameisdominik"]` to kositwest_sk.py. It means you get pinged and auto-assigned when someone files a bug against this source. Entirely up to you — I didn't want to opt you in without asking.

pytest tests/test_source_components.py passes (35 passed). Approving.

mynameisdominik:
Thanks for the positive reception of the code; feel free to list me as a `source_codeowner` in the `kositwest` source—that way, if an error occurs, I can take a look at it.

PR #7160, opened 2026-08-07
Update mohu_bp_hu.py - Fix new API response format

maexbower:
Due to changes on the mohu website, the json object which is containing the html answer is not valid anymore. I adopted the json paths acording to the new structure.
For the street name list and the list of house numbers, the `ajax/` prefix has been removed. The result json has a complete different name `.results`.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes
- [x] `ruff check --fix` and `ruff format` run on changed source files
- [x] No generated files in diff (README.md, info.md, sources.json, translations/*.json — CI regenerates these post-merge)
- [x] `doc/source/<name>.md` created for new sources
- [x] TEST_CASES use real, publicly accessible addresses (not my own)

Issue #7114, opened 2026-08-03
[Source Defect]:  Hungarian MOHO


Weibel73:
### Source Name

mohu_bp_hu

### What's happening?

District 1204
Street: Liliom utca
Number: 7

response: The source returned an invalid response: "'ajax/publicPlaces'". Please check the provided arguments and try again.

### Error Logs

```Shell
None
```

### Source Configuration

```YAML
waste_collection_schedule:
    sources:
    - name: mohu_bp_hu
      args:
        district: 1011
        street: "Apród utca"
        house_number: 10
        verify: true
```

### Checklist

- [x] I have tested with the latest version (master branch) via HACS → 3-dot menu → Redownload → select `master`
- [x] The service provider's website is still working
- [x] I have tested with the example parameters from the documentation
- [x] I have searched past issues (open and closed) for this problem

maexbower:
This is a duplicate of   #7069

zvitkoczi:
https://github.com/mampfes/hacs_waste_collection_schedule/pull/7242

PR #7242, opened 2026-08-28
fix(mohu_bp_hu): accept old and new OctoberCMS partial keys

### LINKED ISSUE #7069: [Source Defect]: MOHU Budapest (mohu_bp_hu)
### Source Name

MOHU Budapest (mohu_bp_hu)

### What's happening?

The MOHU Budapest source (mohu_bp_hu) is failing to fetch with a KeyError: 'ajax/publicPlaces'. It looks like the response from mohubudapest.hu no longer contains the ajax/publicPlaces key that the source code expects to parse. The website itself is up and reachable normally in a browser.

### Error Logs

```Shell
Integration: Waste Collection Schedule (documentation)
First occurred: July 24, 2026 at 21:50:26 (2 occurrences)
Last logged: 01:43:40

fetch failed for source MOHU Budapest: Traceback (most recent call last):
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 187, in fetch
                                     ~~~~~~~~~~~~~~~~~~^^
  File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/mohu_bp_hu.py", line 74, in fetch
    json.loads(r.text)["ajax/publicPlaces"], features="html.parser"
    ~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^
KeyError: 'ajax/publicPlaces'
```

### Source Configuration

```YAML
Tested with both the latest and alpha, with demo address "1011 Apród utca 10"
```

### Checklist

zvitkoczi:
## Summary

MOHU Budapest changed the JSON keys in its OctoberCMS AJAX responses:

- `ajax/publicPlaces` → `.publicPlaces`
- `ajax/houseNumbers` → `.houseNumbers`
- `ajax/calSearchResults` → `.results`

#7160 already updated the source for the new keys, but released versions (e.g. v2.32.0) still ship the old parser and fail with `KeyError: 'ajax/publicPlaces'`.

This PR adds a small helper that accepts **both** key formats (new first, old as fallback), so the source keeps working across API transitions and for users still on older releases until the next tag is published.

Tested live against district `1039`, street `Királyok útja`, house number `122`.

Fixes #7069 #7114

## Type of change

- [x] Bug fix / source fix

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (not run locally — no pytest in environment)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/mohu_bp_hu.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/mohu_bp_hu.py
@@ -48,6 +48,15 @@ ICON_MAP = {
 }
 
 
+def _get_partial_html(data: dict, *keys: str) -> str:
+    """Return the first matching OctoberCMS partial HTML payload."""
+    for key in keys:
+        value = data.get(key)
+        if value is not None:
+            return value
+    raise KeyError(keys[0])
+
+
 class Source:
     def __init__(self, district, street, house_number, verify=True):
         self._district = district
@@ -70,8 +79,10 @@ class Source:
             verify=self._verify,
         )
         r.raise_for_status()
+        data = json.loads(r.text)
         soup = BeautifulSoup(
-            json.loads(r.text)[".publicPlaces"], features="html.parser"
+            _get_partial_html(data, ".publicPlaces", "ajax/publicPlaces"),
+            features="html.parser",
         )
 
         if soup.find("div", attrs={"class": "alert"}) is not None:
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks! Verified locally:

- `python -m pytest tests/test_source_components.py -q` → 35 passed
- `ruff check` / `ruff format --check` → clean, no changes needed
- Live probe against `https://mohubudapest.hu/hulladeknaptar` confirms the API currently returns `.publicPlaces`, so the fallback is a no-op today and purely defensive against a future flip — good.
- Live source test: Test_2/3/5/6 return data as expected.

Trying the new key first and raising `KeyError` on the new key when neither is present is the right ordering — error messages keep pointing at the current API contract.

Two notes, neither blocking:

1. The description mentions this helps users on older releases (v2.32.0). It can't do that retroactively — they'll need the next tag either way. The real benefit is robustness across API transitions, which is reason enough on its own.
2. Test_1/Test_4 (`district 1011, Apród utca 10`) return 0 entries — but they do so on `master` too, so it's pre-existing test-case staleness rather than anything introduced here. Worth a separate follow-up to swap in a live address.

Approving as-is.

PR #7243, opened 2026-08-28
fix(montreal_ca): seasonal ranges drop dates, and biweekly seasons parse as weekly

stradaconsulting:
`calendar.waste_green` was empty for my sector (ANJ-1, Anjou). The Green feed is the only one that uses seasonal `from X to Y` ranges, and three separate bugs stack up on it.

**1. `day_start` captured only the last digit**

```python
day_start = int(re.match(r".*(\d+).*", date_range_start).group(1))
```

The leading `.*` is greedy, so on "September 30" it consumes "September 3" and `group(1)` captures `"0"`. The arming condition is `day_start == day` with `day` iterating `range(1, 32)`, so `0` never matches, `within_dates` is never set, and the entire range produces nothing. This only bites two-digit start days, which is why it went unnoticed — "April 1" gives `1` correctly by luck.

**2. The stop condition fired on day 1 of the stop month**

`day_stop >= day` is true at `day == 1` for any `day_stop >= 1`, so `within_dates` was cleared on the first day of the final month and that whole month was lost. Inverted to `day > day_stop`.

**3. Biweekly seasons were routed into the weekly branch**

`re.search("(every )?week(ly)?", line)` matches the "week" in "every two weeks", so summer lines like

> Summer : June 10 and 24; July 8 and 22; August 5 and 19; September 2 and 16 (every two weeks on Wednesday)

took the weekly branch and emitted nothing. Rosemont writes the same thing as "every second week", so the guard covers both. Routing these to the explicit-date branch then exposed a fourth issue: `\b{month}(.*){MONTH_PATTERN}` is greedy and ran to the *last* month name in the line, so June inherited July/August/September's day numbers and invented June 19 and 22. Made non-greedy with a lookahead.

**Also:** `within_dates` is now initialised per line. It was only assigned inside the `from X to Y` branch, so a sector whose first season line is biweekly raises `NameError` — and since `fetch()` catches every exception per source, that takes out all five feeds for that sector, not just Green. It happens to be masked today because every affected sector lists a range season first.

### Verification

Against the live 2026 feeds:

| Sector | before | after |
|---|---|---|
| ANJ-1 (Anjou) | 5 dates, April only | 26 |
| RPP-RE-22-RV (Rosemont) | 4 | 26 |
| MHM-42-S (Mercier-Hochelaga) | 52 | 52 — unchanged |

ANJ-1 now matches its published schedule exactly: Apr 1–29 and May 6–27 weekly, Jun 10/24, Jul 8/22, Aug 5/19, Sep 2/16 biweekly, then Sep 30 through Nov 25 weekly.

16 Green sectors currently use the biweekly summer format.

`tests/test_montreal_ca.py` uses verbatim `MESSAGE_EN` captures so it runs offline; 6 of the 8 tests fail on master. It follows the shape of `tests/test_olo_sk.py`. Note that `pytest.ini`'s `python_files` doesn't match per-source test filenames, so neither this nor the existing per-source tests are collected by a bare `pytest` in CI — happy to address that separately if you'd like it.

`ruff check`, `ruff format --check`, and `pytest tests/test_source_components.py` (35 passed) are all clean.

### Not fixed here

Kept out to hold the diff down, all pre-existing:

- `re.split(r", | and ", ...)` drops the last date in "July 1, 15, and 29", and ordinals like "July 1st" aren't `isnumeric()` so they're skipped. Both visible in LSL4, which still yields only 3 of its 5 summer dates.
- Sectors whose message has no "every … week" phrasing take the first branch and emit every matching weekday of the year, ignoring their explicit date list (AC-2, MHM-42-S).
- The module has no `COUNTRY` variable, which per `CLAUDE.md` orphans it from the generated docs.

Happy to split any of those out if you want them in scope.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/montreal_ca.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/montreal_ca.py
@@ -101,6 +101,11 @@ FOOD_INCLUDED_IN_GREEN_PATTERN = re.compile(
     r"included in the collection of organic waste", re.IGNORECASE
 )
 
+# Biweekly seasons are written as "every two weeks", "every second week" or
+# "once every two weeks". The substring "week" also matches the weekly branch
+# below, so these lines have to be recognised and excluded from it explicitly.
+BIWEEKLY_PATTERN = re.compile(r"every\s+(?:two|second|other|2)\s+weeks?", re.IGNORECASE)
+
 LOGGER = logging.getLogger(__name__)
 HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
     "en": 'Download on your computer a &lt;a href="https://donnees.montreal.ca/dataset/2df0fa28-7a7b-46c6-912f-93b215bd201e/resource/5f3fb372-64e8-45f2-a406-f1614930305c/download/collecte-des-ordures-menageres.geojson"&gt;Montreal GeoJSON file&lt;/a&gt;&lt;br/&gt;Visit https://geojson.io/&lt;br/&gt;Click on *Open* and select the Montreal GeoJSON file&lt;br/&gt;Find your sector on the map.',
@@ -240,6 +245,7 @@ class Source:
             month_stop = 12
             day_start = 1
             day_stop = 31
+            within_dates = False
             # There could be seasonal schedules, every week, every other week or specific dates
             if re.match(r".*[fF]rom (.*) to (.*)", line):
                 date_range = re.match(r".*[fF]rom (.*) to (.*)", line)
@@ -251,10 +257,9 @@ class Source:
                     if re.search(rf"{month}", date_range_stop, re.IGNORECASE):
                         month_stop = month_id
                 if re.search(r"\d+", date_range_start):
-                    day_start = int(re.match(r".*(\d+).*", date_range_start).group(1))
+                    day_start = int(re.search(r"(\d+)", date_range_start).group(1))
                 if re.search(r"\d+", date_range_stop):
                     day_stop = int(re.search(r"\d+(?!.*\d+)", date_range_stop).group(0))
-                within_dates = False
             elif re.match(r"(.*\d+.*){1,}", line):
                 # Multiple dates ?
                 dates_defined = True
@@ -267,7 +272,9 @@ class Source:
                     continue
                 if dates_defined and month not in months_found:
                     continue
-                if re.search("(every )?week(ly)?", line):
+                if re.search(
+                    "(every )?week(ly)?", line
+                ) and not BIWEEKLY_PATTERN.search(line):
                     for day in range(1, 32):
                         try:
                             if (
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Nicely diagnosed — thank you for the write-up, it made this very easy to review.

I verified against the live feeds by pulling all five GeoJSON files and running
`parse_collection` from master and from this branch over every sector:

| Feed | sectors | identical | more dates | fewer dates | exceptions (master → PR) |
|---|---|---|---|---|---|
| Waste | 114 | 113 | 1 | 0 | 7 → 0 |
| Recycling | 101 | 101 | 0 | 0 | 0 → 0 |
| Green | 93 | 69 | 24 | 0 | 0 → 0 |
| Food | 99 | 99 | 0 | 0 | 0 → 0 |
| Bulky | 100 | 91 | 9 | 0 | 0 → 0 |

Zero sectors lose dates, 34 gain them, and your ANJ-1 / RPP / MHM numbers
reproduce exactly.

Worth calling out that the `within_dates` initialisation is doing more than you
claimed: the 7 Waste-feed exceptions on master are all
`UnboundLocalError: cannot access local variable 'within_dates'` (AC-2, AC-3,
VSMPE-2a and others). Since `fetch()` catches per feed, that was silently
zeroing the entire Waste calendar for those sectors, not just Green.

I pushed one small tweak. Collapsing the two-branch month lookup into the single
non-greedy pattern makes the old `(?:\s*{month} {year})(.*)` fallback
unreachable, and that branch was there to consume an explicit year between the
month name and its day list. Without it a line shaped "April 2026, 5 and 12"
would parse 2026 as a day, raise ValueError inside the `try`, and hit
`except Exception: break`, dropping every remaining month on the line. I folded
an optional `(?:\s+{year})?` into your pattern to preserve that. It is a no-op
on all current live data — the sweep above is byte-identical with and without it.

`ruff check`, `ruff format` and `pytest tests/test_source_components.py`
(35 passed) are clean; the full collected suite is 39 passed and your 8 offline
tests pass.

On the points you kept out of scope: all agreed, and please do open follow-ups if
you have the appetite. Two more for the list:

- `donnees.montreal.ca` now sits behind Cloudflare. Plain `requests` gets a 403
  from datacenter IPs while `curl_cffi` with `impersonate="chrome"` gets 200.
  Residential installs are probably fine, but the source should move to
  `curl_cffi` per CONTRIBUTING.
- On `pytest.ini`: forcing collection with `-o python_files="test_*.py"` already
  fails today without your file, because the existing per-source tests replace
  `sys.modules["waste_collection_schedule"]` with a stub and that breaks
  `test_source_components.py`. So enabling collection needs those tests isolated
  first. Not something to solve here.

Merging. Thanks for the thorough investigation and the offline test coverage.

stradaconsulting:
I'm very happy I could help and will definitely take on the items I listed as well as the two you outlined.

Have a good day.

PR #7248, opened 2026-08-29
fix(borlange_energi_se): handle "Tömning idag" / "imorgon"

### LINKED ISSUE #7238: [Source Defect]: parse_swedish_date crashes on "Tomning idag" / "Tomning imorgon"
### Source Name

borlange_energi_se

### What's happening?

The parse_swedish_date() function only handles dates in the form "<weekday> den <day> <month>" (matched via a regex looking for a digit followed by a month name). When Borlange Energi's API returns the special-case text "Tömning idag" (collection today) or "Tömning imorgon" (collection tomorrow) instead of a dated string, the regex finds no match and the function raises ValueError, which makes the whole fetch for that address fail (repeatedly, since it happens every polling interval on collection day).

This is a real, expected value from the API - not malformed data - so it should be handled rather than raising.

Suggested fix: check for "idag"/"imorgon" in the lowercased text before running the regex, and return today's / tomorrow's date directly in those cases.

### Error Logs

```Shell
ERROR (SyncWorker) [custom_components.waste_collection_schedule.waste_collection_schedule.source_shell] fetch failed for source Borlänge Energi:
  File ".../waste_collection_schedule/source_shell.py", line 187, in fetch
  File ".../waste_collection_schedule/source/borlange_energi_se.py", line 98, in fetch
    date = parse_swedish_date(item["disposalDay"])
  File ".../waste_collection_schedule/source/borlange_energi_se.py", line 51, in parse_swedish_date
    raise ValueError(f"Unrecognized date format: {text}")
ValueError: Unrecognized date format: Tömning idag
```

### Source Configuration

```YAML
waste_collection_schedule:
  sources:
    - name: borlange_energi_se
      args:
        pickup_address: "Example Street 1"
```

### Checklist

- [ ] I have tested with the example parameters from the documentation

Darkslayer3324j:
Fixes #7238

## Problem

Borlänge Energi's API swaps the usual dated string for relative wording on the day of a collection (`Tömning idag`) and the day before it (`Tömning imorgon`). `parse_swedish_date()` only matched `<number> <month>`, so those values reached:

```python
raise ValueError(f"Unrecognized date format: {text}")
```

That fails the whole fetch for the address, and keeps failing on every polling interval for as long as the API returns the relative wording — so it breaks precisely on the day the schedule matters most.

## Fix

These are expected API values, not malformed data, so they're resolved before the date regex runs:

```python
if re.search(r"\bi\s?dag\b", lowered):
    return midnight
if re.search(r"\bi\s?morgon\b", lowered):
    return midnight + timedelta(days=1)
```

Both spellings are accepted (`idag` / `i dag`, `imorgon` / `i morgon`), and the match is anchored on word boundaries so the seven Swedish weekday names ending in `dag` still fall through to the dated branch — I tested each one explicitly, since that was the obvious way to get this wrong.

Also raises `ValueError` rather than letting a `KeyError` escape when the regex matches a number followed by a non-month word:

```python
if month_name not in MONTHS:
    raise ValueError(f"Unrecognized month name in date: {text}")
```

The regex only requires a digit followed by any word, so text like `om 3 dagar` produced `MONTHS["dagar"]` → `KeyError`, which a caller of a function documented to raise `ValueError` wouldn't be catching.

## Verification

Added `tests/test_borlange_energi_se.py` — 18 cases covering both relative forms, all seven weekday names, the unchanged dated path, the year rollover, and the `ValueError` cases.

```
tests/test_borlange_energi_se.py     18 passed
tests/test_source_components.py      35 passed
ruff check / ruff format             clean
```

I also confirmed the new tests fail against the parser as it stands today — the four relative-form cases raise `ValueError` — so they genuinely cover the reported bug rather than just passing.

No generated files touched; the diff is the one source module plus the new test.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/borlange_energi_se.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/borlange_energi_se.py
@@ -1,5 +1,5 @@
 import re
-from datetime import datetime
+from datetime import datetime, timedelta
 
 import requests
 from waste_collection_schedule import Collection, Icons
@@ -44,14 +44,33 @@ ICON_MAP = {
 def parse_swedish_date(text: str) -> datetime:
     """Extract date from Swedish text.
 
-    Handles formats like 'Nästa tömning sker torsdag den 15 januari'.
+    Handles formats like 'Nästa tömning sker torsdag den 15 januari', as well
+    as the relative wording the API uses on the day of a collection ('Tömning
+    idag') and the day before it ('Tömning imorgon'). Both the joined and
+    separated spellings are accepted.
     """
-    match = re.search(r"(\d{1,2})\s+([a-zåäö]+)", text.lower())
+    lowered = text.lower()
+
+    # Expected API values on and just before a collection day, not malformed
+    # data, so they must not raise. Anchored on word boundaries so weekday
+    # names ending in "dag" are unaffected.
+    midnight = datetime.now().replace(hour=0, minute=0, second=0, microsecond=0)
+    if re.search(r"\bi\s?dag\b", lowered):
+        return midnight
+    if re.search(r"\bi\s?morgon\b", lowered):
+        return midnight + timedelta(days=1)
+
+    match = re.search(r"(\d{1,2})\s+([a-zåäö]+)", lowered)
     if not match:
         raise ValueError(f"Unrecognized date format: {text}")
 
     day = int(match.group(1))
     month_name = match.group(2)
+    if month_name not in MONTHS:
+        # The regex only requires a number followed by a word, so input such as
+        # "om 3 dagar" reaches here. Raise ValueError like the branch above
+        # instead of letting a KeyError escape.
+        raise ValueError(f"Unrecognized month name in date: {text}")
     month = MONTHS[month_name]
 
     year = datetime.now().year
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks — tidy, well-scoped fix.

Verified locally against your HEAD:
- `python -m pytest tests/test_source_components.py -q` → 35 passed
- `python -m pytest tests/test_borlange_energi_se.py -q` → 18 passed
- `pre-commit run --files ...` (ruff, ruff-format, codespell, bandit, mypy) → all pass
- No generated files in the diff

The word-boundary anchoring is the right call, and I re-checked it independently rather than taking it on trust: of the seven weekday names only `tisdag` contains an `i` at all, and there it has no preceding word boundary and is followed by `s`, so none of them can reach the relative branch. The extra `month_name not in MONTHS` guard is a real fix too — the regex matches any digit-then-word pair, so `om 3 dagar` used to escape as a `KeyError` out of a function documented to raise `ValueError`.

Two notes, neither blocking and both pre-existing on master — please do not widen this PR for them, I will file them separately:

1. I could not live-verify the TEST_CASES. The `appresource` endpoint in `URL` currently returns HTTP 404 while the site root returns 200, so the source looks broken upstream independently of this change — the Sitevision resource id has presumably rotated. Your fix is unit-verified but not end-to-end verified.
2. `ICON_MAP["Plastförpackningar"]` maps plastic packaging to `Icons.GLASS` (should be `Icons.PLASTIC_PACKAGING`), and `DEFAULT_ICON` is a raw `"mdi:trash-can"` string rather than `Icons.GENERAL_WASTE`.

Approving as-is.

PR #7271, opened 2026-08-30
Fix Pendle Opus4 collection request

Hyperion5088:
## Summary

- increase the Pendle GetFeatureInfo viewport to the 256-pixel minimum now accepted by Opus4
- centre the query at 128,128
- remove the feature_count parameter, which Opus4 now rejects

The council website uses a browser-sized viewport and does not send feature_count. Before this change the source received HTTP 500; the adjusted request returns the four current collections for the public test case.

## Validation

- `python test_sources.py -s pendle_gov_uk -l -i -t` -> 4 entries
- `ruff check custom_components/waste_collection_schedule/waste_collection_schedule/source/pendle_gov_uk.py`
- `ruff format --check custom_components/waste_collection_schedule/waste_collection_schedule/source/pendle_gov_uk.py`
- `git diff --check`


bbr111 (approved):
Thanks for the fix, and for the clear write-up of what changed and why.

I reproduced this end to end. On master the source is currently dead — the test case fails with `500 Server Error` from the Opus4 WMS. With your branch it returns the expected 4 entries.

I also probed the endpoint directly to confirm both halves of the change are independently required, rather than one masking the other:

| Request | Result |
| --- | --- |
| 256x256, x/y=128, no `feature_count` (this PR) | 200, 4 features |
| 256x256, x/y=128, **with** `feature_count=10` | 500 |
| 101x101, x/y=50, no `feature_count` | 500 |
| 255x255, x/y=127, no `feature_count` | 500 |
| 101x101, x/y=50, `feature_count=10` (pre-PR) | 500 |

So `feature_count` really is rejected outright now, and 256 is a genuine minimum — 255 still 500s. Nice diagnosis.

Also confirmed the query pixel stays at the bbox centre (128/256 == 50/101), so the 500/1000/1500 buffer retry loop still samples the same ground point.

`ruff check`, `ruff format --check` and `pytest tests/test_source_components.py` (35 passed) are all green. No changes needed from my side.

PR #7274, opened 2026-08-30
fix(montreal_ca): do not expand a whole year of weekdays over an explicit date list

stradaconsulting:
Last of the follow-ups from #7243, and the one with real user impact.

### The bug

`parse_collection` opens with:

```python
if not re.search(r"(?:every\s+(?:.*)week|of the month)", schedule_message, re.IGNORECASE):
    # emit every matching weekday of the whole calendar year
    return entries
```

A message that states its schedule as a plain list of dates matches neither phrase, so it takes that branch and its list is never read.

28 Green sectors publish exactly that shape. AC-2 lists 29 Thursdays from April to November:

> Collection day(s) in 2026 : Thursday
> - April 9, 16, 23 and 30;
> - May 7, 14, 21 and 28;
> ...

and produces **all 53 Thursdays of 2026**, including January through March and December, when there is no green collection at all.

This is the opposite failure to #7243 and #7256. Those dropped dates. This one invents them, which is worse: the calendar asks for bins on days with no pickup, and any staleness check looking for gaps is quietly satisfied.

### The fix

Route the message to the date parser when its body enumerates days across four or more distinct months.

Two details that needed measuring rather than guessing.

**Counted over the hyphen-delimited body, not the whole message.** The date parser reads only what follows the first hyphen, since `header` is popped. MTN-1 to MTN-4 and OUT-RV list their days *ahead* of any hyphen, so rerouting on the raw text would hand them to a parser that cannot see them, turning 52 wrong dates into none. Counting the body leaves them alone.

**Threshold of four.** Across all 507 sector messages, distinct months per body distributes as:

| months in body | sectors |
|---|---|
| 0 | 388 |
| 1 | 4 |
| 2 | 2 |
| 8 | 23 |
| 9 | 5 |

Nothing sits between 2 and 8, so any threshold in that range selects the same 28 sectors. The number is not a tuning knob. Prose that merely mentions a date, such as *"Since May 2, 2024, food waste collection has been deployed..."*, lands in the low band and is untouched.

### A second fault this exposed

SLE-1 writes its whole schedule on one line, so the last month runs into prose:

> - November 4, 11 and 18. Hours : Take out containers between 7 p.m. the evening before and 7 a.m. ...

Rerouting it produced a **November 7**, a Saturday, read out of the clock times. The first-word-per-token approach hid this: `"18 Hours : Take out ... before"` yields `"18"`, and the following token starts `"7 am the day of collection"`.

A day list is now read as a contiguous run of bare numbers, ending at the first token that carries words after its number. Leading non-numeric tokens are skipped rather than ending the list, so `"July 1st, 15, and 29"` still yields 15 and 29 exactly as it does today.

### Verification

Against the live feeds, one shape per family. Every rerouted sector was checked date by date against its published list, not just by count.

```
sector        before  after   result
---------------------------------------------------------------
Green/AC-2        53     29    matches published list exactly
Green/MHM-42-S    52     26    matches published list exactly
Green/SO-1        52     25    matches published list exactly
Green/SLE-1       52     30    matches published list exactly
Green/VRD-1       52     25    published 26, missing August 31st

Green/MTN-1       52     52    unchanged, dates precede any hyphen
Green/OUT-RV      52     52    unchanged, same reason
Waste/AC-1        52     52    unchanged, prose date only
Waste/SLR-1       52     52    unchanged, two months in body
Green/ANJ-1       26     26    unchanged, seasonal path untouched
Green/RPP-RE-22-RV 26     26   unchanged
Green/LSL4         4      4    unchanged
```

The 28 rerouted sectors are five shape families: AC-1..4, SO-1..4 and SO-5-R1..R4, MHM-41 through MHM-44, SLE-1..4, VRD-1..5. Members of a family differ only in weekday and day numbers, and one of each is verified above.

VRD-1's August 31st is the ordinal case deliberately left unparsed in #7256. Unchanged here.

Three tests added. Two fail on master; the third is a guard that passes on both, which is the point of it. `ruff check` and `ruff format --check` clean, `tests/test_montreal_ca.py` 14 passed, `tests/test_source_components.py` 35 passed.

### Still not addressed

`Green/CD-E-*` and `Green/CD-R-*` say "Wednesday **from April 22 to November 25**", a range with no "every ... week" phrasing, so they also take the whole-year branch and over-emit. That would require routing to the range parser rather than the list parser, and deserves its own measurement. Not touched here.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/montreal_ca.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/montreal_ca.py
@@ -107,6 +107,31 @@ FOOD_INCLUDED_IN_GREEN_PATTERN = re.compile(
 # below, so these lines have to be recognised and excluded from it explicitly.
 BIWEEKLY_PATTERN = re.compile(r"every\s+(?:two|second|other|2)\s+weeks?", re.IGNORECASE)
 
+# A sector that enumerates collection days across several months has
+# published a schedule, not prose that happens to mention a date. Such a
+# message must not take the whole-year weekday expansion, which would
+# invent a collection on every remaining weekday of the year.
+#
+# Counted over the hyphen-delimited body only, because that is what the
+# date parser actually reads. A message whose dates sit ahead of the first
+# hyphen would otherwise be handed to a parser that cannot see them, and
+# would yield nothing at all.
+EXPLICIT_DATE_LIST_MIN_MONTHS = 4
+MONTH_WITH_DAY_PATTERN = re.compile(
+    rf"\b({'|'.join(MONTHS)})\s+\d{{1,2}}\b", re.IGNORECASE
+)
+
+
+def enumerates_dates_across_months(schedule_message):
+    """Whether the parsed body lists days in several distinct months."""
+    months = {
+        match.group(1).lower()
+        for line in schedule_message.split("-")[1:]
+        for match in MONTH_WITH_DAY_PATTERN.finditer(line)
+    }
+    return len(months) >= EXPLICIT_DATE_LIST_MIN_MONTHS
+
+
 LOGGER = logging.getLogger(__name__)
 HOW_TO_GET_ARGUMENTS_DESCRIPTION = {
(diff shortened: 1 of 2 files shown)


bbr111:
@stradaconsulting check merge conflicts

stradaconsulting:
Sorry about that.  Now rebased onto master, conflict was a duplicate of the already-merged #7256 commit from a stale branch base. Now a single commit, mergeable.

bbr111 (approved):

PR #7285, opened 2026-09-01
Fix Newham date parsing, add food waste collections, and fail loudly on bad property IDs

ryanharkins:
### 1. Dates were parsed month-first

Newham recently changed date renders back to UK date format `dd/mm/yyyy`, but the source used `dateutil.parser`, which defaults to month-first. So `03/09/2026` was read as **9 March** instead of **3 September**. Every date with a day of 12 or lower was breaking.

Now parsed with an explicit `datetime.strptime(..., "%d/%m/%Y")`.

### 2. Include newly added "Food waste" collections

The card selectors were hardcoded to the classes `card h-100` and `card h-100 card-recycling`. Newham's food waste card is `card h-100 card-food`, so it was skipped entirely.

Replaced with a scan over every `card`, keyed on the bold bin type in the card header, so the food card is picked up and any future card shape keeps working.

Also switched the recycling icon from `Icons.GLASS` to `Icons.RECYCLING`.

### 3. An unknown property ID silently returned nothing

A well-formed but non-existent property ID returns **HTTP 200** with a fully rendered page — `raise_for_status()` passes, every collection card says "no details available", and `fetch()` returned `[]`. The user got an empty sensor and no explanation.

Only the address card is left blank in that case, so the check keys on that rather than on an empty result — a valid property can legitimately return fewer entries when it has no food waste service. Raises `SourceArgumentNotFound` so the HA UI names the argument at fault.

### 4. Test
Added `tests/test_newham_gov_uk.py` — offline, fixtures copied from live responses (including the non-breaking spaces and the empty `<mark>` a card renders when it has no next collection). Covers day-first parsing, the food card, a property without food waste service, and the unknown-ID path.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/newham_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/newham_gov_uk.py
@@ -5,6 +5,7 @@ import requests
 import urllib3
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "London Borough of Newham"
 DESCRIPTION = "Source for newham.gov.uk services for London Borough of Newham, UK."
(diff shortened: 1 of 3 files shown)


bbr111 (approved):
Thanks Ryan — this is a really well-put-together fix.

Verified the three behaviours against the live site: all three TEST_CASES now return future-dated entries and Test_002 correctly picks up the new Food Waste card. Day-first parsing is confirmed by the 14/16 September results, which month-first would have mangled.

Nice touches worth calling out:
- keying the unknown-property check on the empty *address* card rather than on an empty result set, so a legitimate property with no food waste service isn't falsely rejected;
- the negative lookahead in NEXT_DATE that stops an empty "Next" falling through to the "Previous" date — the parametrised boundary test for that is exactly the right level of coverage;
- using ICON_MAP as the type allowlist, and Icons.RECYCLING over Icons.GLASS.

Both Copilot notes (request timeout, isort:skip on the sys.path-shim imports) are already addressed in your follow-up commits.

I pushed one tiny unrelated tweak: fixed a "Searach" typo and a malformed link in doc/source/newham_gov_uk.md while I was in there.

ruff, mypy, bandit, codespell and pytest tests/ (39 passed, including the 35 structural checks) all green. Approving.

PR #7292, opened 2026-09-02
fix(eastdevon_gov_uk): resolve postcode + address to a UPRN

CRZTFR:
## Summary

`eastdevon_gov_uk` uses the shared `uk_cloud9_apps.Cloud9Client`, which already
exposes `fetch_by_address`, but the source only ever wired up `fetch_by_uprn`.
That made a 12-digit UPRN the only way to configure it.

This accepts `postcode` + `address` as an alternative, mirroring `arun_gov_uk`
and `southend_gov_uk`, which use the same client and already offer both forms.

## Why

I run a public bin-day lookup that uses this source. Over 2026-08-26 to
2026-09-01, 94 of 101 completed East Devon lookups returned an **empty
schedule**: a UPRN that a person guesses or mistypes resolves to no collection
data rather than to an error, so there is no useful feedback. Offering the
address path removes the guessing.

## Compatibility

- Existing `uprn` configurations are unchanged, including the 12-digit zero
  padding (`str(uprn).zfill(12)`).
- All three arguments are optional; `uprn` wins when supplied. Passing neither
  `uprn` nor `postcode` raises `SourceArgumentRequired`.
- An ambiguous address surfaces the client's existing
  `SourceArgAmbiguousWithSuggestions`, so the user gets a candidate list.

## Tests

`TEST_CASES` keeps both existing UPRN cases and adds two address cases:

```
Test_001 {"uprn": "010000246114"}                                -> 3 collections
Test_002 {"uprn": 10000272679}                                   -> 4 collections
Test_003 {"postcode": "EX8 2AN", "address": "1 Dagmar Road"}     -> 4 collections
Test_004 {"postcode": "EX5 2AB", "address": "1 Blackhorse Cottages"} -> 3 collections
```

All four pass against the live API.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/eastdevon_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/eastdevon_gov_uk.py
@@ -1,4 +1,5 @@
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import SourceArgumentRequired
 from waste_collection_schedule.service.uk_cloud9_apps import Cloud9Client
 
 TITLE = "East Devon District Council"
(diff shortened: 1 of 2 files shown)


CRZTFR:

`argument_name` is now `"address"` rather than `"postcode"`. The client's suggestions are whole address strings and the value it reports is the address string it was given, so both belong to that field; picking a candidate refines `address` instead of overwriting the postcode with a full address.

Verified live: all four TEST_CASES pass, and `postcode=EX8 2AN, address=999 Nowhere Road` now raises `SourceArgAmbiguousWithSuggestions` against `address`.

bbr111 (approved):
Looks good. Verified locally against the live API: all four `TEST_CASES` pass, existing `uprn` configs are unaffected (including the `zfill(12)` padding), and `uprn` correctly takes precedence when both forms are supplied.

The Copilot thread about `argument_name` is properly addressed. `Cloud9Client._select_address` raises with `value=address_string`, so `argument_name="address"` makes the argument and the reported value agree — an ambiguous lookup now refines `address` instead of overwriting `postcode` with a full address string. Confirmed live: `postcode=EX8 2AN, address=999 Nowhere Road` raises `SourceArgAmbiguousWithSuggestions` against `address`.

`tests/test_source_components.py`: 42 passed. Ruff clean, no generated files in the diff, doc page updated for both config forms.

Worth noting for a follow-up (not blocking this PR): `arun_gov_uk` and `southend_gov_uk` use the same `Cloud9Client` and still pass `argument_name="postcode"`, so they carry exactly the bug this PR fixes. Happy to take a separate PR aligning those two.

Thanks!

CRZTFR:
Thanks for the review. The arun/southend follow-up is up as #7345 — same one-word fix, both sources' TEST_CASES verified live and the attribution confirmed.

One thing that turned up while doing it: `northherts_gov_uk` is a third source on the same client and looks like the same bug, but it has no single address argument — it joins four fields into one `search_query` — so the reported value matches none of its fields and there is no argument the suggestions can correctly move to. I left it alone and explained why in #7345 rather than guess at it.

PR #7293, opened 2026-09-02
fix(scenicrim_qld_gov_au): raise a proper error for unknown addresses

CRZTFR:
## Summary

An address that is not present in Scenic Rim's CSV register left `service_day`
unbound, so `fetch()` raised `UnboundLocalError` rather than telling the user
what to enter:

```python
for item in address_list:
    if self._address in item[0]:
        service_day: str = item[-2]        # only bound on a match
        ...
service_days = self.generate_dates(DAYS[service_day], ...)   # UnboundLocalError
```

## Why

I run a public bin-day lookup on this source. Over 2026-08-26 to 2026-09-01,
**55 of 57** completed Scenic Rim lookups errored, all of them this crash
surfacing as an opaque HTTP 500. Anyone whose address isn't in exactly the
register's format hits it.

## Changes

1. **Raise `SourceArgumentNotFoundWithSuggestions`** with the register contents,
   so the user gets a usable candidate list instead of a stack trace.
2. **Prefer an exact register match over a substring match, and stop at it.**
   The register contains addresses that are substrings of others (`1 SMITH ST`
   is a substring of `11 SMITH ST`), and the old loop kept the *last* match
   rather than the best one, so it could silently return another property's
   schedule.
3. **Compare on collapsed whitespace.** The register mixes single- and
   double-space runs (`1 ACACIA STREET BEAUDESERT  QLD 4285`). The existing
   `HOW_TO_GET_ARGUMENTS_DESCRIPTION` even warns that this spacing "needs to be
   preserved"; normalising removes that trap without changing which property
   matches.

## Tests

Both existing `TEST_CASES` still pass against the live CSV. Additionally
verified that the single-space forms now resolve to the same properties:

```
{"address": "77A Long Road TAMBORINE MOUNTAIN QLD 4272"}   -> 48 collections
{"address": "1 William Street BEAUDESERT QLD 4285"}        -> 49 collections
{"address": "1 William Street, Beaudesert"}  -> SourceArgumentNotFoundWithSuggestions (was UnboundLocalError)
```

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/scenicrim_qld_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/scenicrim_qld_gov_au.py
@@ -1,6 +1,7 @@
 import csv
 import logging
 from datetime import datetime, timedelta
+from difflib import get_close_matches
 
 import requests
 from dateutil.rrule import FR, MO, TH, TU, WE, WEEKLY, rrule
@@ -94,9 +95,11 @@ class Source:
         # get master schedule from website
         csv_file = s.get(API_URL)
         csv_decoded = csv_file.content.decode("utf-8")
-        address_list: list = csv.reader(csv_decoded.splitlines(), delimiter=",")
-        address_list = [
-            [element.upper() for element in address] for address in address_list
+        rows = list(csv.reader(csv_decoded.splitlines(), delimiter=","))
+        # Row 0 is the header; every later row is
+        # [Street_Address, Street, Street_Locality, Service_Day, Recycle_Week].
+        address_list: list = [
+            [element.upper() for element in row] for row in rows[1:] if row and row[0]
         ]
 
         # Extract service day and recycling code. Prefer an exact match on the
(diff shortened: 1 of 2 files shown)


CRZTFR:

Both `HOW_TO_GET_ARGUMENTS_DESCRIPTION` and `doc/source/scenicrim_qld_gov_au.md` now say case and spacing do not have to match, in place of the instruction to preserve single/double spacing. (The description also had an unclosed parenthesis around the CSV URL, fixed in passing.)

Verified live: both TEST_CASES pass, and `the old avocado farm 77a long road tamborine mountain qld 4272` — lower-cased and single-spaced — resolves to the same 50 entries.

bbr111 (approved):
Thanks — good catch, and the diagnosis is exactly right. Merging with a few tweaks I pushed to your branch.

Pulling the live register (17,010 rows) turned up two things worth handling before this lands:

1. `SourceArgumentNotFoundWithSuggestions` interpolates the suggestion list straight into the message text, so passing the full register produced a ~760 KB error string in the HA UI and logs. Swapped in `difflib.get_close_matches(..., n=5, cutoff=0.4)`, matching what `municipium_it.py` and `stadtwerke_singen_de.py` already do — the message is now ~350 bytes and the candidates are more useful than a raw dump.

2. 157 rows list `Recycle_Week` as `Contact Council` rather than `Week1 Red`/`Week2 Blue`, so `item[-1].split(" ")[1]` gives `"COUNCIL"` and `START_DATES["COUNCIL"]` raises `KeyError`. Your new guard only checks `service_day`, so those addresses still crashed. They do have a valid general waste day, so they now return general waste only with an info-level log rather than blowing up. Shout if you'd rather that were a hard error.

Also moved the header row out of the match loop (it was in `address_list` but excluded from the suggestions slice) and noted both behaviours in `doc/source/`.

Both TEST_CASES pass live at 50 entries each, and `pytest tests/test_source_components.py` is green.

PR #7294, opened 2026-09-02
fix(fairfieldcity_nsw_gov_au): resolve street against the council autocomplete

CRZTFR:
## Summary

Fairfield City Council changed its street autocomplete list from abbreviated
street types (`Dawson ST, FAIRFIELD HEIGHTS`) to spelled-out ones
(`Dawson Street, FAIRFIELD HEIGHTS`). The kerbside form only returns results
for a value taken verbatim from that list, so every previously valid
configuration silently stopped resolving.

**Both existing `TEST_CASES` fail on master:**

```
1 Dawson ST, FAIRFIELD HEIGHTS  -> SourceArgumentNotFound
12 Zadro AVE, BOSSLEY PARK      -> SourceArgumentNotFound
```

## Why

I run a public bin-day lookup on this source. Over 2026-08-26 to 2026-09-01,
**70 of 70** completed Fairfield lookups failed — a 0% success rate — because
the stored abbreviated values no longer exist in the council's list.

## Changes

- Resolve `street_and_suburb` against the council's own live list
  (`fcc-autocomplete.js`, `arrAddressStreetAndSuburb`) and POST the exact
  entry it holds.
- Compare on upper-cased, whitespace-collapsed text with street-type
  abbreviations expanded (`ST`→`STREET`, `AVE`→`AVENUE`, …), so both the old
  and the new spelling resolve. Existing configurations keep working, and so
  will values copied from the current autocomplete.
- Unknown streets raise `SourceArgumentNotFoundWithSuggestions` with the list,
  rather than a bare not-found.
- If the autocomplete is unavailable or restructured, fall back to sending the
  value as given rather than refusing outright.

## Tests

`TEST_CASES` now covers both spellings for the same two properties, so a future
switch in either direction is caught. All four pass against the live site:

```
{"street_number": "1",  "street_and_suburb": "Dawson Street, FAIRFIELD HEIGHTS"} -> 3
{"street_number": "12", "street_and_suburb": "Zadro Avenue, BOSSLEY PARK"}       -> 3
{"street_number": "1",  "street_and_suburb": "Dawson ST, FAIRFIELD HEIGHTS"}     -> 3
{"street_number": "12", "street_and_suburb": "Zadro AVE, BOSSLEY PARK"}          -> 3
```

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/fairfieldcity_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/fairfieldcity_nsw_gov_au.py
@@ -1,20 +1,34 @@
+import re
 from datetime import datetime
 
 import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
-from waste_collection_schedule.exceptions import SourceArgumentNotFound
+from waste_collection_schedule.exceptions import (
+    SourceArgumentNotFound,
+    SourceArgumentNotFoundWithSuggestions,
+)
 
 TITLE = "Fairfield City Council"
 DESCRIPTION = "Source for Fairfield City Council bin collection, NSW, Australia."
 URL = "https://kerbside.fairfieldcity.nsw.gov.au/kerbside/"
 COUNTRY = "au"
 TEST_CASES = {
-    "1 Dawson ST, FAIRFIELD HEIGHTS": {
+    "1 Dawson Street, FAIRFIELD HEIGHTS": {
+        "street_number": "1",
+        "street_and_suburb": "Dawson Street, FAIRFIELD HEIGHTS",
+    },
+    "12 Zadro Avenue, BOSSLEY PARK": {
+        "street_number": "12",
+        "street_and_suburb": "Zadro Avenue, BOSSLEY PARK",
+    },
+    # Abbreviated street types: what the council's autocomplete used to emit,
+    # and therefore what most existing configurations still contain.
+    "1 Dawson ST, FAIRFIELD HEIGHTS (abbreviated)": {
         "street_number": "1",
         "street_and_suburb": "Dawson ST, FAIRFIELD HEIGHTS",
     },
-    "12 Zadro AVE, BOSSLEY PARK": {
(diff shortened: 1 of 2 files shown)


CRZTFR:
Pushed, both points addressed. Rebased on master.

- Whitespace was not collapsed: `re.split(r"(\W+)")` kept the separators and joined them back verbatim. `_canonical()` now uses `re.findall(r"\w+")` and joins on a single space, so punctuation and repeated spaces are dropped on both sides. `Dawson  ST,  FAIRFIELD HEIGHTS` and `Dawson Street FAIRFIELD HEIGHTS` (no comma) both resolve live now. I checked the whole current list for the risk that comes with dropping punctuation: 1968 entries, 0 canonical collisions.
- Network failures bypassing the fallback: `_known_streets()` catches `requests.RequestException` around the GET and `raise_for_status()` and returns `[]`, so a timeout, refused connection or 5xx takes the same fallback path an unparseable list already did. Verified with a session that raises `ConnectionError` — the POST goes ahead with the value as given.

All four live TEST_CASES still pass.

bbr111 (approved):
Thanks for the thorough write-up — the 70/70 failure rate makes the case clearly, and resolving against the council's own list rather than guessing at the spelling is the right fix.

I verified both of Copilot's points are addressed in your latest push: `_canonical()` now drops separators via `re.findall(r"\w+", ...)` so doubled spaces and a missing comma both normalise, and `_known_streets()` catches `RequestException` around the GET *and* `raise_for_status()`, so timeouts, refused connections and 5xx all take the fallback path instead of aborting the POST. Keeping both spellings in TEST_CASES is a nice touch — it will catch a switch back in either direction.

I pushed one small follow-up: `HOW_TO_GET_ARGUMENTS_DESCRIPTION`, `PARAM_DESCRIPTIONS` and `doc/source/fairfieldcity_nsw_gov_au.md` still used `Dawson ST, FAIRFIELD HEIGHTS` as the value to copy from the autocomplete, which is the spelling the council stopped emitting. Updated those to the current spelling and noted that the abbreviated form still works.

Structural suite passes (42 passed). The provider host isn't reachable from my sandbox, so I exercised the resolution logic against a mocked list rather than re-running the live cases — I'm taking your live results at face value.

PR #7295, opened 2026-09-02
feat: add source for the community of Ismaning, Germany

Kufi089:
## Summary

Adds a new source for the waste collection schedule of the community of Ismaning, Germany (official site ismaning.de).

The source resolves the selected street via the community's ajax endpoint. For streets that the community splits into house-number ranges (type 2), it also resolves and validates the house number, raising suggestion-based errors (config-flow dropdowns) when a value is missing or invalid. It then requests the ICS calendar for the matching selection and parses the collection dates into typed entries using the canonical icon catalogue.

Both flows are covered: streets without a house number and streets with one.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] python -m pytest tests/test_source_components.py -q passes
- [x] ruff check --fix and ruff format run on changed source files
- [x] doc/source/<name>.md created for new sources

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ismaning_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ismaning_de.py
@@ -18,6 +18,8 @@ DESCRIPTION = (
 URL = "https://ismaning.de/umwelt-energie/abfall/abfallkalender/"
 COUNTRY = "de"
 
+SOURCE_CODEOWNERS = ["@Kufi089"]
+
 AJAX_URL = "https://ismaning.de/wp-admin/admin-ajax.php"
 HEADERS = {
     "user-agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) Chrome/120 Safari/537.36",
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/ismaning_de.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/ismaning_de.py
@@ -1,0 +1,154 @@
+import re
+from datetime import datetime
+from zoneinfo import ZoneInfo
+
+import requests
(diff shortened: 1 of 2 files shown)


Kufi089:
@copilot /copilot review

Kufi089 (commented):
Changed accordingly to cover year rollover

bbr111 (approved):
Thanks for this — nicely structured source. The two-step street / house-number resolution with suggestion-based exceptions is exactly the right pattern for the config flow, and the UTF-8 decode workaround plus the `_year` property (good catch on the rollover) are both well commented.

Live-tested both TEST_CASES successfully: 43 entries for "Am Englischen Garten" and 46 for "Bahnhofstraße 5". `tests/test_source_components.py` passes (35 passed).

I pushed a few small tweaks to your branch:

- Added `en` entries to `PARAM_TRANSLATIONS` / `HOW_TO_GET_ARGUMENTS_DESCRIPTION` and added a `PARAM_DESCRIPTIONS` dict (en + de). Without an `en` block, `update_docu_links.py` generates English config-flow labels that just repeat the raw argument names.
- Changed `"Rama-Dama"` from `Icons.BULKY` to `Icons.EVENT` — it's the community litter-picking day rather than a bulky-waste pickup, so the sofa icon looked off.
- Fixed a typo in the German error message ("Abfahrtskalender" → "Abfallkalender").
- Declared `_street_gebietsnummer` in `__init__` rather than creating it implicitly in `_resolve_street`.

One thing worth knowing for the future: Ismaning only publishes one year at a time — requesting `year=2027` today returns an empty street list and a WordPress error page. Your `ics_url.startswith("http")` guard handles that gracefully, so nothing to change, but if users report a gap in early January that's the cause.

Approving.

Kufi089:
Thanks my first commit to a public repo YEAH! , will have a look at the changes and see if i can it improve it in a future version

PR #7298, opened 2026-09-03
fix(melton, ballina, hornsby): pass Akamai's bot check with curl_cffi

CRZTFR:
All three councils sit behind Akamai, which scores the TLS/HTTP fingerprint as well as the headers. Each one announced a browser in the User-Agent while handshaking as Python, and was served a 403 "Access Denied" page for every request, the OpenCities warm-up included.

Switching them to curl_cffi's Chrome impersonation makes the handshake agree with the User-Agent and all three answer normally again.

- **Melton** and **Ballina** already had `use_curl_cffi` available on `OpenCitiesConfig`.
- **Hornsby** predates that shared client and calls the same two endpoints over `urllib`, so its `_http_get` moves across directly.
- Both OpenCities configs also take `search_response_format="json_then_xml"`, since content negotiation hands curl_cffi the legacy XML shape unless an explicit JSON `Accept` header wins.

### Ballina address matching

Ballina also capped its fuzzy search at a single result, and that search ranks poorly: the documented test address `1 Grant St, Ballina NSW 2478` comes back with an unrelated Cumbalum property *ahead of* the Grant Street ones, so the cap guaranteed another property's bins:

```
searchfuzzy "1 Grant St, Ballina NSW 2478"
 -> ['2/7 Hartigan St CUMBALUM', '1/49 Grant Street BALLINA', '1/33 Grant Street BALLINA', ...]
```

Now it takes the whole list and disambiguates it with the strict matching lakemac already uses, offering the candidates when that is not conclusive. Its test cases move to two addresses that exist in the council's own register (the old one does not).

### Melton test cases

All four pointed at properties the council no longer holds a waste service for; the `wasteservices` call answers `success: false` for them while neighbouring properties are fine. Replaced with live ones across four suburbs.

### Shared client error message

`OpenCitiesClient` reported the internal geolocation GUID as the value to go and check whenever a property has no waste service:

> We could not find values for the argument 'street_address' with the value '523630c4-462d-4f77-a9e9-6f6481555720'

It now reports the address the visitor actually gave, and says what the council is telling us.

### Testing

All test cases pass for all three sources against the live councils.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
@@ -114,6 +114,21 @@ class OpenCitiesConfig:
     "ambiguous" failure.
     """
 
+    strict_single_result: bool = False
+    """
+    Only meaningful with ``strict_address_matching``. By default a lone search
+    result is trusted even under strict matching, on the assumption that one
+    hit means the API was confident. A few deployments' search is loose enough
+    to answer an unrelated query with exactly one wrong hit -- lakemac returns
+    the single result "2 Lake Ridge Lane, MURRAYS BEACH" for the query
+    "2 Wallarah Rd" -- and there strict matching needs to apply to a single
+    result too, so the caller is offered it as a suggestion to confirm rather
+    than silently handed another property's bins. Left off by default because
+    for most councils the lone hit is right and merely formatted differently
+    from what the user typed (a missing state abbreviation or postcode), which
+    would otherwise turn a correct resolution into a hard failure.
+    """
+
     require_date_precise: bool = False
     """
     If True, only wasteservices blocks additionally carrying a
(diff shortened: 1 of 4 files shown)


CRZTFR:
One note on the `OpenCities.py` diff: it also carries an opt-in `strict_single_result` config flag, because Ballina's fix depends on it.

`strict_address_matching` alone returns `items[0]` whenever the search yields exactly one result, before strict matching is consulted, which disables strict matching precisely where a loose search is most dangerous. Ballina is such a search: `searchfuzzy` answers "1 Grant St, Ballina NSW 2478" with a single unrelated Cumbalum property when capped at one result. The flag is a pure addition and defaults to off, so every existing source behaves identically unless it opts in.

(The docstring cites lakemac, which is the other deployment with this behaviour and would benefit from setting the same flag. Happy to split the flag into its own PR if you would rather review it separately.)

PR #7299, opened 2026-09-03
fix(gleneira_vic_gov_au): resolve the address through the council autocomplete

CRZTFR:
The results page only answers for an address written exactly the way Glen Eira's register holds it: no commas, spelled-out street type, suburb, state and postcode (`4 Staniland Grove ELSTERNWICK VIC 3185`). Anything else gets the no-results page and a "check your address" error, including the single most common thing anyone types, the comma after the street name:

```
"4 Staniland Grove ELSTERNWICK VIC 3185"    -> 3 collections
"4 Staniland Grove, Elsternwick VIC 3185"   -> SourceArgumentNotFound
"4 Staniland Grove ELSTERNWICK"             -> SourceArgumentNotFound
"4 Staniland Gr ELSTERNWICK VIC 3185"       -> SourceArgumentNotFound
```

The council's own address box does not send what was typed either. It sends the `value` of the autocomplete entry the visitor picked, from `/api/nearby/addressautocomplete`. This does the same before asking for the collections: expand street-type abbreviations, drop the trailing state and postcode when the register's substring search does not match with them, and resolve against the candidates.

That endpoint matches on a substring, so a query for `68 Glen Eira Road` also returns `268 Glen Eira Road`. Candidates are therefore filtered to the street number actually asked for, and offered as `SourceArgAmbiguousWithSuggestions` when that is still not conclusive. When the endpoint knows of nothing at all, the address is passed through unchanged so the results page keeps its own say.

### Testing

The existing test case still passes, plus three new ones covering the shapes the page used to reject: comma separated, no suburb, and an abbreviated street type. All four pass against the live council.

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -158,3 +158,4 @@ passt
 tre
 invalide
 regio
+kommas
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/gleneira_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/gleneira_vic_gov_au.py
@@ -1,10 +1,12 @@
 import logging
+import re
 from datetime import datetime
 
 import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import (
+    SourceArgAmbiguousWithSuggestions,
     SourceArgumentNotFound,
     SourceArgumentRequired,
 )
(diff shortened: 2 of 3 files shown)


bbr111 (approved):
Nice fix, and the write-up made it easy to follow. Resolving through the council's own `/api/nearby/addressautocomplete` rather than guessing at the register's formatting is the right call, and filtering the substring hits down by house number before falling back to `SourceArgAmbiguousWithSuggestions` is a good touch.

Live-tested all four cases against the council here and they all return 3 collections.

I pushed two small tweaks on top:

- The `HOW_TO_GET_ARGUMENTS_DESCRIPTION`, `PARAM_DESCRIPTIONS` and `doc/source/gleneira_vic_gov_au.md` still told users to leave out commas and match the autocomplete entry exactly, which is the very thing this PR stops requiring. Updated all of them to describe the new behaviour instead.
- Unrelated but in the same file: `URL` had a typo (`gleneira.vic.gov.a`), so the link published to the README did not resolve. Fixed the missing `u`.

Thanks!

PR #7301, opened 2026-09-03
fix(darebin_vic_gov_au): resolve the address, and stop crashing on no match

CRZTFR:
Two problems, both hit on ordinary input.

### The register form is not what people type

The query is a prefix `LIKE` against `EZI_ADDRESS`, the Vicmap address string: upper case, no commas, street type spelled out, suburb then postcode (`266 GOWER STREET PRESTON 3072`). A comma, an added "VIC", or an abbreviated street type therefore matches nothing:

```
"266 Gower Street PRESTON 3072"        4 collections   (the existing test case)
"266 Gower Street, Preston VIC 3072"   no match
"266 Gower St Preston"                 no match
```

The address is now rewritten into the register's own form (commas dropped, street type expanded, trailing state removed) and tried as progressively shorter prefixes, so a slightly-off suburb or a missing postcode still resolves.

### No match crashed

`features[0]` on an empty result set raised `list index out of range`, which tells the visitor nothing. Now:

- more than one candidate → `SourceArgAmbiguousWithSuggestions` listing the matching addresses;
- none → `SourceArgumentNotFound` explaining the format the council holds.

The second request (by `OBJECTID`) is guarded the same way.

### Quoting

Single quotes in the address are escaped. The `where` clause is SQL-ish, so `O'Hea Street` used to end the string literal early and take the whole query with it.

### Testing

Both existing test cases still pass, plus two new ones for the shapes that used to fail. All four pass against the live council.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/darebin_vic_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/darebin_vic_gov_au.py
@@ -2,6 +2,10 @@ from datetime import timedelta
 
 import requests
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import (
+    SourceArgAmbiguousWithSuggestions,
+    SourceArgumentNotFound,
+)
 from waste_collection_schedule.service.ArcGis import (
     epoch_ms_to_date,
     get_next_n_dates,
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks — this is a good catch on both counts. Verified against the live council service: all four test cases resolve (105 entries each), and the two new ones do indeed fail on master, so the prefix rewriting is doing real work. The quote escaping fix for `O'Hea Street` is appreciated too.

I've made a couple of small additions on top:

- added the `PARAM_TRANSLATIONS` / `PARAM_DESCRIPTIONS` dicts (en) that this module was missing, so the config-flow UI actually explains the address format rather than showing a bare field name;
- updated `doc/source/darebin_vic_gov_au.md`, which still told people the argument had to match the map result exactly.

`tests/test_source_components.py` passes (35 passed) and ruff is clean. Merging.

PR #7302, opened 2026-09-03
fix(joondalup_wa_gov_au): report an unresolved address instead of crashing

CRZTFR:
The property lookup matches the street name as the council spells it:

```
GET /aapi/coj/propertylookup/Ashburton%20Drive%20HEATHRIDGE   33 properties
GET /aapi/coj/propertylookup/Ashburton%20Dr%20HEATHRIDGE       0 properties
```

When nothing matched — an abbreviated street type, a misspelt suburb, or a house number that is not on the street — `self._mapkey` was left as `None`, and the schedule request went on to ask for `bindatelookup/None`. That answers with an empty list, so `json.loads(r.content)[0]` failed as `list index out of range`.

Now:

- the street is retried with the street type written out in full (`Dr` → `Drive`);
- an unknown street raises `SourceArgumentNotFound` naming the street and suburb, and saying to write the street type out;
- an unknown house number raises `SourceArgumentNotFoundWithSuggestions` offering the numbers the council does list for that street;
- the schedule request is guarded rather than subscripted blind.

### Testing

Both existing test cases still pass, plus a new one using `Ashburton Dr`, against the live council.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/joondalup_wa_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/joondalup_wa_gov_au.py
@@ -5,6 +5,10 @@ import requests
 from dateutil.rrule import FR, MO, SA, SU, TH, TU, WE, WEEKLY, rrule
 from requests.utils import requote_uri
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import (
+    SourceArgumentNotFound,
+    SourceArgumentNotFoundWithSuggestions,
+)
 
 TITLE = "City of Joondalup"
 DESCRIPTION = "Source for City of Joondalup (WA) waste collection."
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Nice, well-scoped fix. Verified against the live council API:

- All three TEST_CASES pass (6 entries each), including the new `Ashburton Dr` case that exercises the `Dr` -&gt; `Drive` retry.
- The four failure modes now raise properly instead of dying with `list index out of range`:
  - unknown street -&gt; `SourceArgumentNotFound` naming street and suburb
  - unknown number -&gt; `SourceArgumentNotFoundWithSuggestions` offering the real house numbers for that street
  - unknown suburb -&gt; `SourceArgumentNotFound` pointing at the address
  - unknown mapkey -&gt; `SourceArgumentNotFound` on the empty schedule

`ruff check` / `ruff format` are clean and `tests/test_source_components.py` passes (35 passed). Guarding the `bindatelookup` response rather than subscripting it blind is the right call.

No changes needed from my side. Thanks!

PR #7303, opened 2026-09-03
fix(cockburn_wa_gov_au): accept a postcode with no state, and stop crashing

CRZTFR:
The property search wants the state spelled between suburb and postcode:

```
"23 Snowden Street Hammond Park  WA  6164"   1 result
"23 Snowden Street Hammond Park"             1 result
"23 Snowden Street HAMMOND PARK 6164"        0 results
```

The existing normalisation only rewrote a postcode that already had a `wa` in front of it, so that third form — a postcode with no state, which is what most people type — was sent through unchanged and matched nothing.

Nothing matched then went straight into `r.json()[0]`, so the visitor was shown `list index out of range`. The `if not data` guard immediately below it could never run, because the subscript had already raised.

Now:

- a trailing postcode with no state gets one;
- the search retries without the state and postcode at all, since it resolves happily on street number and name alone;
- a genuine miss raises `SourceArgumentNotFound` saying what to check;
- multiple matches raise `SourceArgAmbiguousWithSuggestions` with the council's own address strings.

### Testing

All existing test cases still pass, plus a new one for the postcode-without-state form, against the live council.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/cockburn_wa_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/cockburn_wa_gov_au.py
@@ -4,7 +4,10 @@ from typing import TypedDict
 
 import requests
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
-from waste_collection_schedule.exceptions import SourceArgumentNotFound
+from waste_collection_schedule.exceptions import (
+    SourceArgAmbiguousWithSuggestions,
+    SourceArgumentNotFound,
+)
 
 TITLE = "City of Cockburn"  # Title will show up in README.md and info.md
 DESCRIPTION = "Source script for cockburn.wa.gov.au"  # Describe your source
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks — good catch, and thanks for the clear write-up.

I verified the premise against the council API directly and it holds: "23+Snowden+Street+HAMMOND+PARK+6164" returns 0 results, while the same address with the state spelled in returns 1. So master really was feeding an empty list into r.json()[0], and the `if not data` guard below it could never run. Nice find.

I pushed two small tweaks:

- `fetch()` now checks `property_no` before `address`. On master the property_no branch ran second and overwrote `full_url`, so property_no won when both arguments were configured; the reordering flipped that. Restored the old precedence, which also happens to be the more precise identifier of the two.
- Trimmed the docstring and comments so they describe what the code does rather than the bug that was fixed, and pulled the suggestion list out into a local so ruff stops wrapping it awkwardly.

All five test cases pass live, and the output for the pre-existing cases is byte-identical to master. Approving.

PR #7304, opened 2026-09-03
fix(cbcity_nsw_gov_au): match the address on a normalised form

CRZTFR:
`fullAddress` is held as `102 Crinan Street, Hurlstone Park 2193`: comma before the suburb, street type spelled out, postcode but no state. Matching against that string verbatim (or as a substring of it) meant three of the most ordinary things a person types found nothing at all:

```
"102 Crinan Street Hurlstone Park 2193"       no comma            -> no match
"102 Crinan Street, Hurlstone Park NSW 2193"  added state         -> no match
"102 Crinan St Hurlstone Park"                abbreviated type    -> no match
```

The whole street list is already downloaded, so both sides can simply be compared on a normalised form: upper case, commas and unit slashes dropped, street-type abbreviations expanded, state removed.

The verbatim check is kept ahead of it, and the partial-match and suggestion behaviour is unchanged apart from running on the normalised text too.

### Testing

All nine existing test cases still pass, plus three new ones for the shapes above, against the live data file.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/cbcity_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/cbcity_nsw_gov_au.py
@@ -27,10 +27,56 @@ ICON_MAP = {
 
 CACHE_DURATION = 30 * 24 * 60 * 60  # 30 days
 
+# fullAddress is held as "102 Crinan Street, Hurlstone Park 2193": comma before
+# the suburb, street type spelled out, postcode but no state. Matching on that
+# string verbatim meant a missing comma, an added "NSW" or an abbreviated
+# street type found nothing, so both sides are compared on the form below.
+STREET_TYPES = {
+    "AV": "AVENUE",
+    "AVE": "AVENUE",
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks — this is a well-argued fix, and normalising both sides from the already-downloaded street list is the right call. Keeping the verbatim check ahead of it means the common path is unchanged.

I pushed two small tweaks:

- `_STATES` held `"NEW SOUTH WALES"`, but the filter runs word by word, so the multi-word member could never match a token — an address ending "Hurlstone Park New South Wales 2193" still found nothing. The phrase is now collapsed to `NSW` before the word filter, with a test case covering it.
- Moved the new test cases to the end of `TEST_CASES` so the Tab 1-6 group stays contiguous, and gave them consistent names.

I also looked at caching the normalised list rather than rebuilding it per fetch, but the live dataset is 131,752 rows: 0.142s to normalise versus ~25 MB resident for an index. Recomputing is the better trade on typical HA hardware, so I left your version alone.

All 13 test cases pass against live data, and the four address shapes all resolve to the same 34 entries as the canonical form.

PR #7305, opened 2026-09-03
fix(folkestone_hythe_gov_uk): report an unknown UPRN instead of crashing

CRZTFR:
The council's page answers 200 with no `#bincollections` panel at all for a UPRN it does not hold, so `findAll("div", {"id": "bincollections"})[0]` raised `list index out of range` — which tells the visitor nothing about what to correct. Both of the source's own test UPRNs currently produce it.

This uses `find()` and says what happened when the panel is absent, and skips a waste type listed with no dates beneath it rather than indexing past the end of `schedules`.

**This does not restore the lookup for those two UPRNs.** `service.folkestone-hythe.gov.uk` is served to UK addresses only, and from outside the UK the connection is dropped, so whether the page changed shape or the test UPRNs went stale could not be established here. The error now says which of those it is rather than hiding it behind an IndexError — if a UK-based maintainer can check the two test UPRNs against the council's own address search, that would settle it and I am happy to follow up with the real fix.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/folkestone_hythe_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/folkestone_hythe_gov_uk.py
@@ -4,6 +4,7 @@ from datetime import datetime
 import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Folkestone and Hythe District Councol"
 DESCRIPTION = "Source for Folkestone and Hythe District Council, United Kingdom."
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks — swapping `findAll(...)[0]` for `find()` plus an explicit guard is the right shape, and I appreciate you being upfront that this doesn't restore the lookup itself.

I made one small tweak while reviewing. Since both test UPRNs currently land in the missing-panel branch, "check it against the council's address search" is probably not the whole story — a changed page layout is at least as likely as a stale UPRN, so I broadened the message to name both and to ask for an issue if the UPRN does check out. I also trimmed the comment down to the condition it guards.

I can't verify the live behaviour either: `service.folkestone-hythe.gov.uk:443` refuses the connection outright from here (TCP-level, not a 403), which matches what you described. So the two test UPRNs remain unconfirmed and the source is still effectively broken for its own TEST_CASES — but this is a clear improvement over `IndexError: list index out of range`, and the new message will make the eventual bug report far more actionable. Merging on that basis; a follow-up from a UK vantage point would be very welcome.

PR #7308, opened 2026-09-03
Fix/issue 7306 gmina sroda slaska

### LINKED ISSUE #7306: [Source Defect]: Sroda slaska stopped working since september
### Source Name

gmina_sroda_slaska_pl

### What's happening?

New schedule by new operator has been provided. No data is being provided by integration anymore. API needs an update.

### Error Logs

```Shell
no output
```

### Source Configuration

```YAML
n/a
```

### Checklist

parencik:
## Summary

Fixes #7306

Updates the Gmina Środa Śląska source after the waste-collection provider changed.

The source now retrieves the current public schedule from COM-D, which is linked by the official Gmina Środa Śląska schedule page. The integration metadata and documentation reference the municipality page as the authoritative source.

The legacy `location_id` represented groups of locations, while COM-D publishes schedules per locality or district. Users therefore need to replace `location_id` with their COM-D URL slug, for example `location: szczepanow`.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [x] Documentation update
- [ ] Other

## Checklist

- [x] `doc/source/<name>.md` created for new sources (not applicable; existing documentation updated)

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/gmina_sroda_slaska_pl.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/gmina_sroda_slaska_pl.py
@@ -1,57 +1,90 @@
 import datetime
 
 import requests
-from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from bs4 import BeautifulSoup
+from waste_collection_schedule import Collection, Icons
 
 TITLE = "Gmina Środa Śląska"
 DESCRIPTION = "Source for Gmina Środa Śląska, Poland"
-URL = "https://waste-collection.sciana.pro"
+URL = "https://srodowisko.srodaslaska.pl/gospodarka-odpadami/harmonogram-odbioru-odpadow-komunalnych/"
+SCHEDULE_URL = "https://www.com-d.pl/komunalne/harm/sroda-slaska"
+COUNTRY = "pl"
 TEST_CASES = {
-    "Ciechów": {"location_id": 3},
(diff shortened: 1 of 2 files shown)


bbr111 (approved):
Thanks for tracking down the provider change and rewriting this — I confirmed the old sciana.pro API still answers with HTTP 200 but its data stops at 2026-08-25, so it is silently stale rather than down. That matches the report in #7306 exactly.

I pushed a couple of small tweaks on top, mostly the two points Copilot raised:

- `location_id` is now accepted as an obsolete keyword and raises `SourceArgumentRequired` with migration instructions. Every current user of this source is configured with `location_id`, so without this they would just get `TypeError: __init__() got an unexpected keyword argument` in the log with no hint about what to do. Same pattern as the deprecated `ortschaft` handling in `buchegg_so_ch.py`.
- The `next()` over the tables now has a default and raises a descriptive error, and a 404 for an unknown slug raises `SourceArgumentNotFoundWithSuggestions` populated from the live COM-D index page, so users get the list of valid slugs. The index is only fetched on the error path, so the normal fetch is unchanged.
- Minor: added `PARAM_TRANSLATIONS`, pulled the `com-d.pl` host into a `BASE_URL` constant, reused a `requests.Session` for the 8 requests, and added test cases for a district (`sroda-slaska-i-rejon`) and for `ciechow` so the old test locality stays covered.

All three test cases fetch live (60 / 57 / 60 entries) and `pytest tests/test_source_components.py` is green. The documented note that `location_id` configs cannot be migrated automatically is the right call — the old IDs were groups of localities. LGTM.

PR #7311, opened 2026-09-03
Add source: Matamata-Piako District Council (mpdc_govt_nz)

danielcompton:
Adds support for MPDC's waste management calendar: 

* https://www.mpdc.govt.nz/rubbish-and-recycling
* https://www.mpdc.govt.nz/calendar

I made this run against the v3 branch as I figured that is where all new development should go. Let me know if you want me to retarget against the current flow.


danielcompton:
The CI failure appears to be unrelated to these changes.

bbr111 (approved):
Reviewed and tested end to end — this is in good shape, and I want to record why the two things that look odd at first glance are both correct, since this targets `release/3.0.0` rather than `master`.

**`tests/fixtures/mpdc_govt_nz/`** is required, not extra. The v3 cassette rule says every pipeline source ships one recorded cassette per `TEST_CASES` entry under `tests/fixtures/<module>/`, slug-named from the case key — which is exactly what's here, and there are 251 sibling directories already. Leaving it out would have failed `test_pipeline_sources_ship_a_cassette`. They replay clean (`pytest tests/test_offline_fixtures.py -k mpdc` → 3 passed) and they're the new format that carries `body`, so they pin the request payload rather than eating into `FALLBACK_BUDGET`.

**The `tests/test_new_architecture.py` change is purely additive** (+38/−0): one new `TestMpdcGovtNzCalendarFiltering` class covering your preprocess/transform chain. Nothing is skipped, xfailed or deleted, and crucially nothing is added to `SOURCES_AWAITING_CASSETTE`, `CASES_AWAITING_CASSETTE`, `SOURCE_LOCAL_STEP_EXCEPTIONS`, `SOURCES_HAND_ROLLING_RETRIEVAL` or `FALLBACK_BUDGET` — the debt registers that would be the way to relax those gates. The source clears every gate on its own merits.

Also correct: **no `doc/source/mpdc_govt_nz.md`.** Pipeline sources have that rendered by `doc_generator.py` post-merge, so hand-writing one would have been the mistake. I rendered it locally and it comes out well.

I checked hardest on whether the generic `ics` platform could have covered this instead, since it's a JEvents feed. It can't: the category feeds mix in community events (book fairs, quiz nights, markets — all visible in the cassettes), and the `ics` source exposes no exclusion filter; a non-matching `regex` keeps the entry rather than dropping it. The `RowFilter` approach is the right one, and thanks for spelling out the reasoning in the module docstring.

Full suite passes locally: 7699 passed, 8 skipped. `pre-commit` (ruff, format, codespell, bandit, mypy, pyright) all clean. **The `validate (current)` CI failure is unrelated** — it's `test_config_flow.py::test_alternatives_group_members_render_their_widgets` hitting an HA-current/voluptuous shim issue on Python 3.14 (`install_as_voluptuous is shadowing an already-imported voluptuous`); PR #7276 fails identically and that file passes locally here. I'll re-run the job.

Two optional things, neither blocking:

1. Would you like to be listed as `SOURCE_CODEOWNERS: ClassVar[list] = ["@danielcompton"]`? It's strongly encouraged for new sources and means you get pinged and assigned on bug reports naming this source. I didn't add it unilaterally since it's an ongoing commitment — happy to push it if you want it.
2. Minor robustness note: `RowFilter` keeps only the two exact summaries, so if the council ever renames one, collections would quietly disappear (`RAISE_ON_EMPTY` catches a full wipe but not a partial one). `RowRelabel` already handles the trailing `*` and the comma variant, so I think the tradeoff is fine — just flagging it.

Targeting v3 was the right call. Approving.

danielcompton:
> Would you like to be listed as SOURCE_CODEOWNERS: ClassVar[list] = ["@danielcompton"]

Yes please. 

For raising on row emptiness, I'm happy with whatever you think is best.

Issue #6482, opened 2026-05-30
[Source Defect]: New Zealand Invercargill


vkarushev:
### Source Name

wastenet_org_nz

### What's happening?

<img width="1080" height="2340" alt="Image" src="https://github.com/user-attachments/assets/cad3ee95-3925-4154-ba38-54fe3ee12e30" />

Tried to set it up for the first time, so I don't know how long it hasn't been working.

### Error Logs

```Shell
I am a newbie, so sorry for not providing any
```

### Source Configuration

```YAML
I just got the integration from HACS directly. Can't withdraw that using a phone, need to tinker around more
```

### Checklist

- [x] I have tested with the latest version (master branch) via HACS → 3-dot menu → Redownload → select `master`
- [x] The service provider's website is still working
- [x] I have tested with the example parameters from the documentation
- [x] I have searched past issues (open and closed) for this problem

markvp:
Thanks for the report! The WasteNet Southland website has migrated to a new platform, which broke all existing endpoints — the `/bin-day/` page is gone and so is the WordPress API underneath it.

A fix is up at #6483 that rewrites the source for the new REST API. It also handles old-format address strings like `156 Tay Street INVERCARGILL` automatically, so existing configs won't need updating. The `entry_id` parameter is no longer supported by the new platform and has been removed.

vkarushev:
Thank you for your response and fast fix! Cheers.

PR #7312, opened 2026-09-04
Add source for Invercargill City Council (NZ)

etamtlosz:
## Summary
Adds a dedicated source script for Invercargill City Council (ICC), New Zealand.
### Provider Details
- **Provider:** Invercargill City Council
- **Country:** New Zealand
- **Website:** https://www.icc.govt.nz
- **API Endpoint:** https://www.icc.govt.nz/client-api/icc/rubbish/search
### Note regarding `wastenet_org_nz`
Invercargill City Council moved their bin collection services and search tools away from the older WasteNet Southland platform onto the council's dedicated website (`icc.govt.nz`). As a result, `wastenet_org_nz` no longer provides reliable data for Invercargill addresses. This new `icc_govt_nz` source directly interfaces with the council's active system.
### Features & Testing
- Tracks both **Red Week** (rubbish) and **Yellow Week** (recycling).
- Handles edge cases where the council API drops same-day collections after 07:00 AM upon HA restart.
- Includes network request timeout (30s).
- Full `TEST_CASES` covering all collection days (Monday to Friday) included and verified against the live API.
- Added UI configuration descriptions for Home Assistant Config Flow.
## Type of change
- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other
## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/icc_govt_nz.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/icc_govt_nz.py
@@ -1,6 +1,8 @@
 import datetime
+
 import requests
 from waste_collection_schedule import Collection  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import SourceArgumentNotFound
 
 TITLE = "Invercargill City Council"
 DESCRIPTION = "Source for Invercargill City Council rubbish collection."
@@ -47,9 +49,9 @@ class Source:
         # Disguise the script and use the correct Referer URL
         headers = {
             "User-Agent": "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/123.0.0.0 Safari/537.36",
-            "Referer": "https://www.icc.govt.nz/services/rubbish-recycling/bin-collections"
+            "Referer": "https://www.icc.govt.nz/services/rubbish-recycling/bin-collections",
         }
-        
+
         # Use a GET request with 'params' instead of a POST request with 'data'
         response = requests.get(
             API_URL, params={"address": self._address}, headers=headers, timeout=30
@@ -57,7 +59,7 @@ class Source:
         response.raise_for_status()
 
         if not response.text.strip():
-            return []
+            raise SourceArgumentNotFound("address", self._address)
 
         data = response.json()
         entries = []
@@ -71,7 +73,9 @@ class Source:
                 continue
 
             try:
-                collection_date = datetime.datetime.strptime(date_str, "%Y-%m-%dT%H:%M:%S").date()
+                collection_date = datetime.datetime.strptime(
+                    date_str, "%Y-%m-%dT%H:%M:%S"
+                ).date()
             except ValueError:
                 continue
 
@@ -83,6 +87,9 @@ class Source:
                 )
             )
 
+        if not entries:
+            raise SourceArgumentNotFound("address", self._address)
+
         # ---------------------------------------------------------------------
         # Fix for Home Assistant restart on collection day:
         # The ICC API server drops today's collection from 'NextDates' after 07:00 AM.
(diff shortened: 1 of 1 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/icc_govt_nz.py
```diff
@@ -0,0 +1,110 @@
...
+        }
+        
+        # Use a GET request with 'params' instead of a POST request with 'data'
+        response = requests.get(
+            API_URL, params={"address": self._address}, headers=headers, timeout=30
+        )
+        response.raise_for_status()
+
+        if not response.text.strip():
+            return []
```
5ila5: `return []` is not the best idea as it just fails silently.
Raising an exception is prefered here as it shows up in the HA logs.

etamtlosz: Updated to raise SourceArgumentNotFound and formatted with ruff

bbr111 (changes requested):
Thanks for the detailed write-up and for the test cases — the API research here is solid, and you clearly hit a real problem. Unfortunately I don't think a new source module is the right shape for the fix.

**This is the same provider as the existing `wastenet_org_nz` source.** That module is titled "Gore, Invercargill &amp; Southland" and already calls `https://www.wastenet.org.nz/client-api/icc/rubbish/search` — the same `client-api/icc/rubbish/*` API you're calling, just under a different hostname. Both hostnames are the same Datascape platform:

- `www.icc.govt.nz` → `icc.lb-dds-mc2.datascape.cloud` → 13.237.238.81, 52.63.157.74
- `www.wastenet.org.nz` → `wastenet.lb-dds-mc2.datascape.cloud` → 13.237.238.81, 52.63.157.74

`wastenet_org_nz` was rewritten for this exact new API in #6483 (fixing #6482, "New Zealand Invercargill"). So rather than adding a second module, could you fold your findings into the existing one? Either:

1. repoint `ADDRESS_LIST_URL` / `SEARCH_URL` at `www.icc.govt.nz` if WasteNet is genuinely being retired, or
2. add a host/council selector if Gore &amp; Southland need to stay on the wastenet.org.nz hostname.

Either way existing users keep their config working, which a brand-new source ID would not.

A few other things worth carrying over rather than dropping, whichever route you take:

- **Bin names.** The existing source maps `Red Week` → `General Waste` and `Yellow Week` → `General Waste` + `Recycling` for ICC (council id 1), because Invercargill rubbish goes out *every* week and recycling only on yellow weeks. Emitting the raw `"Red Week"` / `"Yellow Week"` labels as the collection type loses the weekly rubbish collection and breaks existing users' `customize:` / icon settings.
- **The "restart on collection day" block.** Reconstructing today's collection by assuming exactly 7-day spacing and a strict alternating cycle will invent a collection that never happened whenever a public holiday shifts the run. We'd rather show only what the API actually returns — could you drop that block?
- **Address validation.** `raise_for_status()` doesn't fire on HTTP 202, and this endpoint currently answers 202 with an empty body / an AWS WAF challenge page to non-browser clients (I reproduced this on both hostnames just now). That makes `if not response.text.strip(): raise SourceArgumentNotFound(...)` report a WAF block as a bad address. The existing source's `_resolve_address()` against `/client-api/icc/rubbish/address-list` with `SourceArgumentNotFoundWithSuggestions` handles this much better. If the WAF turns out to block server-side clients persistently, we'd need `curl_cffi` (`requests.Session(impersonate="chrome")`) — see `east_renfrewshire_gov_uk.py`.
- Small stuff for whatever lands: `doc/source/&lt;id&gt;.md` is still required if a new module survives, and `SOURCE_CODEOWNERS = ["@etamtlosz"]` would be very welcome so you get pinged on future Invercargill reports.

Happy to help review a follow-up that updates `wastenet_org_nz` instead — that's likely a much smaller diff than what you already have here.

etamtlosz:
Thanks @bbr111 for the thorough review and the helpful context!

The main reason I originally built a dedicated ICC source was that the WasteNet search had been broken for quite some time, and Invercargill City Council had launched the bin finder directly on their own website. At the time, it looked as though WasteNet might be deprecated in favor of the council's direct tool.

Now that WasteNet is functioning again and knowing they share the exact same Datascape backend, I completely agree that updating `wastenet_org_nz` is the cleaner and better path forward.

Regarding your other points:
- **Bin mapping:** Totally agree. Emitting `General Waste` weekly and adding `Recycling` on yellow weeks for ICC makes complete sense to preserve existing user configs and dashboards.
- **Collection day reconstruction:** Understood — dropped that block so we only report ground truth from the API and avoid inventing inaccurate dates when public holidays shift the schedule.
- **Address lookup & WAF:** Kept `wastenet_org_nz`'s `_resolve_address()` with suggestions.
- **Test cases & Codeowners:** Ported over the ICC test cases (covering Mon–Fri) and added myself to `SOURCE_CODEOWNERS`.

I've opened the follow-up PR here: #7338.

Closing this PR in favor of the new one.

PR #7324, opened 2026-09-07
feat(shoalhaven_nsw_gov_au): accept a street address, not only a Geolocation ID

CRZTFR:
`shoalhaven_nsw_gov_au` only takes `geolocation_id`, a GUID that has to be copied out of the council's internal `wasteservices` request in the browser's network tab. A Shoalhaven resident reported it after doing exactly that: the address was right, and the developer console was the only way through.

The council runs the same OpenCities/MyArea widget as the other sources on `service/OpenCities.py`, so the `/api/v1/myarea/search` endpoint the shared client already resolves addresses with is available here too. This is a config change rather than new scraping.

The search is forgiving about shape — `2 Cherry Plum Way`, `2 cherry plum way worrigee`, `2 Cherry Plum Wy Worrigee` and the same address with a state and postcode appended all resolve to the one property.

Two things were in the way:

- The **search** endpoint sits behind Akamai, which fingerprints the TLS handshake as well as the headers and answers `403 Access Denied` to plain requests/urllib3 whatever User-Agent it announces. `wasteservices` itself is unprotected, which is why the id-only path never hit this. Fixed with `use_curl_cffi=True`, as melton/ballina/hornsby already do.
- Without an explicit `Accept: application/json` the search returns XML — the same quirk already noted on banyule and cambridge.

`geolocation_id` keeps working and still takes precedence when both are given, so no existing configuration changes.

### Testing

All six TEST_CASES verified live (three addresses, three existing geolocation ids):

```
Testing source shoalhaven_nsw_gov_au ...
  found 2 entries for 2 Cherry Plum Way, WORRIGEE
  found 2 entries for 10 Station Street, NOWRA
  found 2 entries for 3 The Park Drive, SANCTUARY POINT
  found 2 entries for Elizabeth Dr, VINCENTIA
  found 2 entries for The Park Dr, SANCTUARY POINT
  found 2 entries for Station St, NOWRA
```

`pytest tests/test_source_components.py` passes (35 passed); `ruff check` and `ruff format --check` clean.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/OpenCities.py
@@ -56,6 +56,14 @@ class OpenCitiesConfig:
     argument_name: str = "address"
     """Exact ``Source.__init__`` kwarg name, used in raised exceptions."""
 
+    direct_argument_name: str | None = None
+    """
+    ``Source.__init__`` kwarg holding a geolocation id given instead of an
+    address, for the sources that accept one. Exceptions raised on that path
+    name this argument, so the config flow marks the field the visitor
+    actually filled in rather than the empty address one.
+    """
+
     search_fuzzy: bool = False
     """Use ``/api/v1/myarea/searchfuzzy`` instead of ``/api/v1/myarea/search``."""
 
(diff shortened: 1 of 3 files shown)


CRZTFR:

`OpenCitiesConfig` gains an optional `direct_argument_name`, used by `get_waste_services_html()` when `address is None`; Shoalhaven sets it to `geolocation_id`. So a property with no service now reports against `street_address` when an address was given and `geolocation_id` when the GUID was, instead of always marking an empty `street_address` field invalid while showing it the GUID. Default is `None`, so every other council on the client is unchanged.

Verified both paths against a stubbed `success: false` response, and all six live TEST_CASES still pass.

bbr111 (approved):
Looks good — merging as-is.

The Copilot note about argument attribution is properly addressed: `direct_argument_name` defaults to `None` and the client falls back to `argument_name`, so the other eight OpenCities councils are bit-for-bit unchanged on that path. Nice touch keeping the "no waste service for this property" error pointing at whichever field the user actually filled in.

Verified locally:
- all six TEST_CASES live (3 address, 3 geolocation id), 2 entries each
- `kiama_nsw_gov_au` regression-checked on the shared client, unchanged
- `pytest tests/test_source_components.py` → 42 passed
- full `pre-commit` on the three changed files → clean (ruff, format, codespell, bandit, mypy)

The Akamai/TLS-fingerprint and `Accept: application/json` explanations in the config comments are exactly the kind of thing that saves the next person an hour. Thanks!

PR #7330, opened 2026-09-08
fix(canadabay, cumberland, innerwest): match the waste-info register, and say when it does not

CRZTFR:
`canadabay_nsw_gov_au`, `cumberland_nsw_gov_au` and `innerwest_nsw_gov_au` share one copy of the waste-info.com.au flow: match a locality by name, then a street, then `"<number> <street> <suburb>"`. Each match compared the raw strings, and each failure `return []`, so an ordinary address came back as "no collections found" with nothing to act on.

**Case and spacing.** The register holds `Berala` and `Merrylands Road`; visitors type `berala` and `merrylands  road`. Comparing case-folded and single-spaced makes those the same address, which is what they are.

**A building name between the number and the street.** Canada Bay files the Five Dock library as:

```
'4-12 five dock library Garfield Street Five Dock'
'1m/4-12 Garfield Street Five Dock'
...
```

so for `4-12 / Garfield Street / Five Dock` the number, street and suburb are all correct and the row still does not compare equal. That address is this source's own TEST_CASE and had never resolved — the silent empty list made it look like a pass. The number must still match in full, so `4-12` does not answer for `1m/4-12`.

The three now raise `SourceArgumentNotFoundWithSuggestions` with what the register does hold at the depth that failed, as `brisbane_qld_gov_au`, `redland_qld_gov_au` and `impactapps_com_au` already do on the same platform. Inner West searches three councils in turn, so its locality suggestions are the union across the ones reached rather than whichever list happened to be last.

The shared matching lives in `service/WasteInfo.py`. Sources with their own richer matching are untouched.

### Testing

All three sources' TEST_CASES verified live — Five Dock Library resolves for the first time (106 entries). Lower-cased and upper-cased inputs resolve. An unknown suburb now lists the council's localities instead of returning nothing. `pytest tests/test_source_components.py` passes; `ruff check` and `ruff format --check` clean.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/WasteInfo.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/WasteInfo.py
@@ -1,0 +1,83 @@
+"""Register matching for councils on the waste-info.com.au platform.
+
+Those councils expose the same v1 API: ``localities.json``, then
+``streets.json?locality=`` and ``properties.json?street=``, each answering with
+``{"name": ..., "id": ...}`` rows that a source has to match by name. The names
+are the council's own register wording, and matching them literally is what
+makes an ordinary address look unserviced:
+
+- case and spacing vary ("berala" and "Merrylands  Road" are what people type,
+  "Berala" and "Merrylands Road" is what the register holds);
+- a property row may carry a building name between the number and the street
+  ("4-12 five dock library Garfield Street Five Dock"), so the number, street
+  and suburb are all correct and the row still does not compare equal.
+
+Sources with their own richer matching (brisbane, impactapps, redland) are
+untouched; this is for the ones that compared raw strings.
+"""
+
+from __future__ import annotations
+
+from collections.abc import Iterable
+
+
+def norm(value: object) -> str:
+    """Register-comparable form: single-spaced and case-folded.
+
+    Only ``None`` is empty. Street numbers arrive as ``int`` as well as ``str``,
+    so ``0`` has to keep its own value rather than fall back to "".
+    """
+    if value is None:
+        return ""
+    return " ".join(str(value).split()).casefold()
+
+
+def same(a: object, b: object) -> bool:
+    """True when two register names differ only by case or spacing."""
+    return norm(a) == norm(b)
+
+
+def property_matches(
+    register_name: object, street_number: object, street_name: object, suburb: object
+) -> bool:
+    """True when a ``properties.json`` row is the address that was asked for.
+
+    Accepts the register's canonical ``"<number> <street> <suburb>"`` and the
+    same row with a building name inserted after the number. The number must
+    still match in full, so "4-12" does not answer for "1m/4-12".
+    """
+    name = norm(register_name)
+    if name == norm(f"{street_number} {street_name} {suburb}"):
+        return True
+    return name.startswith(norm(street_number) + " ") and name.endswith(
+        norm(f"{street_name} {suburb}")
+    )
+
+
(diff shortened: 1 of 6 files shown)


CRZTFR:
Pushed, all five review points addressed. Rebased on master.

- `norm(0)` returned `""`: only `None` is empty now, so `same(0, None)` is False and an integer street number (Cumberland has one as a TEST_CASE) still compares.
- Suggestion extraction on unnormalised input, in all three sources: replaced with a shared `street_number_suggestions()` in `WasteInfo.py`. It locates the street on the same case- and space-insensitive form the matcher uses, then returns the register's own wording for the part before it, taking the last occurrence so a building name repeating the street name doesn't truncate early. Checked live: `street_name="garfield  street"` with a bad number now suggests `['1-3', '4-12 five dock library', '1m/4-12', ...]` where it previously suggested nothing.
- Unit tests: `tests/test_waste_info.py`, 18 cases covering case/space normalisation, the inserted building name, rejecting `1m/4-12` for `4-12`, the zero case, and the suggestion helper's boundaries.

All nine live TEST_CASES across the three sources still pass. Ruff clean.

bbr111 (approved):
Thanks — this is a well-scoped fix, and the write-up made it easy to verify.

I confirmed the bug against the live Canada Bay register: there is no exact `"4-12 Garfield Street Five Dock"` row, only `'4-12 five dock library Garfield Street Five Dock'` alongside `'1m/4-12 ...'`, `'1t/4-12 ...'` and friends. So the source's own Five Dock Library TEST_CASE had genuinely never resolved, and the `return []` hid it. Replacing that with `SourceArgumentNotFoundWithSuggestions` is the right call and matches what brisbane/redland/impactapps already do on this platform. All nine live TEST_CASES pass here, and the `startswith(number + " ")` guard correctly keeps `1m/4-12` from answering for `4-12`.

One small thing I pushed on top: `pytest.ini` pins collection to an explicit `python_files` allowlist, so `tests/test_waste_info.py` was never actually running under the bare `pytest` that CI invokes — `pytest tests/` collected 46 tests rather than 64. Since your file is pure string logic with no network access, I added it to the allowlist so the boundaries you locked down are genuinely enforced. Bare `pytest` now passes 64.

Nice touch using the union of localities across the councils Inner West actually reached rather than whichever list happened to be last.

CRZTFR:
Thanks, and good catch on `pytest.ini` — I had missed that the allowlist meant the file never ran. Taken the same lesson to #7339 (a Sutherland Shire parity fix landed today), which adds `test_sutherlandshire_nsw_gov_au.py` to the same `python_files` line. Whichever of the two merges second will need both filenames on that line.

bbr111:
are we sure, that we need all this individual tests?

bbr111 (approved):
Following up on my own earlier push: I've reverted the pytest.ini change I made on top of this branch.

Two reasons. First, master gained test_ics_timezone.py on that same python_files line in #7173 after this branch was cut, so the two edits conflict, and resolving in favour of this branch would have quietly dropped the ICS timezone tests from collection. Reverting makes the merge clean.

Second, we're tightening up on individual test files generally, and the python_files allowlist is where that gets decided. tests/test_waste_info.py stays in the tree — it's a shared-service test rather than a per-source one, and the 1m/4-12 and norm(0) boundaries it pins down are the kind that a live TEST_CASE would pass straight through while resolving the wrong property. It just won't run under bare pytest; `pytest tests/test_waste_info.py` still works.

The fix itself is unchanged and still good. Thanks for the thorough write-up and for chasing down all five Copilot points.

PR #7331, opened 2026-09-08
fix(coventry, geelong): raise argument errors, not RuntimeError and Exception

CRZTFR:
Both sources reported an unresolvable address by raising an untyped exception (`RuntimeError`, bare `Exception`), which a consumer can only treat as a server fault. The information needed to put the message against the field was there all along.

**Coventry** additionally offered the wrong thing. The council retired the street directory this source searches — `/directory/search?directoryID=82` answers `404` — and the 404 page's own navigation carries the same `list__link` class the street results did, so the not-found path was ready to hand back candidate "streets" like:

```
['Skip to content', 'Skip to navigation', 'Menu', 'Adult social care and health', ... 'Pay for it']
```

The 404 is now detected and reported for what it is: the council has moved collection days to its My Account service, which this source does not read yet. A genuine no-match on a directory that *does* answer says so instead, without a suggestion list.

**Geelong**'s "We couldn't find a match for" means the property is not in the register, so it is a `SourceArgumentNotFound` on the address it was given, naming the council's own search as the place to check it.

### Testing

Verified live: Coventry reports the retired directory instead of raising `RuntimeError`; Geelong reports the unmatched address instead of raising `Exception`, and its TEST_CASES still resolve. `pytest tests/test_source_components.py` passes; `ruff check` and `ruff format --check` clean.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/coventry_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/coventry_gov_uk.py
@@ -5,6 +5,10 @@ from datetime import date, datetime
 import requests
 from bs4 import BeautifulSoup
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
+from waste_collection_schedule.exceptions import (
+    SourceArgumentException,
+    SourceArgumentNotFound,
+)
 
 TITLE = "Coventry City Council"
 DESCRIPTION = "Source for waste collection services for Coventry City Council"
@@ -14,6 +18,9 @@ HEADERS = {"user-agent": "Mozilla/5.0"}
 API_URLS = {
     "search": "https://www.coventry.gov.uk/directory/search",
     "directory_record": "https://www.coventry.gov.uk",
+    # Where the council moved bin days to. This source cannot read it, so the
+    # only useful thing left to hand the user is the address of the service.
+    "my_account": "https://myaccount.coventry.gov.uk/service/find_my_bin_day",
 }
 TEST_CASES = {
     "Test_001": {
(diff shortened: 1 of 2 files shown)


bbr111:
Thanks for this — I verified the core claims and they hold up.

- Confirmed live: `https://www.coventry.gov.uk/directory/search?directoryID=82&...` now returns 404, and the 404 page's own nav does carry the `list__link` class the old code searched for — so pre-PR, a lookup really would have offered "Skip to content", "Menu", etc. as candidate streets. Good catch.
- Confirmed: collection lookups have moved to `https://myaccount.coventry.gov.uk/service/find_my_bin_day`. Would you be open to naming that URL in the error message itself? Right now it says data "moved to the council's My Account bin day service" without a link — since this source can't read that service, the URL is the only actionable thing left to hand the user.
- I couldn't reproduce the Geelong TEST_CASES passing from here — I get `403` with `cf-mitigated: challenge` (Cloudflare interactive challenge on our IP), and `curl_cffi` with `impersonate="chrome"` doesn't get past it either. Your diff doesn't touch Geelong's network path, so I don't think this blocks the PR — just flagging in case it's a wider issue for other users, or if you're seeing something different from your end.

Separately: Coventry's directory lookup is now dead for all three TEST_CASES (they can only ever reach the 404 branch). Not a problem with this PR — it's the right interim fix — but the source will need a rewrite against the My Account endpoint at some point. Happy to open a tracking issue for that if useful.

CRZTFR:
Thanks for checking it live — that all matches what I see.

- Message now names the URL: it reads "Collection days have moved to https://myaccount.coventry.gov.uk/service/find_my_bin_day, which this source does not read yet." `my_account` sits alongside the other two in `API_URLS` so it stays with them if it moves again. Confirmed 200 just now.
- Geelong: all four TEST_CASES resolve from here (12 entries each), no Cloudflare challenge. So it looks IP- or region-specific rather than a change in the source — nothing in this diff touches its network path either way.
- Coventry rewrite against My Account: yes please, a tracking issue would be useful. I'm happy to take it. The service is a Firmstep/Achieve form rather than a plain GET, so it needs its own lookup rather than a URL swap, which is why this PR stops at reporting the retirement.

bbr111 (approved):
Diff and CI both confirm the fix — thanks for the follow-up on the My Account URL and for taking the Coventry rewrite tracking issue. Approving.

PR #7337, opened 2026-09-10
Fix PUP Saubermacher schedule endpoint

MaticKonecnik:
## Summary

The previous PUP Saubermacher endpoint /php/dobi_tabelo.php?q=<place_id> now returns HTTP 500.
The schedule is now available through /index.php/domov/urnik-odvoza-odpadkov?q=<place_id>.
This change updates the endpoint and adapts the parser to the full HTML schedule page.
Tested against the current PUP schedule page for place ID 412177, successfully parsing all 3 waste types and their collection dates.

## Type of change

- [ ] New source
- [x ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x ] `python -m pytest tests/test_source_components.py -q` passes

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/pup_si.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/pup_si.py
@@ -33,10 +33,7 @@ PARAM_TRANSLATIONS = {
     }
 }
 
-BASE_URL = (
-    "https://www.pup-saubermacher.si/"
-    "index.php/domov/urnik-odvoza-odpadkov"
-)
+BASE_URL = "https://www.pup-saubermacher.si/index.php/domov/urnik-odvoza-odpadkov"
 
 
 class Source:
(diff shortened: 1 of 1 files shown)


bbr111 (approved):
Thanks for tracking this down — confirmed the old /php/dobi_tabelo.php endpoint is dead and the new one parses cleanly. Live test gives 23 entries for Sostanj 1 and 11 for Sostanj 2, all three waste streams present.

I also checked the failure path: an invalid place_id returns HTTP 200 with a full page but no matching headings, so your `if not data:` guard correctly raises SourceArgumentNotFound. Good replacement for the old `response.text == "null"` check.

I pushed a couple of small tweaks on top:
- The three heading strings were duplicated between `parse_to_obj` and `get_type`, so I hoisted them into a module-level `WASTE_TYPE_PREFIXES` mapping used by both.
- `get_type` fell through to `"B"` for anything unrecognised; it now returns `None` and `parse_to_obj` uses that as the filter, so a renamed heading can't silently land in the organic bin.
- `get_date` now wraps `strptime` in a try/except instead of relying on the token count, so a bare `16.09.2026` with no trailing weekday still parses and stray text is skipped.

Re the Copilot suggestion about a mocked-response regression test: not applicable here, this repo doesn't carry per-source fixtures and test_source_components.py is metadata-only by design.

LGTM, merging.

PR #4572, opened 2025-08-10
Fix/source darebin vic gov au

jdndm:
- corrected API_URL as it seems to have changed (closes #4577)
- fixed test cases that didn't work
- added support for street sweeping (closes #3877)
- bugfix: was dropping the next collection date when getting future dates sometimes
- reduced the amount of dates returned for green and recycling to 1 years worth
- added unit tests to ensure fixes working as expected

I'm not sure if unit tests for sources is something that you are welcoming, I made sure that it wouldn't impact on how any of the current tests are already run. Happy to pull this out if you want.

I did note that this api only returns the very next collection date, but the source file generates additional dates by using a time delta. These generated dates have the potential to be wrong, as when a collection date falls on a public holiday, the collection will occur on the following date. 
I left it in for now as others might be relying on these dates (even if potentially wrong). Happy to rework the source to only provide what the api returns, not sure what the project prefers.

unit test results before fix
```
====================================================================== test session starts =======================================================================
platform linux -- Python 3.13.2, pytest-8.4.1, pluggy-1.6.0
rootdir: /home/jack/src-personal/hacs_waste_collection_schedule
configfile: pytest.ini
plugins: anyio-4.10.0
collected 12 items                                                                                                                                               

tests/test_darebin_vic_gov_au.py .FFFF.......                                                                                                              [100%]

============================================================================ FAILURES ============================================================================
____________________________________________________________________ test_start_date_is_today ____________________________________________________________________
tests/test_darebin_vic_gov_au.py:41: in test_start_date_is_today
    assert results == expected, (
E   AssertionError: Expected [datetime.date(2025, 8, 9), datetime.date(2025, 8, 16)] for start_date=2025-08-09, n=2, delta=7 days, 0:00:00, but got [datetime.date(2025, 8, 16), datetime.date(2025, 8, 23)]
E   assert [datetime.dat...(2025, 8, 23)] == [datetime.dat...(2025, 8, 16)]
E     
E     At index 0 diff: datetime.date(2025, 8, 16) != datetime.date(2025, 8, 9)
E     Use -v to get more diff
______________________________________________________________ test_start_date_after_today_no_skip _______________________________________________________________
tests/test_darebin_vic_gov_au.py:55: in test_start_date_after_today_no_skip
    assert results == expected, (
[...]

PR #7339, opened 2026-09-11
fix(sutherlandshire_nsw_gov_au): recycling and garden waste were the wrong way round

CRZTFR:
A resident in Engadine reported the source offering the green-lid garden bin for Monday 14 September 2026, when the council's calendar has that week as a yellow-lid recycling week. It is not a one-week slip, and it is not confined to their property: the source is inverted on every day of the published year, in both zones.

## Cause

The GIS layer this source reads (`WebOnline/MapServer/4`) gives a property's collection day, zone and recycling frequency, but nothing about which fortnight is which. I confirmed there is no date field anywhere on that layer, so the alternation has to be anchored in code. The anchor was:

```python
_REFERENCE_MONDAY = date(2024, 1, 1)
_ZONE_OFFSETS = {"1": 0, "2": 7}   # 0 = recycling on even fortnights
```

Both halves are one week out.

## Evidence

The council publishes a print calendar per zone. A yellow-highlighted week is the yellow-lid recycling week and a green one the green-lid garden week — from the legend on the calendar itself:

> Weeks with yellow highlight: Place out Yellow lid recycling bin and Red-lid garbage bin
> Weeks with green highlight: Place out Green lid garden waste bin and Red lid garbage bin

- Zone 1: `/__data/assets/pdf_file/0020/121934/SSC-Waste-Calendar-2026-27-Zone1-Print.pdf`
- Zone 2: `/__data/assets/pdf_file/0021/121935/SSC-Waste-Calendar-2026-27-Zone2-Print.pdf`

I extracted the highlight colour behind each date cell from both PDFs. They cover 2026-07-01 to 2027-06-30, 260 collection days each, alternate cleanly week to week with no break over Christmas, and are opposite to each other on every single day — which is the zone model the source already assumes, just assigned the wrong way round.

| Monday | council, Zone 1 | source before | source after |
| --- | --- | --- | --- |
| 2026-08-31 | Recycling | Garden Waste | Recycling |
| 2026-09-07 | Garden Waste | Recycling | Garden Waste |
| 2026-09-14 | Recycling | Garden Waste | Recycling |
| 2026-09-21 | Garden Waste | Recycling | Garden Waste |

Before: wrong on 260 of 260 days in Zone 1 and 260 of 260 in Zone 2. After: 0 of 520.

## Change

```python
_ZONE1_RECYCLING_WEEK = date(2026, 8, 31)   # a yellow week on the Zone 1 calendar
_ZONE_WEEK_OFFSETS = {"1": 0, "2": 1}       # whole weeks, Zone 2 is the other fortnight
```

Swapping the two offsets would have been enough, but `date(2024, 1, 1)` is an arbitrary Monday that cannot be checked against anything. The new anchor is a date a reviewer can look up in the council's own calendar, which is the property that was missing.

## Why nothing caught it

Both existing TEST_CASES are Zone 2 addresses, and a live test only asserts that collections come back, not which bin is on which day — so a full year of inverted entries looks like a pass.

- Added `4-20 Eton Street, SUTHERLAND` (the council's own administration building) as a Zone 1, Friday case, so the other zone and a non-Monday weekday are both exercised.
- Added `tests/test_sutherlandshire_nsw_gov_au.py`, which pins the parity for both zones against dates read off the printed calendars, checks the two bins alternate for 52 weeks, and covers the weekly-recycling properties.

## Verification

- All three live TEST_CASES fetch (104, 130 and 106 entries).
- `_generate_collections` matches the published calendar on all 260 days in each zone.
- The reported address, 11 Nelson Street Engadine (Zone 1, Monday), now returns Recycling on 2026-09-14 and Garden Waste on 2026-09-21.
- 19 unit tests pass; ruff check and format clean.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/sutherlandshire_nsw_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/sutherlandshire_nsw_gov_au.py
@@ -24,6 +24,14 @@ TEST_CASES = {
         "street": "Waratah Street",
         "house_number": "20",
     },
+    # The council's own administration building, and the only Zone 1 case
+    # here: both of the above are Zone 2, so nothing exercised the other
+    # fortnight.
+    "4-20 Eton Street, SUTHERLAND": {
+        "suburb": "SUTHERLAND",
+        "street": "Eton Street",
+        "house_number": "4-20",
+    },
 }
 
 ICON_MAP = {
@@ -77,15 +85,19 @@ _WEEKDAY_MAP = {
     "sunday": 6,
 }
 
-# Reference Monday used to calculate which fortnight is "recycling week".
-# Zone 1 uses this reference directly; Zone 2 is offset by one week.
-_REFERENCE_MONDAY = date(2024, 1, 1)
+# The GIS layer gives a property's zone but nothing about which fortnight is
+# which, so the alternation needs an anchor. This is the Monday of a week the
+# council's own Zone 1 calendar highlights yellow (yellow-lid recycling; green
+# highlight is the green-lid garden week):
+# https://www.sutherlandshire.nsw.gov.au/__data/assets/pdf_file/0020/121934/SSC-Waste-Calendar-2026-27-Zone1-Print.pdf
+_ZONE1_RECYCLING_WEEK = date(2026, 8, 31)
 
-# Known zone offsets in days from _REFERENCE_MONDAY.
-# 0 = recycling on even fortnights, 7 = recycling on odd fortnights.
-_ZONE_OFFSETS: dict[str, int] = {
+# Whole weeks added before taking the parity. Zone 2's recycling week is Zone
+# 1's garden week; the two published calendars are opposite on every one of the
+# 260 collection days they cover.
+_ZONE_WEEK_OFFSETS: dict[str, int] = {
     "1": 0,
-    "2": 7,
+    "2": 1,
 }
 
 _SCHEDULE_WEEKS = 52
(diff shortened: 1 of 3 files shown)


bbr111:
@CRZTFR please check new Master - rebase branch -

bbr111 (approved):
Thanks for this — the diagnosis is exactly right, and the write-up is the most thorough bug report I've seen on this repo.

I wrote the current source in #6825 and the `date(2024, 1, 1)` anchor was, honestly, a guess: the GIS layer carries day/zone/frequency but nothing that says which fortnight is which, and I couldn't reach the council's servers from my environment to check, so I picked an arbitrary Monday and the two zone offsets landed the wrong way round. Your arithmetic confirms it's a clean one-week inversion — 2026-08-31 minus 2024-01-01 is 139 weeks, odd, so every day in both zones flips. That matches your 260/260 finding rather than a partial slip, which is the right signature for this kind of error.

Anchoring on a date a reviewer can look up in the published calendar, instead of an arbitrary Monday, is the real improvement here — that was the missing property, as you say. Good catch that both existing TEST_CASES were Zone 2, and that a live test only proves collections come back, not which bin is on which day; adding a Zone 1 Friday address closes both gaps.

Checks on my side: `pytest` passes, ruff check and format are clean, and the full pre-commit run (codespell, bandit, mypy) passes. I also ran your test file forced first in collection order to make sure the `sys.path` import doesn't leak into `test_source_components.py` — it doesn't. No generated files in the diff, `doc/source/` entry already exists, and `ICON_MAP` already uses the `Icons` enum.

Two notes, neither blocking:

- I still can't verify the anchor against the council myself — their site sits behind Cloudflare (403 challenge even with `curl_cffi` impersonation) and `geoserver.ssc.nsw.gov.au` refuses connections from my network, so all three TEST_CASES fail locally for me with connection errors. I'm taking your PDF highlight extraction on trust; the point is that the new anchor is checkable by a human against a published calendar, which the old one never was.
- I pushed a small tweak to `pytest.ini`: I'm keeping your new test file (it's a genuinely useful regression check, and it's the only thing that would catch a re-inversion) but not adding it to the CI-collected `python_files` list. We deliberately keep that list narrow (#4572) and there are ~19 other per-source test files that are runnable manually but not auto-collected — this keeps things consistent. Nothing else about your PR needed to change.

Merging.

PR #7340, opened 2026-09-11
Preserve provider display colors on waste collections

PrzemyslawKlys:
A provider can use different display colors from the canonical waste-type palette. Preserve those colors on individual collections while keeping the canonical type and global defaults unchanged.

Sources can pass `color` to `Collection` or configure a transformer with `color_key` (a response field or callable). Ecoharmonogram retains the color supplied with each schedule description. Missing or invalid source colors fall back to the canonical default; an explicit YAML `customize.color` takes precedence.

Generic details include `type_id`, `color`, and `color_source`. Grouped dates retain each collection's metadata in `collections`, alongside the existing `types` list, so mixed pickups do not inherit one arbitrary color. Colors describe provider presentation and do not assert a physical bin color.

Targets the v3 release branch because it owns the canonical waste-type model. No global palette, type mapping, GUI picker, or generated translation changes are included.

Validation: 7,687 offline tests passed, 8 skipped, including all 14 recorded Ecoharmonogram replays. The config-flow test now imports Home Assistant before its validation library so current HA and the minimum supported version use consistent validator types. All pre-commit checks pass. Live Ecoharmonogram cases return schedules except an existing Ukrainian-language fixture whose requested side label is no longer returned by the provider.


bbr111 (approved):
Nice, well-scoped change — the precedence model (customize > source > canonical default) is clear, the normalisation is strict about what it accepts, and I like that `color_source` makes the provenance explicit rather than leaving users guessing. Keeping the canonical `WasteType` untouched and carrying the colour per-collection is the right call.

A few small things worth addressing before merge:

- `colors.py` is a new core-library module but isn't in `pyrightconfig.json`'s include list or the pyright pre-commit hook's `files:` filter, so it's currently analysed only as an untyped dependency and pyright errors in it would never be reported. Worth adding it to both.
- `_HEX_COLOR`'s trailing `\Z` anchor is redundant since it's only ever used with `fullmatch()`.
- The `import calendar  # noqa: F401` comment in the new test module could be completed to match the explanatory form the other nine test files use for that stdlib-shadowing guard.

I verified all three locally but wasn't able to push a branch with the fixes to this PR, so leaving them as suggestions rather than a pushed commit.

The Copilot note about import order in `test_yaml_color_validation_...` is already addressed by 19cfcfc8 — confirmed the test passes in isolation.

Two things to keep in mind rather than change here:

- `Collection.as_dict()` now carries three extra keys for every collection, and `CollectionGroup.as_dict()` embeds a full nested copy of each member under `collections`. With `details_format: generic` that list goes straight into `extra_state_attributes`, so attribute payloads grow for users with long lookaheads. Additive and fine for a major release, just worth watching against recorder column limits.
- `color` is YAML-only for now (`init_ui.py` doesn't pass it), which matches the PR description and the docs — flagging only so the asymmetry is a deliberate, recorded decision.

Verified locally: full offline suite 7687 passed / 8 skipped, `test_source_components.py` 35 passed, pyright clean, ruff clean.

cc @markvp — flagging this in case you have thoughts on the color-display feature.

PrzemyslawKlys:
Thanks @bbr111 — addressed the three actionable follow-ups in 89f67994: added colors.py to the pyright include and pre-commit trigger, removed the redundant regex anchor, and expanded the calendar shadowing comment. The full pyright run, applicable pre-commit hooks, and all 14 color tests pass locally.

bbr111:
@PrzemyslawKlys thanks for the quick response. Waiting for feedback from @markvp. Could take some time. Currently very busy.

bbr111:
will merge this for upcoming new 3.0.0 release.

PR #7345, opened 2026-09-12
fix(arun, southend): attach the address suggestions to the address argument

CRZTFR:
Follow-up to #7292, which @bbr111 asked for on that review:

> Worth noting for a follow-up (not blocking this PR): `arun_gov_uk` and `southend_gov_uk` use the same `Cloud9Client` and still pass `argument_name="postcode"`, so they carry exactly the bug this PR fixes.

Both do. `Cloud9Client._select_address` raises with `value=address_string` and suggestions that are whole address strings, so attaching them to `postcode` turns that field into an address dropdown, and picking a candidate submits a full address as the postcode instead of refining `address`. One word each.

## Verification

All live TEST_CASES pass unchanged:

| source | cases | result |
| --- | --- | --- |
| `arun_gov_uk` | 4 (2 postcode+address, 2 uprn) | 3 entries each |
| `southend_gov_uk` | 3 (1 postcode+address, 2 uprn) | 4, 5, 5 entries |

And the attribution is now right:

- `arun`: `postcode=BN17 5JA, address=999 Nowhere Road` → `SourceArgAmbiguousWithSuggestions` against `address`
- `southend`: `postcode=SS3 9JD, address=999 Nowhere Road` → `SourceArgAmbiguousWithSuggestions` against `address`

Ruff check and format clean.

## northherts_gov_uk, deliberately not touched

It is the third source on this client and it looks like the same bug, but it is not the same fix. It has no single address argument — it joins `address_name_numer`, `address_street`, `street_town` and `address_postcode` into one `search_query` and passes that as `address_string`. So the value the client reports back matches none of its four fields, and there is no argument the suggestions can correctly move to: `address_postcode` is wrong for the reason above, and `address_street` would display a full address as the street's current value.

Making that one right means giving it an `address` parameter the way eastdevon, arun and southend have, which is a bigger change than this and probably wants its own PR. Flagging it rather than guessing.

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/arun_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/arun_gov_uk.py
@@ -58,5 +58,9 @@ class Source:
         return self._client.fetch_by_address(
             postcode=self._postcode,
             address_string=self._address or "",
-            argument_name="postcode",
+            # The client's suggestions are whole address strings and it reports
+            # the address it was given as the offending value, so they belong to
+            # the address argument: picking one has to refine `address`, not
+            # overwrite the postcode with a full address.
+            argument_name="address",
         )
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/southend_gov_uk.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/southend_gov_uk.py
@@ -61,5 +61,9 @@ class Source:
         return self._client.fetch_by_address(
             postcode=self._postcode,
             address_string=self._address or "",
-            argument_name="postcode",
+            # The client's suggestions are whole address strings and it reports
+            # the address it was given as the offending value, so they belong to
+            # the address argument: picking one has to refine `address`, not
+            # overwrite the postcode with a full address.
+            argument_name="address",
         )


CRZTFR:
@bbr111, hey mate, just wanted to flag something. There's been malware injected into this repo via a fake pypi/requests


requirements.txt on master points requests at a Git URL instead of PyPI. Commit ce77c095, message "allow lower versions":


-requests>=2.31.0
+requests @ git+https://github.com/pypls/requests.git
That is not psf/requests. The pypls account was created on 2026-09-05 and has one repository. Its setup.py hides code after 434 spaces following import sys, so it sits off-screen in a normal diff. Two base64/zlib layers decode to a downloader that fetches a second stage from a Telegraph page and launches it detached. Stage two persists (SysHelper under HKCU\Software\Microsoft\Windows\CurrentVersion\Run on Windows, ~/Library/LaunchAgents/com.user.syshelper.plist on macOS) and runs a clipboard loop that swaps cryptocurrency addresses. I decoded it statically and did not execute it.

The code executes during pip's build-requirement discovery, before installation completes, so a failed pip install is not evidence of safety. It also exits on Linux, which is why CI fails loudly rather than quietly succeeding, and why the exposure falls on contributors using Windows and macOS.

Normal HACS installation looks unaffected. Home Assistant resolves dependencies from manifest.json, which is untouched and never listed this. The exposure is anyone who clones and installs the root requirements file, plus any automation that does the same.

Every open PR is currently red at dependency installation for this reason, including this one. Rebasing does not help, because the merge checkout uses master as its base.

The commit is unsigned, has no associated PR, the branch is unprotected, and it is the only commit in the last 644 whose committer is the owner's account rather than GitHub's PR-merge identity. That points towards unauthorized access, but it does not establish who pushed it or how, and I am not claiming anyone's account was taken over.

What would help: someone with verified access reverting ce77c095. If you have a way to reach Steffen outside GitHub, that seems better than raising it through this repo. Until it is reverted, please do not run pip install -r requirements.txt from a clone of master.

I am reporting the dependency repository to GitHub Support as malware. I have the full decode, SHA-256 hashes for each stage, and the hardcoded wallet addresses, and can send those wherever is most useful, including privately.

bbr111:
> @bbr111, hey mate, just wanted to flag something. There's been malware injected into this repo via a fake pypi/requests
> 
> requirements.txt on master points requests at a Git URL instead of PyPI. Commit [ce77c09](https://github.com/mampfes/hacs_waste_collection_schedule/commit/ce77c095114d6aa4e434ef69f00a95bfae3ecdcb), message "allow lower versions":
> 
> -requests>=2.31.0 +requests @ git+https://github.com/pypls/requests.git That is not psf/requests. The pypls account was created on 2026-09-05 and has one repository. Its setup.py hides code after 434 spaces following import sys, so it sits off-screen in a normal diff. Two base64/zlib layers decode to a downloader that fetches a second stage from a Telegraph page and launches it detached. Stage two persists (SysHelper under HKCU\Software\Microsoft\Windows\CurrentVersion\Run on Windows, ~/Library/LaunchAgents/com.user.syshelper.plist on macOS) and runs a clipboard loop that swaps cryptocurrency addresses. I decoded it statically and did not execute it.
> 
> The code executes during pip's build-requirement discovery, before installation completes, so a failed pip install is not evidence of safety. It also exits on Linux, which is why CI fails loudly rather than quietly succeeding, and why the exposure falls on contributors using Windows and macOS.
> 
> Normal HACS installation looks unaffected. Home Assistant resolves dependencies from manifest.json, which is untouched and never listed this. The exposure is anyone who clones and installs the root requirements file, plus any automation that does the same.
> 
> Every open PR is currently red at dependency installation for this reason, including this one. Rebasing does not help, because the merge checkout uses master as its base.
> 
> The commit is unsigned, has no associated PR, the branch is unprotected, and it is the only commit in the last 644 whose committer is the owner's account rather than GitHub's PR-merge identity. That points towards unauthorized access, but it does not establish who pushed it or how, and I am not claiming anyone's account was taken over.
> 
> What would help: someone with verified access reverting [ce77c09](https://github.com/mampfes/hacs_waste_collection_schedule/commit/ce77c095114d6aa4e434ef69f00a95bfae3ecdcb). If you have a way to reach Steffen outside GitHub, that seems better than raising it through this repo. Until it is reverted, please do not run pip install -r requirements.txt from a clone of master.
> 
> I am reporting the dependency repository to GitHub Support as malware. I have the full decode, SHA-256 hashes for each stage, and the hardcoded wallet addresses, and can send those wherever is most useful, including privately.

@CRZTFR Thanks, I've reached Steffen see issue #7349 and reverted the commit.

bbr111 (approved):

PR #7346, opened 2026-09-12
fix(vincent_wa_gov_au): query the council's current Pozi dataset proxy

### LINKED ISSUE #7281: [Source Defect]: City of Vincent (vincent_wa_gov_au) — WFS query returns "No features found" for valid in-area addresses
### Source Name

vincent_wa_gov_au

### What's happening?

The source consistently fails with `SourceArgumentNotFound`, even for valid,
unambiguous, in-area addresses — including the documentation's own example address.

Confirmed this is NOT an address-spelling or geocoding issue:
- Queried the ArcGIS World geocoder (the service this source uses internally)
  directly for the documentation's example address — it returns a clean score-100
  candidate with accurate coordinates.
- Checked the same address against City of Vincent's own public "What's Here" /
  bin-collection lookup map — it resolves correctly with a real Collection Area and
  full collection-day schedule.
- Also tested against several other single, unambiguous residential addresses within
  the City of Vincent LGA — same result each time: geocoding succeeds, council's own
  map confirms the data exists, but this source's WFS query returns zero features.

So the underlying data is live and publicly queryable via the council's own tools —
this source's specific WFS endpoint/query
(`https://mapping.vincent.wa.gov.au/pozi/qgisserver`, layer `Waste_Collection`,
WFS 1.1.0) appears to be stale, misconfigured, or superseded by whatever backend the
council's current public map tool now uses. Possibly analogous to #4620 (Cardinia),
where the source's geocoding backend had been silently migrated by the council and the
integration kept querying the old one.

Separately (secondary, not the primary bug): for a multi-unit/strata address, the
council's own lookup and a plain ArcGIS geocode of the same address string resolved to
noticeably different points (~26m apart) — looks like a units-ambiguity issue at that
specific address type rather than a general precision bug, but worth knowing about
since `query_wfs_layer`'s point `Intersects` filter has no buffer/tolerance, so any
address resolving to the "wrong" unit's point would produce a false "no features found"
even on a healthy endpoint.

### Error Logs

```Shell
(Address/coordinates below replaced with the documentation's public example address for privacy — error type, structure, and endpoint details are unmodified.)

Integration: Waste Collection Schedule

fetch failed for source City of Vincent:
Traceback (most recent call last): File ".../waste_collection_schedule/source/vincent_wa_gov_au.py", line 66, in fetch props = query_wfs_layer(
WFS_BASE_URL, ... lng=location["x"], )
File ".../waste_collection_schedule/service/Pozi.py", line 143, in query_wfs_layer raise PoziWfsError( f"No features found for point ({lat}, {lng}) in layer {typename}" )
waste_collection_schedule.service.Pozi.PoziWfsError: No features found for point (-31.928405967508, 115.851307565277) in layer Waste_Collection

The above exception was the direct cause of the following exception:

Traceback (most recent call last): File ".../waste_collection_schedule/source_shell.py", line 187, in fetch entries: Iterable[Collection] = self._source.fetch() File ".../waste_collection_schedule/source/vincent_wa_gov_au.py", line 74, in fetch raise SourceArgumentNotFound("address", self._address) from e waste_collection_schedule.exceptions.SourceArgumentNotFound: We could not find values for the argument 'address' with the value '8 Kadina St, North Perth WA 6006', please check the spelling and try again.
```

### Source Configuration

```YAML
Configured via the UI config flow (not YAML). Address set at initial setup (not editable via the options flow — reconfiguring the integration is required to change it).

Integration: waste_collection_schedule
Source: City of Vincent (vincent_wa_gov_au)
Address: 8 Kadina St, North Perth WA 6006 # documentation's own example address
Sensors: FOGO, Recycling, General Waste
Options:
calendar_title: City of Vincent
separator: ", "
fetch_time: 01:00
fetch_interval_days: 1
day_switch_time: 10:00
day_offset: 0
ignore_duplicates: false
```

### Checklist

HallyAus:
Fixes #7281

## Problem
Every address failed with `SourceArgumentNotFound`, including the documented example, and all three test cases fail on `master`. @i-am-snappy's report in #7281 established that geocoding and the council's own lookup both work, so I traced the WFS endpoint itself.

`https://mapping.vincent.wa.gov.au/pozi/qgisserver` (`MAP=C:/Pozi/Waste.qgs`) still advertises the layer — GetCapabilities lists `Waste_Collection`, and `DescribeFeatureType` still lists its fields — but it no longer serves any features for it:

| Request | Result |
|---|---|
| `GetFeature` unfiltered, GML and JSON | empty `FeatureCollection` (envelope filled with `DBL_MAX` sentinels) |
| `GetFeature` with `BBOX` in the layer's native `EPSG:7850` | 0 features |
| `GetFeature` with the point `Intersects` filter (both axis orders, `EPSG:4326` and URN form, WFS 1.0.0 and 1.1.0) | 0 features |
| WMS `GetFeatureInfo` on the same project (`queryable="1"`) | empty `FeatureCollection` |

So this was not an axis-order, CRS, geometry-field or filter problem — the project simply has no rows behind it. It did work when the source was added in #5980 (April 2026), so the council has since moved the data.

The [bin day page](https://www.vincent.wa.gov.au/your-home/waste-recycling/your-bin-day.aspx) now embeds the Pozi viewer (`vincent.pozi.com`), which reads the same QGIS project through the tenant's dataset proxy. That endpoint serves the data, and **the source's existing spatial filter works against it unchanged**:

```
Intersects geom EPSG:4326 "lng lat" -> 1 feature
  Address = 8 Kadina St, North Perth
  Collection Area = Collection Area A
  General Waste Collection Day = ...11 Sep 2026 - 2 x weekly (Wednesday/friday)
  FOGO Collection Day = ...16 Sep 2026 - Weekly (Wednesday)
  Recycling Collection Day = ...16 Sep 2026 - Weekly  (Wednesday)
```

## Changes
- Point the source at the tenant dataset proxy, discovered via `/api/v1/public/maps/waste/datasets` rather than hard-coding the dataset id, which changes whenever the council republishes the project. Discovery happens outside the `except PoziWfsError` block, so a missing dataset isn't reported as a bad address.
- `Pozi.query_wfs_layer`: `map_path` becomes an optional keyword — the proxy URL already points at a project and rejects a `MAP` parameter. Vincent is its only caller; `frankston_vic_gov_au` and `bendigo_vic_gov_au` use `query_geojson_zones` and are unaffected (re-tested live below).
- Added `tests/test_vincent_wa_gov_au.py` for the weekly / fortnightly / 2×weekly parsing, including the double space the council's Recycling value contains.

The `address` argument, the parsing and the docs are unchanged, so **no config migration is needed**.

The reporter also noted that `Intersects` has no tolerance, so a strata address geocoding to a neighbouring unit's point could still miss. That's untouched here — worth a separate issue if it bites, since it needs a buffered filter or a nearest-feature fallback.

## Testing
```
test_sources.py -s vincent_wa_gov_au      ->  104 / 52 / 52 entries (all 3 cases; all 3 fail on master)
test_sources.py -s frankston_vic_gov_au   ->  9 entries x 6 cases (unchanged)
test_sources.py -s bendigo_vic_gov_au     ->  104 / 104 / 106 entries (unchanged)
pytest tests/test_vincent_wa_gov_au.py    ->  9 passed
pytest tests/                             ->  46 passed (unchanged)
```
`ruff check` and `ruff format` (v0.15.17) are clean. As with the other per-source test files, `pytest.ini`'s `python_files` means the new tests run when named explicitly.

https://claude.ai/code/session_017QPGfr3cQiCskofMNoP8hG

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/Pozi.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/Pozi.py
@@ -13,8 +13,9 @@ Typical workflows:
       2. Receive the properties dict of the matching zone feature.
 
   WFS spatial query (e.g. Vincent):
-      1. Call query_wfs_layer() with the QGIS server URL, map path, layer name,
-         and coordinates.
+      1. Call query_wfs_layer() with the WFS endpoint URL, layer name and
+         coordinates. Pass map_path as well for QGIS servers that expect a MAP
+         parameter; omit it for dataset proxies that already point at a project.
       2. Receive the properties dict of the matching feature.
 """
 
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/Pozi.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/Pozi.py
@@ -85,23 +85,24 @@ def query_geojson_zones(
 
 def query_wfs_layer(
     base_url: str,
-    map_path: str,
     typename: str,
     lat: float,
     lng: float,
     *,
+    map_path: str | None = None,
     timeout: int = 20,
 ) -> dict[str, Any]:
     """Query a Pozi QGIS WFS endpoint with a spatial intersects filter.
 
     Args:
-        base_url: Base URL of the QGIS WFS server
-            (e.g. 'https://mapping.vincent.wa.gov.au/pozi/qgisserver').
-        map_path: Server-side path to the QGIS project file
-            (e.g. 'C:/Pozi/Waste.qgs').
+        base_url: Base URL of the QGIS WFS server, or of a Pozi dataset proxy
+            that already points at a project (e.g.
+            'https://vincent.pozi.com/proxy/v1/maps/waste/datasets/<id>').
         typename: WFS typename / layer name (e.g. 'Waste_Collection').
         lat: Latitude of the point (EPSG:4326).
         lng: Longitude of the point (EPSG:4326).
+        map_path: Server-side path to the QGIS project file, for endpoints that
+            take a MAP parameter (e.g. 'C:/Pozi/Waste.qgs').
         timeout: Request timeout in seconds.
 
     Returns:
@@ -122,7 +123,6 @@ def query_wfs_layer(
     )
 
     params: dict[str, str] = {
-        "MAP": map_path,
         "TYPENAME": typename,
         "LAYERS": typename.replace("_", " "),
         "STYLES": "default",
@@ -133,6 +133,8 @@ def query_wfs_layer(
         "OUTPUTFORMAT": "application/json",
         "FILTER": wfs_filter,
     }
+    if map_path is not None:
+        params["MAP"] = map_path
 
     r = requests.get(base_url, params=params, timeout=timeout)
     r.raise_for_status()
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/vincent_wa_gov_au.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/vincent_wa_gov_au.py
@@ -1,6 +1,7 @@
 import re
 from datetime import datetime, timedelta
 
+import requests
 from waste_collection_schedule import Collection, Icons  # type: ignore[attr-defined]
 from waste_collection_schedule.exceptions import SourceArgumentNotFound
 from waste_collection_schedule.service.ArcGis import ArcGisGeocodeError, geocode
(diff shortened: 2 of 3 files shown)


bbr111:
are you sure, that we need this individual test file?

bbr111 (approved):
Thanks for the thorough investigation write-up — the endpoint-by-endpoint table made this very easy to verify.

Confirmed on my side: all three test cases return data through the dataset proxy (104 / 52 / 52), and the two other Pozi consumers are unaffected by the signature change — `frankston_vic_gov_au` (9 × 6) and `bendigo_vic_gov_au` (104 / 104 / 106) both still pass, and `vincent_wa_gov_au` is indeed the only caller of `query_wfs_layer`. Discovering the dataset id at runtime rather than pinning it is the right call given the council republishes the project.

I pushed two small tweaks:
- `_waste_dataset_url()` now raises `PoziError` rather than a bare `ValueError`, so all failure modes of this source come out as Pozi-domain exceptions.
- Updated the `Pozi.py` module docstring, which still told callers to pass a map path unconditionally.

Keeping the per-source test file — it matches the existing convention in `tests/` and the `2 x weekly` / double-space cases are worth pinning down offline.

Agreed that the `Intersects` tolerance issue deserves its own issue rather than riding along here.

PR #7347, opened 2026-09-12
fix(sunshinecoast_qld_gov_au): accept the address as the council shows it

### LINKED ISSUE #7209: [Source Defect]: Sunshine Coast QLD
### Source Name

sunshinecoast_qld_gov_au

### What's happening?

When entering address in the same format as the linked council bin schedule website the following error is returned: The source returned an empty response. Please check the provided arguments and try again.
https://www.sunshinecoast.qld.gov.au/living-and-community/waste-and-recycling/bin-collection-days

### Error Logs

```Shell
No relevant entries in logs
```

### Source Configuration

```YAML
No config. Error message prevents finalisation of setup
```

### Checklist

- [ ] I have tested with the example parameters from the documentation

HallyAus:
Fixes #7209

## Problem
@betterbits reported that setup fails with *"The source returned an empty response"* when entering the address in the same format as the council's bin collection page. The source's own test cases pass, so the source isn't broken as such — it just can't cope with what users type.

The council's API searches its **own abbreviated street names for a substring**, and the source passed the configured text through unchanged:

| Query | Result |
|---|---|
| `hospital rd` | 1 — `Hospital Rd, Nambour` |
| `hospital` | 1 — substring match anywhere |
| `hospital road` | **0** — the API stores `Hospital Rd` |
| `12 hospital rd` | **0** — a house number matches nothing |
| `hospital rd, nambour` | **0** — a suburb matches nothing |
| `main st` | **5** — Buderim, Eumundi, Montville, Palmwoods, + `Little Main St` |

Two defects follow from that:

1. **No match was silent.** `fetch()` returned `[]`, which surfaces as "empty response" with no hint that the street type or house number is the problem.
2. **An ambiguous name picked one arbitrarily.** `Main St` exists in five suburbs and the source used `r.json()[0]`, so users could get another suburb's schedule without knowing.

## Changes
- Reshape the query: strip a house number (`12`, `1/20`, `184-202`) and a trailing `, suburb`, then retry with the street type abbreviated (`road` → `rd`) and finally with the type dropped entirely.
- Add an optional `locality` argument, taken from the street name after a comma when not given explicitly.
- Report failures instead of swallowing them: `SourceArgumentNotFoundWithSuggestions` when nothing matches or when a `locality` doesn't (listing the suburbs on offer), and `SourceArgAmbiguousWithSuggestions` listing `"Street, Suburb"` when the name isn't unique.
- Document `locality` and the accepted formats; add `HOW_TO_GET_ARGUMENTS_DESCRIPTION` / `PARAM_DESCRIPTIONS` so the config flow explains this.
- Add offline tests for the query reshaping and the week 1 / week 2 start dates.

Existing configs keep working — `street_name` is unchanged and `locality` is optional.

I left the date arithmetic alone. Note the doc cites a base date of 11 Dec 2023 while `BASE_DATE` is 11 Dec 2021; those are exactly 104 weeks apart, so the week-1/week-2 parity is identical and it's harmless.

## Testing
`test_sources.py -s sunshinecoast_qld_gov_au` (live) — the last three cases are new:
```
  found 30 entries for Hospital Rd
  found 30 entries for Great Keppel Way
  found 30 entries for House number and suburb      (12 Hospital Rd, Nambour)
  found 30 entries for Street type spelled out      (hospital road)
  found 30 entries for Street in several suburbs    (Main St + locality: Eumundi)
```
Error paths, verified live:
```
Source("Main St")            -> SourceArgAmbiguousWithSuggestions: ... please specify one of:
                                ['Little Main St, Palmwoods', 'Main St, Buderim', 'Main St, Eumundi',
                                 'Main St, Montville', 'Main St, Palmwoods']
Source("Main St","Nowhere")  -> SourceArgumentNotFoundWithSuggestions: 'locality' ... one of the following:
                                ['Buderim', 'Eumundi', 'Montville', 'Palmwoods']
Source("Nonexistent Rd")     -> SourceArgumentNotFoundWithSuggestions: 'street_name' ... no suggestions
```
```
pytest tests/test_sunshinecoast_qld_gov_au.py  ->  17 passed
pytest tests/                                  ->  46 passed (unchanged)
```
`ruff check` and `ruff format` (v0.15.17) are clean. As with the repo's other per-source test files, `pytest.ini`'s `python_files` means the new tests run when named explicitly.

https://claude.ai/code/session_017QPGfr3cQiCskofMNoP8hG


bbr111 (approved):
Thanks — this is a thorough fix, and the write-up in the description made it easy to follow.

The diagnosis is right: the council's API substring-matches its own abbreviated street names, so anything a user copies off the bin-collection page (house number, spelled-out street type, trailing suburb) misses. Retrying with the abbreviation and then without the type is the right shape, and reporting a miss instead of returning `[]` fixes the actual complaint in #7209.

I especially like that the ambiguity error is round-trippable — `Main St, Eumundi` pasted straight back into `street_name` works, because `split_locality()` picks the suburb off the comma. That turns the error message into the fix.

Verified here: all five TEST_CASES return 30 entries live, the 17 offline tests pass, `tests/` is 46 passed, and ruff is clean.

One small tweak pushed: the How-to section still cited a base date of 11/12/23 while `BASE_DATE` is 2021-12-11. You called this out as harmless — and it is, the two are 104 weeks apart so the parity is identical — but quoting a date the code doesn't use sends the next reader checking the arithmetic down the wrong path, so I've made the doc match. Fixed two typos in the same sentence while I was in there.

Worth noting for the record: configs with an ambiguous street name that previously "worked" will now fail setup rather than silently using whichever suburb the API listed first. That's the correct trade — those users may well have been on the wrong schedule — and the error tells them exactly what to set.

PR #7354, opened 2026-09-12
add(maur_ch): Add waste collection for Gemeinde Maur, Zurich, Switzerland

Vummmi:
- Add new source maur_ch.py scraping the official municipality website
- Add documentation in doc/source/maur_ch.md
- Supports all waste types: Gruengut, Kehricht, Karton, Sonderabfall, Haecksel-Service, Hauptsammelstelle

Generated by Mistral Vibe.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/source/maur_ch.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/source/maur_ch.py
@@ -7,7 +7,8 @@ from waste_collection_schedule import Collection, Icons
 
 TITLE = "Gemeinde Maur"
 DESCRIPTION = "Source for waste collection in Maur, Canton of Zurich, Switzerland."
-URL = "https://www.maur.ch/themen/bauen-umwelt/abfall-recycling/termine.html"
+# URL contains German word "termine" (dates/schedule) - not a typo
+URL = "https://www.maur.ch/themen/bauen-umwelt/abfall-recycling/termine.html"  # codespell: ignore
 COUNTRY = "ch"
 
 TEST_CASES = {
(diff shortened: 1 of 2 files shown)


custom_components/waste_collection_schedule/waste_collection_schedule/source/maur_ch.py
```diff
@@ -0,0 +1,142 @@
...
+ICON_MAP = {
+    "Grüngut": Icons.ORGANIC,
+    "Kehricht": Icons.GENERAL_WASTE,
+    "Karton": Icons.PAPER,
+    "Sonderabfall": Icons.HAZARDOUS,
+    "Häcksel-Service": Icons.GARDEN,
+    "Häckseldienst": Icons.GARDEN,
+    "Hauptsammelstelle": Icons.RECYCLING,
+    "Hauptsammelstelle Werkhof Ebmatingen offener Samstag": Icons.RECYCLING,
+    "Häcksel-Service ab 19. Oktober 2026 - in Ebmatingen, Maur, Uessikon": Icons.GARDEN,
```
bbr111: sure you want to have this in code?
need to be changed every year.
try to normalize this.

bbr111:
@Vummmi there is also an codespell error - please add inline ignore

custom_components/waste_collection_schedule/waste_collection_schedule/source/maur_ch.py
Vummmi: That's a fair point. I didn't like it either.

bbr111 (approved):
Thanks for this! Nice clean scraper, and thanks for following up on the earlier
normalization feedback - dropping the date-specific ICON_MAP key was exactly
right.

I pushed a few tweaks rather than sending it back to you:

- **Pagination.** This was the one real find. The event list at `termine.html/924`
  is spread over three pages (`/eventsjsRequest/0/eventspage/N`), so the source
  was only seeing the first 16 of 44 events - about five weeks of schedule. It now
  follows the pagination until an empty page comes back (bounded by `MAX_PAGES`,
  with a `(date, type)` dedup set). A live run returns the full calendar out to
  2026-12-30.
- **`ValueError` instead of bare `Exception`** in the two parse-failure paths, per
  the Copilot comment - that matches the other Swiss HTML sources.
- **Dropped the inline `# codespell:ignore` comments.** Sorry for the runaround
  here - it turns out the repo's `.codespellignore` already covers `termine`,
  `januar`, `februar` and friends, and `April`/`Mai`/`August` etc. were never
  flagged in the first place. Verified codespell passes cleanly on both files
  without them.
- **Doc touch-ups:** added `Grüngut/Christbaum`, `Papiersammlung` and `Metall` to
  the waste-type table (the source emits them), and documented the title
  normalization.

`pytest tests/test_source_components.py` passes (42 passed) and ruff is clean.

One optional ask: would you like to be listed as the source owner? Adding
`SOURCE_CODEOWNERS = ["@Vummmi"]` means you get pinged automatically on bug
reports for Maur. Entirely up to you - happy to merge as-is either way.

PR #7356, opened 2026-09-12
fix(ics): preserve event-local collection dates (backport of #7173 to release/3.0.0)

bbr111:
Backport of #7173 to \`release/3.0.0\`.

\`release/3.0.0\` still carried the pre-fix date extraction in both \`ICS.convert()\` and \`ICS.convert_events()\`, and its newer parsers call those same methods, so the bug is present there too.

## What this changes

Both conversion paths now call \`icalevents.events(..., strict=True)\`, which retains each event's own date/time values instead of normalising every event to the calendar-wide timezone (UTC by default). A shared \`_event_start_date()\` helper reduces the occurrence to its wall-clock date without a second timezone conversion, so a feed's intended collection day no longer depends on the timezone the Home Assistant host runs in.

Deliberately *not* converting back to the original \`DTSTART\` tzinfo: on recurring events whose \`TZID\` is a Windows zone name (e.g. \`AUS Eastern Standard Time\`) with a \`VTIMEZONE\` whose DST rules differ from the recurrence expansion, that shifts occurrences across midnight — 26 of 52 Thursdays became Wednesdays in the fixture. \`ZoneInfo\` also raises on those names, which real Australian council feeds emit.

\`strict\` is available in icalevents 0.2.1, the pinned minimum, so no dependency change is required.

## Cherry-pick notes

Cherry-picked from master (\`e1908ff1\`, squash-merge of #7173). \`ICS.py\` and \`tests/test_ics_timezone.py\` applied cleanly. \`pytest.ini\` needed a manual merge: \`release/3.0.0\` has a longer \`python_files\` allowlist plus \`addopts\`/\`markers\`, so \`test_ics_timezone.py\` was appended to the existing allowlist rather than replacing the line. The repo's pre-commit \`ruff\`/pyupgrade hook additionally modernised two lines in the new test from \`datetime.timezone.utc\` to \`datetime.UTC\` (release/3.0.0 targets a newer Python baseline than master) — a semantics-preserving alias change, no behavior difference.

## Validation on this branch

- \`pytest tests/test_ics_timezone.py\` — 112 passed
- \`pytest tests/\` (full gating suite, including the offline cassette/fixture tests for ICS-backed providers) — 7783 passed, 8 skipped, 0 failed

Original PR: #7173

### DIFF
--- a/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
+++ b/custom_components/waste_collection_schedule/waste_collection_schedule/service/ICS.py
@@ -440,6 +440,19 @@ def _event_location_description(e: Any) -> tuple[str | None, str | None]:
     return loc, desc
 
 
+def _event_start_date(e: Any) -> datetime.date | None:
+    """Extract the calendar date from an event parsed with strict=True.
+
+    Keep the occurrence's wall-clock date. Converting to the original DTSTART
+    tzinfo can shift recurring Windows-TZID events across midnight when the
+    recurrence timezone and the embedded VTIMEZONE have different DST rules.
+    """
+    start = getattr(e, "start", None)
+    if isinstance(start, datetime.datetime):
+        return start.date()
+    return start if isinstance(start, datetime.date) else None
+
+
 class ICS:
     def __init__(
         self,
@@ -473,7 +486,12 @@ class ICS:
 
         # parse ics data
         events: list[Any] = icalevents.events(
-            start=start_date, end=end_date, string_content=ics_data.encode()
+            start=start_date,
+            end=end_date,
+            string_content=ics_data.encode(),
+            # Preserve each event's calendar date/time instead of normalising
+            # all events to a calendar-wide timezone (UTC by default).
+            strict=True,
         )
 
         # Inherit summary for recurrence exceptions that lack one.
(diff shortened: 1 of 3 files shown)


bbr111 (commented):
Reviewed as a backport: the criterion is parity with the merged master commit, and this hits it.

Verified locally:
- The `ICS.py` hunk is byte-identical to `e1908ff1` — `_event_start_date()`, both `strict=True` call sites, and both `dtstart` replacements match character-for-character.
- The `pytest.ini` manual merge is correct: `test_ics_timezone.py` was appended to `release/3.0.0`'s longer allowlist, with `addopts`/`norecursedirs`/`filterwarnings`/`markers` all preserved.
- The only deviation from master is the documented `datetime.timezone.utc` -> `datetime.UTC` alias change in the new test. Safe here: `ruff.toml` on this branch targets py312 and the CI matrix is 3.12/3.14.
- Independently confirmed the dependency claim by inspecting the `icalevents==0.2.1` sdist — `strict` is a real parameter of `events()` at the pinned floor, documented as 'return dates, datetimes and datetime with timezones as specified in ical'. No dependency bump needed.
- Both `icalevents.events()` call sites in the shared service are covered; no new-architecture parser on this branch calls `icalevents` directly.

Test results reproduce the PR description: `tests/test_ics_timezone.py` 112 passed, `tests/test_source_components.py` 35 passed, full gating suite 7783 passed / 8 skipped. ruff clean, mypy clean for `ICS.py`, all three upstream CI lanes green.

Out of scope, noting for a follow-up: `source/awido_de.py` calls `icalevents.events()` directly instead of going through the shared `ICS` service, so it still carries the unfixed date extraction. `master` has the identical pattern, so it is a pre-existing gap on both branches rather than a backport defect — fixing it here would break cherry-pick parity. Better as a separate issue against `master`, then backported.

No changes needed. Good to merge.

PR #7384, opened 2026-09-13
feat(avfallsapp_se): add Nodra support

segersvard:
## Summary

- adds Nodra (Norrköping) as an Avfallsapp.se provider
- sends the provider-specific `X-App-Token` header required by Nodra's Nova API
- maps the returned `Kärl 1` / `Kärl 2` labels and documents YAML configuration

## Type of change

- [x] Bug fix / source fix
- [x] Documentation update

## Checklist

- [x] `python -m pytest tests/` passes (160 passed)
- [x] `ruff check --fix` and `ruff format` run on the changed source file
- [x] No generated files in the diff
- [x] Existing source documentation updated
- [x] Live-tested locally against Nodra: returned two collection entries

No user device ID/API key is included in this PR.


bbr111 (approved):
Thanks for adding Nodra! This slots neatly into the existing multi-provider structure: the new SERVICE_PROVIDERS entry flows into EXTRA_INFO automatically, the Kärl 1/Kärl 2 icons use the Icons enum, PARAM_TRANSLATIONS/PARAM_DESCRIPTIONS are untouched, and there are no generated files in the diff. I verified the new API base URL is live (next-pickup/list returns device_not_registered rather than 404) and tests/test_source_components.py passes (42 passed), along with ruff, bandit and codespell.

I pushed two small cosmetic tweaks on top: a comment marking the X-App-Token as the public client token from the app rather than a user secret, and a line in the docs pointing out that Nodra does not support in-integration registration, so the device ID has to be copied from the mobile app.

PR #7385, opened 2026-09-14
fix(wm_com): handle explicit null values in WM API responses

zivagolee:
## Summary

WM's API sometimes returns keys that are present but explicitly `null` (e.g. `"pickupDates": null`). `dict.get(key, default)` only applies the default when the key is missing, so these came back as `None` and caused `TypeError`/`AttributeError` when iterated or chained.

Switched the lookups to `.get(key) or default` in:
- `_get_services`: `services`
- `_get_pickup_info`: `data`, `wasteStreams`, `services`, `pickupScheduleInfo`, `pickupDates`
- `_get_holiday_adjustments`: `holidayData`, `holidayHours`

No change to `TEST_CASES` or behaviour for non-null responses.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


bbr111 (approved):
Clean, well-scoped fix — thanks!

The `.get(key, default)` → `.get(key) or default` change is exactly right: the default in `dict.get` only applies when the key is *absent*, so an explicitly-`null` value sailed straight through as `None` and blew up downstream. All six call sites you touched are correct, and `_search_account` already used the `or []` form, so this brings the rest of the module in line.

Verified locally on PR HEAD: `ruff check` clean, `ruff format --check` clean, `pytest tests/test_source_components.py -q` → 42 passed, mypy clean for this file. No generated files in the diff. CI is green.

I couldn't live-run `test_sources.py -s wm_com -l` from the review sandbox — `rest-api.wm.com` returns a CloudFront "Request blocked" 403 for this host's IP at the `account/search` step. That code path is byte-identical to master, so it's an environmental edge/WAF block rather than anything to do with this change.

Two optional follow-ups, non-blocking and not observed in practice: inside `_get_pickup_info` a `null` *value* in `wasteStreams` (e.g. `{"GARBAGE": null}`) would still trip `stream_data.get("services")`, and in `_get_holiday_adjustments` a `null` element within `holidayData` would still trip `holiday.get("holidayHours")`. Same bug class, but no need to widen this PR for them.

Approving as-is.

PR #7387, opened 2026-09-14
fix(woollahra_nsw_gov_au): stop fuzzy date parsing inventing a weekly "Recycle problem waste" event

sh00t2kill:
## Summary

The source emitted a bogus `Recycle problem waste` collection every Saturday, for a service that has no scheduled collection at all.

The council's waste-services widget returns a promo tile for that service with no date in it — just prose linking to the drop-off event page:

The fix is to drop the fuzzy parsing, and look for an explicit date rather than just a day of the week.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


sh00t2kill:
Before change; 7 entries, including the bogus one;
`2026-09-19 : Recycle problem waste`

After the change
```
2026-09-14 : General Waste
2026-09-14 : Green Waste
2026-09-14 : Recycling
2026-01-05 : Summer Clean-Up
2026-06-15 : Winter Clean-Up
2026-09-07 : Spring Clean-Up
```

bbr111 (approved):
Nice, well-scoped fix. Confirmed locally:

- `pytest tests/test_source_components.py -q` → 42 passed
- `ruff check` / `ruff format` → clean, nothing to reformat
- Live run against both TEST_CASES returns 6 entries each, matching your posted output — the phantom weekly `Recycle problem waste` entry is gone and all six real services survive.

I also exercised the new `DATE_RE` path directly: `Mon 14/9/2026`, `Monday 5 January 2026` and `Next service Sun 05/01/2026` all parse correctly, while `Every Saturday` / bare `Saturday` / prose containing a weekday now correctly yield `None`. Swapping `fuzzy=True` for an explicit date match is exactly the right call here — fuzzy parsing latching onto a weekday in a dateless promo tile was the root cause.

Two non-blocking notes for later, both pre-existing rather than anything you introduced:

1. This source is an unmigrated OpenCities/MyArea deployment (`api/v1/myarea/search` + `ocapi/Public/myarea/wasteservices`). The shared `service/OpenCities.py` client already handles dateless blocks via `require_date_precise` / `exclude_types`, and migrating would drop ~200 lines here including the generic `Exception`/`ValueError` raises. Worth a separate PR if you fancy it.
2. `DATE_RE` doesn't match abbreviated months (`5 Jan 2026`). Not an issue with today's data, just something to remember if the widget's formatting ever shifts.

Approving as-is.

PR #7388, opened 2026-09-14
Adding sperrmuell erlensee de

SgtSeppel:
## Type of change

- [X] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [X] `python -m pytest tests/test_source_components.py -q` passes --> 42 passed
- [X] `ruff check --fix` and `ruff format` run on changed source files --> All checks passed, 1 file left unchanged
- [X] No generated files in diff (README.md, info.md, sources.json, translations/*.json — CI regenerates these post-merge) --> only sperrmuell_erlensee_de.py and doc/source/sperrmuell_erlensee_de.md added.
- [X] `doc/source/<name>.md` created for new sources --> yes, see above.


bbr111 (approved):
Thanks for the new source — nicely done. Scraping both the street list and the
event-type IDs from the live form rather than pinning them is exactly right, and
picking the rolling "6 Monate" timeframe over the year-pinned option was a good
call.

I pushed one small fix. The ICS summaries come through as
"Restmüll (MT) (Am Rathaus)", so `split(" (")[0]` was cutting at the first
parenthesis and collapsing "Restmüll (MT)" down to "Restmüll" — which made MT
days look like duplicate Restmüll entries and left your `ICON_MAP["Restmüll (MT)"]`
entry unreachable. It now strips the exact trailing " (<street>)" suffix instead.

That also matters for the one street in the dropdown that has parentheses in its
own name, "Oberhörr (Sandhof / Sonnenhof)" — I added it as a third test case so
that edge case stays covered.

Live tested: 41 entries for all three test cases, and Restmüll / Restmüll (MT)
now come through as distinct types. Structural suite passes (42 passed).

PR #7389, opened 2026-09-14
test(redbridge_gov_uk): cover the food waste collection

chrisns:
## Summary

Redbridge added a food waste collection to its household rounds. The calendar PDF now contains `Food` rows, and `redbridge_gov_uk` already maps them to `Icons.BIO_KITCHEN`. No test covered that, so a later change to the PDF parser could drop the type without a test failure.

This PR adds the missing cover and documents the collection type names. It changes no source code.

- `tests/test_redbridge_gov_uk.py`: unit tests built from a real household calendar, with the uprn redacted and no address in the fixture. They cover the food waste dates and icon, the `Food Waste` label variant, the other three services, a bank holiday shift to the Tuesday and `fetch()`.
- `doc/source/redbridge_gov_uk.md`: a table of the four collection types. The names are the PDF labels, so a user who wants a food waste sensor knows to use `Food`, and a user with a garden sensor knows to use `Garden` and not `Garden Waste`.

I did not add a `TEST_CASES` entry for food waste. The only property I can confirm has the service is my own home, and the checklist asks for addresses that are not my own. Tell me if you want one and I will find a public property with the service.

Run the new tests with `pytest tests/test_redbridge_gov_uk.py`. `pytest.ini` does not auto-discover them, which matches the other per-source test files.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [x] Documentation update
- [x] Other: test coverage

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (42 passed)
- [x] TEST_CASES use real, publicly accessible addresses (not my own) — unchanged by this PR


bbr111 (changes requested):
Thanks for digging into this, and for the clear write-up — the food waste rollout is real and worth capturing.

I'm going to ask for a change of approach though, because as written this PR doesn't get you the regression protection it's aiming for.

**The test file never runs.** `pytest.ini` sets `python_files = test_source_components.py test_fetch_retry.py test_ics_timezone.py`, and CI runs a bare `pytest` from the repo root (`.github/workflows/validate.yaml`). I confirmed locally that `pytest tests/ --collect-only` picks up zero tests from `tests/test_redbridge_gov_uk.py`. So the scenario in your summary — "a later change to the PDF parser could drop the type without a test failure" — would still happen exactly as described after this merges. You spotted the non-discovery yourself; I'd rather not add a file that reads like coverage but can never fail. The other per-source test files in `tests/` predate our current policy and are equally inert — we're not adding more (CLAUDE.md, "What NOT to do" #12).

**What would actually protect it: a `TEST_CASES` entry.** That's the mechanism reviewers run (`python test_sources.py -s redbridge_gov_uk -l`) and it hits the live PDF, so it catches parser regressions *and* Redbridge changing the calendar layout. I ran the three existing cases just now — none of them return `Food`, so you're right that there's a genuine gap:

```
council office recycling only              -> Recycling
refuse and recycling only                  -> Refuse, Recycling
a church vicarage, garden, recycling, ...  -> Refuse, Garden, Recycling
```

Please find a public property with a food collection — you don't need to use your own address, and you shouldn't. A library, community centre, school or church in Redbridge on a household round will do. https://www.findmyaddress.co.uk/ gives you the UPRN, and https://my.redbridge.gov.uk/RecycleRefuse/GetFile?uprn=<uprn> lets you confirm `Food` appears before committing it. One entry named something like `"food, refuse, recycling"` is all that's needed.

Two smaller notes, in case any of the test logic migrates elsewhere:
- `mock_get.assert_called_once_with(...)` pins the exact `requests.get` signature, so adding a `timeout=` or a header to the source later breaks the test for no real reason.
- The fixture hardcodes Sept/Oct 2026 dates, so it freezes in time in a way a live `TEST_CASES` run doesn't.

**The doc change I'd like to keep exactly as-is.** The collection-types table is genuinely useful — `Garden` vs `Garden Waste` is exactly what people get wrong when setting up a `types:` filter — and the four `mdi:` strings match the `Icons` enum values correctly.

If you drop `tests/test_redbridge_gov_uk.py` and add the food waste `TEST_CASES` entry, this is a straight approve.

bbr111:
Please also consider porting this change to release/3.0.0, as we’ll be switching development to the 3.x branch soon.

bbr111 (approved):
Thanks for making the change — this addresses it exactly right. The inert test file is gone, and the new `TEST_CASES` entry (`food, refuse, recycling`) actually exercises the `Food` path through the live PDF parser, which is the real regression protection this needed.

Re-verified locally: `tests/test_source_components.py` passes (42 passed), and a live run of `test_sources.py -s redbridge_gov_uk -l` against the new UPRN returns `Food` entries. The doc table addition is unchanged and still good.

Approving.

PR #7390, opened 2026-09-14
fix(config_flow): stop the options flow looping when a customization is deleted

chrisns:
## Summary

Deleting a customization in the options flow loops for ever.

`async_step_customize` increments `self._customize_select_idx` inside the branch that keeps the customization. On the delete path the index stays put, so the recursive call shows the same form again. Ticking `delete` again repeats it. The flow can never reach `async_step_sensor`, so the user has to abandon it and loses every other change made in that flow.

`async_step_sensor` already increments its own index on its delete path. This PR makes `async_step_customize` behave the same way.

```diff
                 self._options[CONF_CUSTOMIZE][
                     self._customize_select[self._customize_select_idx]
                 ] = user_input
-                self._customize_select_idx += 1
+            self._customize_select_idx += 1
             return await self.async_step_customize()
```

The deleted customization needs no further work. `async_step_init` already rebuilds the customize dict without every key in `customize_select`.

## Tests

`tests/test_options_flow_customize.py` drives the step directly. Three of its four tests fail before the fix and all four pass after it.

The tests use plain `asyncio.run`, so they add no new test dependency. They load the core library by file path, because putting the integration directory on `sys.path` would shadow the standard library `calendar` module with the integration `calendar` platform.

I found this while adding a food waste sensor to my own Redbridge installation over the options flow API.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (42 passed)
- [x] `doc/source/<name>.md` created for new sources (not applicable)
- [x] TEST_CASES use real, publicly accessible addresses (not my own) (not applicable)


bbr111 (approved):
Thanks for this — nice catch, and the diagnosis is exactly right. The delete path never advanced `_customize_select_idx`, so the recursive call re-rendered the same form and the flow could never reach `async_step_sensor`. Your one-liner restores symmetry with `async_step_sensor`, which already increments on its own delete path, and `async_step_init` does indeed rebuild the customize dict without the selected keys, so nothing further is needed. Merging the fix as-is.

One tweak: I've removed `tests/test_options_flow_customize.py`. Not a reflection on the tests — I reverted your fix locally and confirmed 3 of the 4 fail without it, and they're properly mocked with no network calls. The issue is that `pytest.ini` sets `python_files = test_source_components.py test_fetch_retry.py test_ics_timezone.py` and CI runs bare `pytest`, so the file is never collected. As you noted in the docstring yourself, it only runs if invoked explicitly — which means in practice it would give us no regression protection and would quietly rot alongside the ~20 other orphaned test files already sitting in `tests/`. We keep per-unit test files out by default for that reason.

If you think this path deserves standing CI coverage, the productive route is a separate PR adding the filename to `pytest.ini`'s `python_files` allowlist — that's a project-level call worth making on its own merits rather than inside a bug fix. Happy to look at that if you want to raise it.

PR #7395, opened 2026-09-14
test(redbridge_gov_uk): add a food waste TEST_CASE with its cassette

chrisns:
## Summary

Ports the food waste coverage from #7389 to the 3.x branch, as asked in https://github.com/mampfes/hacs_waste_collection_schedule/pull/7389#issuecomment-5665026691.

Redbridge added a food waste collection to its household rounds. `redbridge_gov_uk` already maps `FOOD` to `wt.FOOD_WASTE`, but no test case exercised it, so a parser change could drop the type without a failure.

- `source/redbridge_gov_uk.py`: one new `TEST_CASES` entry, `"food, refuse, recycling": {"uprn": 10034926463}`.
- `tests/fixtures/redbridge_gov_uk/food_refuse_recycling.json`: the cassette for it, recorded with `tests/record_fixtures.py`.

UPRN 10034926463 is Caretakers Lodge, South Park Primary School, Water Lane, Seven Kings, IG3 9HF. It is a school property on a household round, not my address. It returns general waste, recyclables and food waste with no garden waste, so it covers a service mix the other three cases do not.

`record_fixtures.py` refreshes every case, so I reverted the three re-recorded fixtures. The diff stays on the new case.

## Verification

```
pytest tests/test_offline_fixtures.py -k redbridge      4 passed
pytest tests/test_declared_waste_types.py tests/test_doc_generation.py
                                          924 passed, 4 skipped

offline replay of the new cassette:
  replayed collections: 28
  by type: {'General Waste': 13, 'Recycling': 13, 'Food Waste': 2}
  food dates: 2026-11-20, 2026-11-27
```

## What I did not port

The master PR also adds a collection-types table to `doc/source/redbridge_gov_uk.md`. I left it out here on purpose. That table lists the PDF labels, `Refuse` and `Garden`. This branch maps through `ICSTransformer` onto canonical `WasteTypes`, so the names a user filters on are `General Waste`, `Recycling`, `Garden Waste` and `Food Waste`. The table would be wrong here, and the `Garden` versus `Garden Waste` confusion it documents does not exist on this branch. Tell me if you want a table written against the canonical names instead.

## One thing worth knowing about the rollout

I probed 28 public non-commercial properties across the borough. Every one that has food starts it in November 2026, the earliest on 2 November. The food rollout is phased and has not reached these rounds yet, so the cassette holds food only at the tail of the three month window. That is the rollout, not a parser limit. It also explains why the three existing test cases return no food.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other: test coverage

## Checklist

- [x] `pytest tests/test_offline_fixtures.py -k redbridge` passes (4 passed)
- [x] `ruff check` and `ruff format` run on changed source files


bbr111 (approved):
Approving as-is; nothing for me to change.

I verified the claims in the description rather than taking them on trust. Replaying the new cassette offline gives 28 collections with `food_waste` on 2026-11-20 and 2026-11-27, exactly as stated. The fixture is structurally identical to the three existing Redbridge cassettes and, importantly, carries a `body` field, so it is request-pinned and does not add to the unpinned-body ratchet in `test_offline_fixtures.py`. Full gating suite is green (`pytest -m "not live"`: 7918 passed, 8 skipped), as is `ruff`.

Thanks for reverting the three incidentally re-recorded cassettes — the diff really is just the new case, which made this quick to review.

On the doc table: your call not to port it is the right one, and for a slightly stronger reason than the one you gave. On this branch `doc/source/&lt;id&gt;.md` for `BaseSource` sources is derived from class metadata by `doc_generator.render_source_doc`, so a hand-written table would drift out of sync anyway. Combined with the fact that users here filter on the canonical names, not the PDF labels, there is nothing to add. No table needed.

One thing I want to note for the record, not as a change request. This case does not strictly pin the FOOD mapping: `test_offline_replay` only asserts the replay yields valid Collections, and `test_declared_waste_types` asserts returned types are a subset of declared ones, so a regression that silently dropped FOOD would leave both green. I checked whether there is a way to express "this case must contain food waste" and there is not — the `expected` key exists in only 2 of 718 fixtures and is specific to cascading-select choice fixtures. So you used the only mechanism available, and you were right not to add a per-source test file for it. The case still earns its place: it is the only Redbridge cassette whose PDF has a FOOD section, so it guards the grid parser against that layout.

The rollout detail is useful context — thanks for probing broadly enough to establish that the November start is the phased rollout rather than a parser window limit. Worth remembering when these cassettes are next refreshed: once the rollout completes, a re-record will move food into the body of the window rather than the tail.

PR #7396, opened 2026-09-14
fix(montri_fr): skip address ids that expose no calendar

tdenolle:
### Problem

Configuring the `montri_fr` source fails for some addresses with:

```
The source returned an invalid response: "404 Client Error: Not Found
for url: https://frontend.montri.fr/v1/calendar/tablemonth/fixed/9/2026"
```

`/v1/address/search/bycontract/<query>` can return **several ids sharing the same `label` and `citycode`**, and only some of them are bound to a collection calendar. The source stopped at the first `citycode` match, which for these addresses is an id with no calendar:

```python
for candidate in candidates:
    if candidate.get("citycode") == city["inseeCode"]:
        match = candidate
        break
```

Example — `30 La Ville Es Nouvelle`, Iffendic (35750):

| id | `/v1/calendar/tablemonth/fixed/9/2026` |
|---|---|
| `35133_375uqb_00030` | 404 `No calendar available for address` |
| `35133_375uqb_00030_bis` | 404 |
| `35133_e453_00030` | **200** |
| `35133_e453_00030_bis` | 200 |

The two existing test cases (Dinozé, Épinal) only ever return one id, which is why this went unnoticed.

### Fix

Try each candidate in turn and keep the first one that actually returns a calendar. The probe response is reused for the first month, so the nominal case costs **no extra request**. When no candidate has a calendar, raise `SourceArgumentException` with actionable text instead of letting a bare HTTP 404 bubble up.

### Testing

```
found 38 entries for 6 Place Georges Boizot, Dinozé
found 155 entries for 2 Place Georges Clémenceau, Épinal
found 26 entries for 30 La Ville Es Nouvelle, Iffendic
```

`pytest tests/` → 160 passed. `ruff check` / `ruff format` clean. No generated files touched.


bbr111 (approved):
Thanks — nice catch, and a well-written problem description. The multi-id case is real and the probe-then-reuse approach means the common single-id path costs no extra request, which is the right trade-off.

Verified locally: all three test cases fetch live (38 / 155 / 26 entries), tests/test_source_components.py passes, ruff clean, no generated files in the diff.

I pushed one small tweak: the `match` variable was only used for the following `is None` check and never read afterwards, so I folded it into `first_table`. Purely cosmetic — no behaviour change.

bbr111 (approved):
Nice fix — thanks for the thorough write-up.

I reproduced the root cause against the live API and can confirm the diagnosis:

    35133_375uqb_00030      -> 404
    35133_375uqb_00030_bis  -> 404
    35133_e453_00030        -> 200
    35133_e453_00030_bis    -> 200

So the old break-on-first-citycode-match was indeed landing on a calendar-less id.
Reusing the probe response for the first month so the nominal case costs no extra
request is a nice touch, and raising SourceArgumentException with actionable text
is the right call here — the address *was* found, it just has no calendar, so
SourceArgumentNotFound would have been misleading.

Verified locally:
- pytest tests/test_source_components.py -> 42 passed
- test_sources.py -s montri_fr -l -> 38 / 155 / 26 entries, all three cases green
- ruff check + ruff format clean
- no generated files in the diff

Merging as-is. One optional follow-up if you're interested: montri_fr has no
SOURCE_CODEOWNERS entry yet, and you clearly know this API well now — feel free to
add ["@tdenolle"] in a future PR if you'd like to be pinged on montri_fr bug reports.

PR #7412, opened 2026-09-14
New source: Axla AffaldOnline Apps, Denmark

superrob:
Adds a source for the AffaldOnline Apps produced by Axla. These have vastly superior info compared to the existing AffaldOnline source, and have all future collections in a nice and neat JSON package.

The implementation of the newer AffaldOnline apps have been requested in #3044

I also considered adding this as an extension to the existing AffaldOnline source. However the method that this source uses is vastly different to the older source and thus could break existing installations.

## Summary

> **New source?** Development focus is currently on the upcoming v3.0.0 release — please target this PR at the `release/3.0.0` branch instead of `master`. Bug fixes for the current stable release still go against `master`.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes
- [x] `ruff check --fix` and `ruff format` run on changed source files
- [x] No generated files in diff (README.md, info.md, sources.json, translations/*.json — CI regenerates these post-merge)
- [x] `doc/source/<name>.md` created for new sources
- [x] TEST_CASES use real, publicly accessible addresses (not my own)

*Replaces pull request #7409 to target release/3.0.0 instead of master*

Issue #3753, opened 2025-02-09
[Source Request]:


jakobisbo:
### Municipality / Region

Nyborg, Denmark

### Collection Calendar Webpage

https://www.affaldonline.dk/kalender/nyborg/showToemCal.php?year=2025&forbid=551485&altid=0&type=1

### Example Address

I found my City under your examples here https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/doc/source/affaldonline_dk.md

But they only provide a PDF and HA reports that it can't use PDF.

Kind regards Jakob

### Collection Data Format

As PDF Download

Issue #3044, opened 2024-11-16
[Source Request]: AffaldsApp APP, Denmark


martinhansen1980:
### I Have A Problem With:

A specific source

### What's Your Problem

When i try to add "Denmark", "Fredericia Kommune", and put in my value according to the guide, "Fuglsang+Allé|58||||7000|Fredericia|17009|18935742|0" I get the error "Parser method for pdf not implemented".

### Logs

```Shell
No relevant logs
```

### Checklist Source Error

- [X] Use the example parameters for your source (often available in the documentation) (don't forget to restart Home Assistant after changing the configuration)
- [X] Checked that the website of your service provider is still working
- [X] Tested my attributes on the service provider website (if possible)
- [X] I have tested with the latest version of the integration (master) (for HACS in the 3 dot menu of the integration click on "Redownload" and choose master as version)

### Checklist Sensor Error

- [X] Checked in the Home Assistant Calendar tab if the event names match the types names (if types argument is used)

### Required

- [X] I have searched past (closed AND opened) issues to see if this bug has already been reported, and it hasn't been.
- [X] I understand that people give their precious time for free, and thus I've done my very best to make this problem as easy as possible to investigate.

5ila5:
The affaldonline_dk source you're using seems to not support your municipality, as they only offer PDFs we cannot parse, while other municipalities by affaldonline show the dates in the HTML on the website.

I removed the "Unsupported" regions from the Configuration GUI and README.md.

I saw they have an app, so we might be able to add this though the app (Did not check how the app communicates to an API yet) 

https://play.google.com/store/apps/details?id=dk.openexperience.genbrugsapp.openx&hl=en

dansollok:
Is this still not working ?

bbr111:
Thanks for your patience on this one.

To recap where things stand: the original bug (the confusing "Parser method for pdf not implemented" error) was already fixed back in November 2024 — Fredericia and the other affaldonline.dk municipalities that only publish PDF schedules are now excluded from the config-flow GUI and README, so you shouldn't be offered a broken option anymore.

This issue is now tracking a genuinely new source request: adding support via the "AffaldsApp" / Genbrugsapp mobile app (dk.openexperience.genbrugsapp.openx), as an alternative to the PDF-only affaldonline.dk feed for Fredericia.

[...]

Issue #2977, opened 2024-11-03
[Source Request]: 


rostank:
### Municipality / Region

Viborg, Denmark

### Collection Calendar Webpage

https://www.revas.dk/borger/din-affaldsordning/selvbetjening

### Example Address

Pilevej 7 (8850)

### Collection Data Format

As html on a webpage, Dedicated IOS/Android APP, As PDF Download

kefdk:
+1 for this

ahlmann78:
I would also appreciate it

ReneNulschDE:
Analysis:

Request 1 GET:
https://dagrenovation.viborg.dk/minside/Adresse/Typeahead?query=Pilevej%207%20(8850)&limit=30000

Result 1:
[{"ExtensionData":{},"AdresseId":25186,"Beliggenhed":"Pilevej 7 (8850)","Latitude":56.3788146714046,"Longitude":9.67240436215996}]

Pick AddressId: 25186

Request 2 GET:
https://dagrenovation.viborg.dk/minside/Tomningsinfo?AdresseId=25186&siteType=R&kom=0&lang=dk

Result = HTML page, search for 
```
                    <h2>Kommentar til t&#248;mninger:</h2>
                    <ul class="list">
                            <li>11-12-2024 10:25 - Madaffald - T&#248;mt - T&#248;mt</li>
                            <li>28-11-2024 13:40 - Metal/Glas - T&#248;mt</li>
                            <li>26-11-2024 07:31 - Papir/Pap - T&#248;mt</li>
                            <li>26-11-2024 07:30 - Papir/Pap - T&#248;mt</li>
                            <li>28-11-2024 13:39 - Plast/MDK - T&#248;mt</li>
                            <li>11-12-2024 10:25 - Restaffald - T&#248;mt - T&#248;mt - T&#248;mt</li>
                    </ul>
  
```

ahlmann78:
The results for request 2 is comments to pickups that have already been done. All it says is that the bins have been emptied (tømt in danish), but I think you're on the right track👍

ReneNulschDE:
Can you explain where we can find the future collections? My danish is underdeveloped ;-)

ahlmann78:
Under each type of container there are dates for upcoming pickups (kommende tømninger). Otherwise I can see there is a PDF file in the bottom of the page. The one without 2024 or 2025. It has upcoming pickups with dates adjusted for holidays, unlike the two others.

[...]

PR #7414, opened 2026-09-14
Source Rewrite - Affaldonline - New API flow

superrob:
Originally meant as a new source see #7412. 

However I discovered that I could cover all previously available municipalities in AffaldOnline and even some new municipalities.

Statisfies a few requests:
#3044 #3753 and #2977

The new API flow replaces the old screen scraping approach and enables new municipalities previously not covered. Also brings back multiple municipalities that had switched to unparseable pdfs.

Should work just fine for existing installations, as the old "values" variable is used, with the same format as before.

Some namings might differ from the old. Re-adding the integration can also help with adding the new typed sensors, which the source now properly support.

Tested on my own installation, which worked with no issue.

## Summary

> **New source?** Development focus is currently on the upcoming v3.0.0 release — please target this PR at the `release/3.0.0` branch instead of `master`. Bug fixes for the current stable release still go against `master`.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


superrob:
Could have sworn that i ran ruff on the code. But aparrently missed the ending newline 😅

bbr111 (approved):
Thanks for this — replacing the screen-scraping with the app API is a clear
win, and it's great to see fanoe/ffv/nyborg come back from the dead "pdf"
parser plus three genuinely new municipalities.

I've pushed a few small tweaks so CI goes green and to tidy a couple of
things I noticed while testing:

- the two pre-commit failures: the missing newline at EOF and the
  "abreviations" typo
- `SourceArgumentException(values, ...)` was passing the user's address
  string where the argument *name* goes — its signature is
  `(argument: str, message: str)` and the HA UI uses that to attach the
  error to the right field, so it's now `SourceArgumentException("values", ...)`
- the `SourceArgumentNotFoundWithSuggestions` suggestion list included
  `soroe`, which is the one entry you deliberately left without a `client`;
  it now only suggests municipalities that actually work
- camelCase locals renamed to snake_case to match the rest of the repo

I replayed all 15 test cases live and collected every fraction name the API
returns. Five weren't in `ICON_MAP` and were falling back to the general
waste icon, so I added four obvious ones: `Haveaffald` → GARDEN,
`Pap/Papir` → PAPER, `Glas og metal` → GLASS, and
`Papir/Pap og Plast/Mad- og drikkekartoner` → RECYCLING.

The fifth is "Blå" from Nyborg — a colour-only name I didn't want to guess
at. Since you're now the codeowner for this source, could you confirm what
Nyborg's blue bin actually holds and add it in a follow-up? Also `holbaek`
only returns 2 entries where the others return 13-32; probably fine for that
address, but worth a sanity check that the sample is representative.

One deliberate non-change: I left the missing `raise_for_status()` alone,
since the `{"message": ...}` handling below it gives a better error than a
raw HTTPError would.

Verified before pushing: `pytest tests/test_source_components.py -q` (42
passed), the full offline suite `pytest tests/ -m "not live"` (7916 passed),
`pre-commit run --all-files` (all green), and a live
`test_sources.py -s affaldonline_dk` with all 15 cases returning entries.

Noting for the release: this is a breaking change (the waste-type labels all
move), which is exactly why `release/3.0.0` is the right target.

PR #7415, opened 2026-09-14
Fix flaky voluptuous/probatio import-order race in test suite

GreenDavidA:
## Summary

Fixes the `validate (current)` CI flakiness identified while working on #7276 (root-caused there, split out as its own PR per @bbr111's suggestion).

HA 2026.9+ replaces `voluptuous` with its own shim (`probatio`) the moment `homeassistant` is imported:

```python
# homeassistant/__init__.py
from probatio.compat import install_as_voluptuous
install_as_voluptuous()   # must run before the first `import voluptuous`
```

`install_as_voluptuous()` unconditionally does `sys.modules["voluptuous"] = <probatio shim>`, but any module that already did `import voluptuous as vol` before that point keeps its own reference to the *real* voluptuous — the two don't update retroactively. So whether `test_config_flow.py`'s `vol.Optional` and `config_flow.py`'s `vol.Optional` end up the same class depends on which of them (or something they transitively import) triggers `homeassistant`'s import first. That's an import-order race across the whole test session, which is why it doesn't reproduce identically across PRs — a different diff collects a different set of test/source modules, shifting the ordering.

Reproduced directly:

```python
>>> import voluptuous as vol_before
>>> import homeassistant
.../homeassistant/__init__.py:10: RuntimeWarning: install_as_voluptuous is shadowing an
already-imported voluptuous; call it before anything imports voluptuous so references
resolve to probatio
  install_as_voluptuous()
>>> import voluptuous as vol_after
>>> vol_before is vol_after
False
```

## Fix

Add a root `tests/conftest.py` whose first statement is `import homeassistant`. pytest guarantees every `conftest.py` loads before any test module in its tree, so this pins `install_as_voluptuous()` ahead of anything that could grab a stale `voluptuous` reference, closing the race regardless of collection order:

```python
>>> import homeassistant
>>> import voluptuous as vol_a
>>> import voluptuous as vol_b
>>> vol_a is vol_b
True
```

`validate (minimum)` (pinned pre-`probatio` HA) is unaffected either way.

## Type of change

- [x] Other (test-harness fix)

## Checklist

- [x] Full offline suite passes: `pytest tests/` → 8000 passed, 8 skipped, with HA 2026.9.1 (`probatio`-shimmed voluptuous) installed
- [x] `pre-commit run --all-files` on the changed file passes (ruff, codespell, bandit, mypy)
- [x] N/A — not a source change, no `TEST_CASES`/`doc/source/*.md` involved


bbr111 (approved):
Nice root-cause fix — this replaces what had become a pattern of local workarounds (test_config_flow.py orders its homeassistant import ahead of voluptuous and asserts behaviourally rather than with isinstance; test_collection_colors.py does a defensive in-function `import homeassistant`), and a conftest pin is the right place to solve it once for the whole session. Verified the placement too: pytest.ini's `python_files` allowlist only collects under `tests/`, so this conftest loads before every collected test module and a root-level one would add nothing.

Reproduced your numbers locally: full offline suite is 8000 passed, 8 skipped, and pre-commit is clean on the new file.

One small tweak pushed: the comment referred to "the PR description", which isn't reachable from the file once merged, so I inlined the mechanism and left a #7415 pointer for the long version. No behaviour change.

Thanks!

PR #7416, opened 2026-09-15
New Source: Wagga Wagga City Council (NSW)

### LINKED ISSUE #4387: [Source Request]: wagga wagga
### Municipality / Region

Wagga Wagga Australia 

https://wagga.nsw.gov.au/imagesfiles/documents/services/waste/bin-collection-calendar-zone1.pdf

### Example Address

For Marshall St Ashmont NSW Australia , unfortunately you have to provide a street number for the it to work 

https://wagga.nsw.gov.au/services/waste-and-recycling/what-is-my-bin-day

### Collection Data Format

As PDF Download

### Additional Information

_No response_

CozyRocket:
Disclosure: I did get Claude to write this integration.

## Description

Adds a new source for Wagga Wagga City Council, NSW: `wagga_nsw_gov_au`.

The council's "What Is My Bin Day?" tool (https://wagga.nsw.gov.au/services/waste-and-recycling/what-is-my-bin-day) looks like it needs a Google Places address + a reCAPTCHA, but neither is actually required server-side:

- The reCAPTCHA token is never validated — a plain request without one still returns full results.
- The real lookup key is a pair of MGA Zone 55 (GDA94, EPSG:28355) coordinates, not the address text. The site's own frontend geocodes the address client-side and reprojects it to MGA55 before making the request.
- Results are returned as server-rendered HTML with the schedule also embedded as a JSON comment.

The address is geocoded via Esri's World Geocoding Service (`service/ArcGis.py`, already a project dependency used by other AU sources), asked to output MGA Zone 55 coordinates directly via `outSR`. This resolves to house-level/rooftop precision (verified within ~8m of the council's own tool for a real address), so an address alone is enough to reliably return all three rosters (green waste, domestic waste, recycling) — no manual coordinate lookup needed for the common case.

`x`/`y` MGA Zone 55 coordinates remain available as an optional override, for the rare case where geocoding doesn't land precisely enough for a given property (e.g. right on a service-zone boundary): they're trivial to get once, from the URL the council's own tool navigates to after a search.

The site also sits behind a WAF that fingerprints the TLS handshake itself (not just headers) — `curl_cffi` with `impersonate="chrome124"` is used for the actual bin-day request to get past this; it's already a project dependency.

## Checklist

- [x] I have read the [CONTRIBUTING.md (https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/CONTRIBUTING.md) document.
- [x] I have run `test_sources.py` locally against my own address and confirmed correct results.
- [x] I have not modified any of the auto-generated files (`sources.json`, `README.md`, `info.md`, translations, `doc/ics/*`).

Closes #4387


CozyRocket:
Heads up: I've also opened #7418 against release/3.0.0 for the same provider, as a BaseSource pipeline conversion. Both PRs add wagga_nsw_gov_au.py at the same path with different content, so whichever merges second - or the next master→release/3.0.0 sync, whichever comes first - will hit an add/add conflict there. When that happens, the pipeline version in #7418  should win; doc/source/wagga_nsw_gov_au.md from this PR shouldn't carry over, since pipeline sources get that generated automatically.

bbr111 (changes requested):
Thanks for this, and thanks especially for the write-up in the description — the WAF/TLS-fingerprint finding and the "the reCAPTCHA isn't actually validated, the real key is an MGA55 coordinate pair" analysis are genuinely useful, and reusing `service/ArcGis.py`'s `geocode()` with `outSR=28355` instead of hand-rolling the projection is exactly the right instinct.

The code itself is in good shape: `tests/test_source_components.py` passes (42 passed), `ruff check` / `ruff format` are clean, no auto-generated files in the diff, `COUNTRY = "au"` is valid, `ICON_MAP` uses the `Icons` enum, exceptions are `SourceArgumentNotFound`, and `doc/source/wagga_nsw_gov_au.md` is there. I also confirmed Wagga isn't already covered by any of the shared platforms.

I'm requesting changes purely on **branch targeting**, not on quality. As of #7399 (`CONTRIBUTING.md` and the PR template), new sources should target `release/3.0.0` rather than `master` while development focus is on v3. You've already opened #7418 for exactly that, so let's consolidate there — I'd like to **close this PR in favour of #7418**, which also removes the add/add conflict you flagged on `.../source/wagga_nsw_gov_au.py` and keeps `doc/source/wagga_nsw_gov_au.md` out of `release/3.0.0` (pipeline sources get that generated by `doc_generator.py`). Your read in that comment was right on both counts.

Three things from reviewing this version that are worth carrying into #7418:

1. **Roster completeness.** Here, `fetch()` raises if any one of the three rosters is missing and throws away everything it did parse — a property served by only some streams would get a hard failure instead of a partial calendar. Good news: #7418 already relaxes this to "raise only when nothing parsed" plus `RAISE_ON_EMPTY = True`. Keep it that way.
2. **Projection horizon.** `range(12)` (and `Schedule(..., 12)` in #7418) gives ~12 weeks of weekly green waste but ~24 weeks of fortnightly domestic/recycling. Projecting to a fixed *horizon* rather than a fixed *count* would make the three calendars consistent.
3. **Predicted vs confirmed dates.** I decoded your cassette to check — the page really does expose only the next date per bin plus a frequency word (the JSON comment is just run-day/run-week), so forward projection is unavoidable and is fine; it's derived live, not hardcoded. But the projected entries can't reflect public-holiday shifts, so a sentence in `HOWTO` saying later entries are predicted would set expectations well.

One note that only applies to this legacy version and not to #7418: `PARAM_TRANSLATIONS` / `PARAM_DESCRIPTIONS` are missing, so `address` / `x` / `y` would render unlabelled in the config flow. #7418's `PARAMS` descriptors handle that already, so nothing to do.

Please close this one and we'll continue the review on #7418. Thanks again for doing the porting work up front.

bbr111:
Since you already flagged here that #7418 (the pipeline version) should win any conflict, and development is currently focused on `release/3.0.0` rather than adding new legacy-style sources to `master` — would you like to close this one yourself in favor of #7418? Happy to leave it open if you'd rather keep it as a `master`/2.x fallback, just flagging our general preference.

CozyRocket:
Closing pull request, we will just move forward with #7418 as development focus has shifted to 3.0 release.

Issue #7139, opened 2026-08-05
v3: the reuse gate cannot see provider logic passed as a prepare= callback


markvp:
`test_pipeline_sources_reuse_shared_components` enforces the campaign's reuse rule by looking for a `Retriever` or `Parser` subclass declared in a source module. Provider behaviour passed as a plain function is invisible to it.

`source_local_step_classes` matches step *classes* only. A source that writes its provider logic as a module-level function and hands it to a component, most commonly `YearlyRetriever(prepare=...)`, passes the gate with the behaviour sitting exactly where the rule says it must not.

## How it was found

#7100 asks for a shared decoder for the abfallkalender vendor module. Both consumers keep their decoders as module-level functions:

- `frankenberg_de.py:51-65` `_resolve_district`, `:68-96` `_resolve_street`
- `zva_sek_de.py:33-39` `_js_options`, `:42-56` `_resolve_js_id`

Both sources are fully migrated and neither appears in `SOURCES_WITH_LEGACY_STEP_OVERRIDES`. The duplication survived both migrations because no gate could see it. `frankenberg_de.py:13-17` even ships a docstring telling the next contributor not to extract the logic.

The two copies have already drifted into four different readings of one reply format, two of them carrying bugs. That is precisely the outcome the reuse rule exists to prevent: "the same logic gets written again, slightly differently, and the two drift".

## Why this shape keeps recurring

This is the third gate found this week that reads as comprehensive while missing a whole form of the thing it checks:

- `test_pipeline_sources_ship_a_cassette` counted a source as recorded when its fixture directory existed but was empty (fixed in #7135)
- `_is_gated` and `_discover_new_style_sources` use `PARAMS` truthiness as a proxy for "is a pipeline source", so 21 zero-parameter pipeline sources are outside the entire v3 gate suite (#6935)
- this one

Worth stating as a rule when writing a gate: match on the property you care about, not on a proxy that usually correlates with it.

## Suggested fix

Widen the detection to provider logic reachable from a source module regardless of how it is packaged: module-level functions referenced from a component's callback arguments (`prepare=`, `fetch=` and similar). A narrow version keyed on those argument names is provably safe and catches the known cases.

Prefer a gate with no allowlist. If the widened gate lights up sources that are genuinely single-consumer, that is a finding worth reading rather than a reason to add exemptions.

bbr111:
[...]

PR #7418, opened 2026-09-15
New Source: Wagga Wagga City Council (NSW) for V3.0.0

### LINKED ISSUE #4387: [Source Request]: wagga wagga

CozyRocket:
Disclosure: I did get Claude to write this integration.

## Description

Adds Wagga Wagga City Council, NSW (`wagga_nsw_gov_au`) as a `BaseSource` pipeline source, targeting `release/3.0.0`. This is the same provider as #7416 against `master`, written for the new pipeline architecture instead of the legacy `fetch()` contract.

The council's "What Is My Bin Day?" page isn't on any existing shared platform, so retrieval and parsing are source-specific, but composed from shared pipeline components rather than hand-rolled:

- **Geocoding**: the address is resolved to MGA Zone 55 (GDA94, EPSG:28355) coordinates via the already-shared `service.ArcGis.geocode()` (`outSR`), at house-level precision — no custom projection math, no API key.
- **Retrieval**: `retrievers.HttpGetRetriever`, using the shared curl_cffi session (browser-impersonating by default, which is what gets past the site's WAF — no source-specific headers/impersonation needed here, unlike the legacy version).
- **Parsing**: a small local `_BinDayPageParser` (the council's own `binDay__*` HTML structure isn't shared by any other provider, so this stays source-local, but as a configured component instance rather than an overridden pipeline method).
- **Recurrence expansion**: `preprocessors.RecurrenceExpander` / `Schedule`, instead of a hand-rolled date-stepping loop.
- **Type mapping**: `transformers.ICSTransformer` with a `type_value_map` onto the canonical `GARDEN_WASTE` / `GENERAL_WASTE` / `RECYCLABLES` types.

## Testing

- `ruff check` / `ruff format` / `mypy` / `bandit`: clean.
- `tests/test_source_components.py`, `tests/test_doc_generation.py`, `tests/test_doc_examples.py`: pass.
- `tests/test_new_architecture.py` (pipeline-specific gates: cassette requirements, reuse-of-shared-components, no local step overrides): pass.
- Live `test_sources.py -s wagga_nsw_gov_au -l` against both TEST_CASES (address-only, and address+manual x/y override): both return 36 correct, identically-dated entries.
- Recorded offline cassettes included under `tests/fixtures/wagga_nsw_gov_au/` per the cassette rule for pipeline sources.

## Checklist

- [x] I have read `doc/contributing_source.md` and the pipeline-source section of `CLAUDE.md`.
- [x] I have run `test_sources.py` locally and confirmed correct results.
- [x] I have recorded and committed offline cassettes for both TEST_CASES.
- [x] I have not modified any auto-generated files. `doc/source/wagga_nsw_gov_au.md`  is intentionally NOT included — it's generated from the class metadata for pipeline sources.

Closes #4387 (pipeline port of #7416)


bbr111 (changes requested):
Thanks for the v3 port, and for the write-up — the cassettes are well recorded (both carry a `body`, so they pin the request payload, and the coordinate-override case correctly replays with no geocode call), and the metadata, `WASTE_TYPES`, `RAISE_ON_EMPTY` and exception handling are all right. `tests/test_source_components.py`, `tests/test_new_architecture.py`, the offline replay and the pre-commit hooks all pass as submitted. Three things to sort out before this can go in, all in the same area: the source keeps behaviour locally that the platform already has a home for.

**1. `_BinDayPageParser` is a source-local pipeline step.** It passes `test_pipeline_sources_reuse_shared_components` only because it declares no base class — that gate matches on base names containing `Parser`/`Retriever`, so writing it the idiomatic way (`class _BinDayPageParser(Parser[list[dict]])`) would fail CI. `SOURCE_LOCAL_STEP_EXCEPTIONS` is empty: no pipeline source in the tree defines its own step, and this one shouldn't be the first. This is the same shape as #7139, where hand-rolled retrieval was spelled `def` instead of `class` and so slipped past a gate that read as comprehensive.

The good news is that you don't need a new shared component at all. Replaying your own cassette, `parsers.HtmlParser("div.binDay__serviceDetails")` selects exactly the three sections:

    binDay__greenWasteDetails    | Collected weekly      | Wed 16/09/2026
    binDay__domesticWasteDetails | Collected fortnightly | Wed 23/09/2026
    binDay__recyclingDetails     | Collected fortnightly | Wed 16/09/2026

Everything `_describe` needs is reachable from those tags, and reading a base date and cadence out of a provider's own layout is precisely what `RecurrenceExpander`'s `describe` callable is documented for. So: `parse = parsers.HtmlParser("div.binDay__serviceDetails", require=[...])`, and move the `h4`/`.binDay__collectionDate` reads into `_describe`. The local class then disappears, and the source gets shorter.

**2. `_query_params` duplicates `ArcGis.geocoded_params()`.** That helper already exists for exactly this case — "a provider that is not itself on ArcGIS but keys its endpoint by a lat/lon" — including the same `ArcGisGeocodeError` → `SourceArgumentNotFound` translation you wrote out by hand; its docstring even shows the `HttpGetRetriever(url=..., params=ArcGis.geocoded_params(...))` composition. The only things it lacks are a non-default `out_sr` and your manual x/y override. Please add those two to the shared helper and configure it from the source, rather than keeping a private copy: the next AU council on a projected-CRS endpoint then gets both for free.

**3. The "no schedule found" guard never fires.** `_SCHEDULE_COMMENT_RE` uses `[^}]*`, which cannot cross the inner `}` of the nested comment payload (`{"schedule":{...},"errorMessage":null}`), so `\}\s*-->` never anchors. Replaying the recorded response, the regex returns `None` — the whole guard block is skipped on every successful fetch, and the failure path has no cassette, so it is unverified in both directions. Rather than patching the regex, use `parsers.ArgumentGuard`, which is the shared component for this exact situation (a provider answering an unknown lookup with HTTP 200 and an ordinary page), e.g. `ArgumentGuard(HtmlParser(...), argument="address", contains="binDay__collectionDate", hint=...)`. A third cassette for an out-of-area address would then give that path offline coverage too.

Two smaller, non-blocking notes:

- `Schedule(..., step, 12)` gives ~12 weeks of weekly green waste but ~24 weeks of the fortnightly rounds. Per-cadence counts, or `Schedule(until=...)` windows, would give every round the same horizon.

Once the parse step is composed from shared components and the geocoding goes back through `ArcGis`, this should be a quick approve. Re-record the cassettes after the change and ping me.

CozyRocket:
@bbr111 
Thanks for the detailed review — pushed a commit addressing all three.

**1. `_BinDayPageParser` → `parsers.HtmlParser`**: done as suggested. `parse = parsers.HtmlParser("div.binDay__serviceDetails")`, with the `h4`/`.binDay__collectionDate` reads moved into `_describe`, reading straight off the selected `Tag`. The local class is gone.

**2. `_query_params` → `ArcGis.geocoded_params()`**: added `out_sr` and `override_lon_param`/`override_lat_param` to the shared helper (docstring updated with the composition you described) and configured it from the source instead of keeping a private copy. Confirmed `jacksonville_fl_us` (the only other caller) still replays correctly with the new optional kwargs defaulted off.

**3. Regex guard → `parsers.ArgumentGuard`**: done, but the exact marker in your example doesn't actually work — `"binDay__collectionDate"` as plain text is present in the page's static `<style>` block on *every* response, success or failure (it's a CSS selector: `.binDay__serviceDetails .binDay__collectionDate { margin-left: .5rem; }`), so a guard checking for that substring would never fire — same failure mode as the bug you flagged, just moved. Confirmed live against an out-of-area address (Sydney CBD, which geocodes fine but returns no schedule): the literal string `binDay__collectionDate` is present, but `class="binDay__serviceDetails` (the actual rendered element's opening tag) is not. Used that instead — verified live it's present for an in-area address and absent for the out-of-area one.

Didn't end up adding a third cassette for that failure path: `tests/record_fixtures.py`'s `recording()` only persists a cassette when `fetch()` returns without raising (`cassette.py`'s `success` flag is never set on an exception), so an `ArgumentGuard`-triggering case can't go through the standard tool as-is. Neither `mzv_rotenburg_bebra_de` nor `wellington_govt_nz` (the tree's other two `ArgumentGuard` users) have a failure-path cassette either, so I didn't want to be the first to route around that in a source PR. Verified the guard live instead (address swapped out before committing). Happy to file a separate issue about teaching `record_fixtures.py` to capture an expected-exception case, if that's useful — didn't want to scope-creep it into this one.

**Non-blocking notes**: switched `Schedule(..., step, 12)` to a per-cadence count (26 weekly / 13 fortnightly), so every round now gets the same ~6-month horizon.

Cassettes re-recorded (52 collections each now, up from 36, since the horizon evened out). `ruff`/`mypy`/`bandit`/`test_source_components.py`/`test_new_architecture.py`/`test_doc_generation.py`/offline replay all pass.

bbr111 (approved):
Thanks — all three points are properly addressed, and I've verified each against the recorded cassettes rather than taking the commit message for it.

1. `_BinDayPageParser` is gone; `parse = ArgumentGuard(HtmlParser("div.binDay__serviceDetails"), ...)` with the date/frequency reads in `_describe` off the selected `Tag` is exactly the shape I was after. No local step subclass, no overridden pipeline method, and `_describe` issues no HTTP so it isn't the hand-rolled-retrieval shape either.

2. `out_sr` + `override_lon_param`/`override_lat_param` on the shared `ArcGis.geocoded_params()` is the right layer, and the defaults keep `jacksonville_fl_us` byte-identical (its replay still passes). The cassettes confirm both paths: the by-address recording has the geocode call with `outSR={"wkid": 28355}` then the council page; the by-coordinates recording has the council page only, no geocode.

3. You're right and I was wrong about the marker — `binDay__collectionDate` appears 5x in a success response, two of them inside the static `<style>` block (`.binDay__serviceDetails .binDay__collectionDate { ... }`), so a guard on that string would never have fired. Thanks for checking rather than just applying the suggestion. `class="binDay__serviceDetails` is a sound substitute.

Not adding a failure-path cassette is the right call given `record_fixtures.py`'s `recording()` only persists on a clean `fetch()` — please do open that separate issue about capturing an expected-exception case, it would help the other `ArgumentGuard` sources too.

Verified locally at 2ea01725: full `pytest -m "not live"` (8025 passed, 8 skipped, since `ArcGis.py` is shared), `test_source_components.py` + `test_new_architecture.py` (6112 passed), `test_declared_waste_types.py -k wagga`, and `pre-commit run --all-files` all green. Both cassettes correctly reject a mutated request body under `-p tests.mutate_requests`, so they pin the payload rather than just method+URL.

Two non-blocking notes, nothing to change now:
- `class="binDay__serviceDetails` is also a prefix of `binDay__serviceDetailsWrapper`. That's the safe direction to be loose in (`RAISE_ON_EMPTY` catches an empty result anyway, whereas a trailing-quote match would break the day the council adds a second class), so I'd leave it — maybe just note it in the existing comment.
- In `geocoded_params`, supplying only `x` or only `y` silently falls back to geocoding and drops the value given. Fine as documented; a `SourceArgumentNotFound` on the half-filled pair would be friendlier if you ever revisit it.

I've asked over on #7416 whether you'd like to close that one in favor of this pipeline version, per your own note there.

Approving.

CozyRocket:
@bbr111 
Pushed the two non-blocking follow-ups from review in [aaf15fed](https://github.com/CozyRocket/hacs_waste_collection_schedule/commit/aaf15fed):

- `ArcGis.geocoded_params()`: supplying only one of `override_lon_param`/`override_lat_param` now raises `SourceArgumentNotFound` on the missing coordinate, instead of silently falling back to geocoding and dropping the value that was given.
- `wagga_nsw_gov_au`: added a comment noting the `ArgumentGuard`'s `contains` check is deliberately a prefix match (it also matches `binDay__serviceDetailsWrapper`), since that's the safe direction to be loose in given `RAISE_ON_EMPTY`.

Full non-live suite + `test_source_components.py`/`test_new_architecture.py`/`test_declared_waste_types.py -k wagga` + `pre-commit run --all-files` all green locally. Let me know if there's anything else before merging.

bbr111:
Both non-blocking notes addressed cleanly — the half-filled override pair now raises `SourceArgumentNotFound` instead of silently dropping a value, and the prefix-match comment makes the intent explicit. Re-verified `test_source_components.py`, the offline fixtures, and pre-commit at `aaf15fed` — all green, no behavior change for `jacksonville_fl_us`. LGTM.

PR #7419, opened 2026-09-15
Fix kiedysmieci_info: move to the kiedysmieci.info proxy API

kjanas:
## Summary

`kiedysmieci_info` talks directly to the provider backend at
`cloud.fxsystems.com.pl:8078`. kiedysmieci.info now fronts the same data
through `schedule-proxy.php`, so this moves the source over and fixes a
refresh bug found on the way.

The old endpoint is still up, so this is not an outage fix — but the proxy
removes two long-standing warts:

- **Custom SSL context gone.** The direct host needed
  `ssl_ctx.set_ciphers("DEFAULT")` and raw `urllib` plumbing to connect at
  all. The proxy is a normal HTTPS host, so the source now uses `requests`
  like the rest of the tree.
- **Complete location cascade.** `dostepne_gminy/v5` returned an incomplete
  municipality list, worked around in #6066 by validating locations through
  the streets endpoint. The proxy returns every level
  (`listaWojewodztw` / `listaPowiatow` / `listaGmin` / `listaUlic`), so that
  workaround is gone and a wrong argument now reports suggestions for the
  level it actually belongs to. An unknown street is caught too: the API
  answers those with HTTP 200 and a single `"brak harmonogramu"` placeholder
  entry, which the old code passed through as a collection.

### Refresh bug

The schedule was fetched in `__init__` and `fetch()` only iterated over the
stored list. `SourceShell.create()` constructs the `Source` once and calls
`fetch()` on every update, so the schedule was frozen at setup time and only
changed when Home Assistant restarted. The fetch now happens in `fetch()`.

A healthy update costs exactly one request; the location lists are queried
only to pinpoint a bad argument after something failed.

### Further changes

To avoid having to submit an empty form just to populate the dropdowns
fetched from the API, I am going to implement a multi-step `config_flow`,
building on the not-yet-released 3.0.0 changes. Once that is done, the
wizard will only ever offer current, valid data coming from the source,
with no free-text fields. That work is out of scope here and will come as a
separate PR against `release/3.0.0`.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

- [ ] `doc/source/<name>.md` created for new sources — n/a, arguments unchanged

## Testing

`TEST_CASES` are unchanged, so existing user configurations keep working:

```
Testing source kiedysmieci_info ...
  found 42 entries for Nadolany, podkarpackie, sanocki, Bukowsko
  found 18 entries for Kędzierz, podkarpackie, dębicki, Dębica
  found 19 entries for Parkowa, lubelskie, zamojski, Szczebrzeszyn
```

`test_sources.py -s kiedysmieci_info -d` reports no difference between two
consecutive fetches. Error paths were checked by hand — a bad voivodeship,
district, municipality and street each report suggestions for their own
level, and lookups are case-insensitive.


bbr111 (approved):
Nice cleanup — thanks! Verified against the live proxy: all three TEST_CASES fetch (42 / 18 / 19 entries), and I checked the error paths by hand too — a bad voivodeship, district, municipality and street each report suggestions for their own level, and the lookups are case-insensitive as advertised.

Dropping the custom SSL context and the #6066 streets-endpoint workaround is a real simplification, and moving the fetch out of `__init__` fixes a genuine refresh bug: `SourceShell.create()` builds the `Source` once, so the schedule was frozen at setup time until Home Assistant restarted.

One small tweak pushed: the final "No schedule published for …" raise is caused by the user's municipality/street pair, so it's now a `SourceArgumentExceptionMultiple(("municipality", "street"), …)` rather than a bare `Exception` — that lets the HA UI attribute the error to the right fields. The provider-level raise in `_get()` I left as a plain `Exception`, since that one really is a transport failure and not an argument problem.

Optional follow-up, not blocking: the live run also returns `popiół` and `elektrośmieci`, which fall through `ICON_MAP` to the default. `elektrośmieci` would map nicely to `Icons.ELECTRONICS` if you want to add it sometime.

`tests/test_source_components.py` passes (42 passed) and pre-commit is clean.
Looking forward to the multi-step config flow PR against `release/3.0.0`.

kjanas:
@bbr111 Good catch about ICON_MAP. Tried to solve it the smart way by getting distinct list of waste types across all voivodeships/districts/municipalities, found bunch of typos as well :). Unfortunately, it looks each municipality keeps own list and naming convention for a waste types and there is no validation or simple spelling check.

Any feedback much appreciated!

bbr111 (approved):
Reviewed the two follow-up commits too — nice work, this is exactly the kind of mapping I'd flagged as an optional nit.

Verified: `pytest tests/test_source_components.py -q` → 42 passed, `ruff check` / `ruff format --check` clean on the current head. The `_icon_for()` position-based tie-break for bundled rounds ("odpady wielkogabarytowe, opony, elektrośmieci" → BULKY, since it leads) reads correctly, and `Icons` is a `StrEnum` so the `min()` comparison over `(position, icon)` tuples can't blow up even on a rare tied position.

Good call keeping ash and the still-unmapped streams (opony, gruz, budowlan) on the default rather than mislabeling a bundled round after a keyword that isn't actually leading.

Approving.

PR #7420, opened 2026-09-15
fix(ipswich_qld_gov_au): stop reporting an exhausted Google quota as an unknown street

CRZTFR:
## The bug

`ipswich_qld_gov_au` resolves the address through the council app's own Google geocoding key (read from the app's `/Config` endpoint) before asking for a schedule. That key is shared by every user of the app and is regularly over its daily quota, at which point Google answers `OVER_QUERY_LIMIT` with an empty `results` list and an `error_message`.

`_geocode()` never read `status`. It iterated `results`, found nothing, and raised `SourceArgumentNotFound("street", ...)` — the same error an unknown street produces. The visitor was told to correct an address that was right all along.

Measured from here, seven of ten consecutive lookups for the existing `Random` TEST_CASE failed this way:

```
geocode:OVER_QUERY_LIMIT/n=0: 7
geocode:OK/n=1: 3
```

The council's config carries a second key (`googleAddressSearchURLPlaces`), but it answers `REQUEST_DENIED` to server-side requests, so it is not a usable fallback.

## The fix

The geocoding turns out not to be needed at all. The service matches on the address components, not the coordinates — a payload carrying `lat/lng` of `0/0` resolves the same 210 entries as the rooftop-accurate result Google returns. I confirmed this by substituting coordinates ~40 m away and getting an identical schedule.

What it does require is the `administrative_area_level_2` council-area component and a post code belonging to the suburb. `WhatBinDayService.build_address_data()` emits neither, which is why simply calling it in place of Google returns "No Bin Sevices Found", and a wrong post code returns "Device Key not valid for location".

So:

- `post_code` becomes an optional argument. When given, the address is built locally and Google is never called.
- When it is not given, the previous path still runs, but a non-`OK`/`ZERO_RESULTS` status now raises a `ValueError` naming the status instead of blaming the street.

## Testing

All three TEST_CASES carry a `post_code` and pass live:

```
Camira State School: OK 208 entries
Random:              OK 210 entries
Ipswich CBD:         OK 210 entries
```

I deliberately did not leave a case without a `post_code`: while the shared key is out of quota such a case fails on a third party's billing rather than on this source, so CI would be red for reasons no one here can fix. The fallback path is still exercised by anyone who omits the argument, and it now reports the real cause.

`ruff check` and `ruff format --check` clean. Doc updated.


bbr111 (approved):
Thanks — the diagnosis of the underlying bug is spot on, and the measurement in the description (7 of 10 lookups returning OVER_QUERY_LIMIT) is exactly the kind of evidence that makes a fix like this easy to review. Reading `status` before iterating `results` is clearly right: an exhausted key and an unknown street were genuinely indistinguishable before.

I pushed one simplification on top. The description says `build_address_data()` can't be used because it emits neither the `administrative_area_level_2` council area nor a post code. I probed the live service to confirm that before merging, and it turns out to be one component off:

- `administrative_area_level_2` ("Ipswich City") isn't required at all — dropping it resolves all three test addresses with identical entry counts (210 / 210 / 208).
- `build_address_data()` does emit `postal_code`.
- The real blocker is `administrative_area_level_1.long_name`. "Queensland" works, "QLD" fails with "No Bin Se[r]vices Found" for every address. The shared builder writes its `state` argument into both `long_name` and `short_name`, so `state="QLD"` fails and `state="Queensland"` works — that's the entire difference.

So `_address_data()` is now a six-line call to the shared builder with `state="Queensland"` and `coordinates={"lat": 0, "lng": 0}`, and the 47-line local payload is gone. No change to `service/WhatBinDay.py`, so the other three WhatBinDay sources are unaffected. Your street-splitting guard and its `SourceArgumentNotFound` are kept as-is.

Two smaller things: the non-OK status now raises `RuntimeError` rather than `ValueError`, matching the "Device configuration failed" raise twenty lines above it in the same method; and I reworded a docstring to avoid quoting the provider's misspelled "No Bin Sevices Found" literally, since codespell rejects it.

Agreed on not adding a TEST_CASE without a `post_code` — a case that fails on someone else's billing quota isn't a useful CI signal.

Live run after the change: 210 / 210 / 208 entries for the three cases. `pytest tests/test_source_components.py` 42 passed, pre-commit clean.

PR #7423, opened 2026-09-15
Add source: Gemeinde Ebbs, Tyrol, Austria (ebbs_gv_at)

rr62wghzwp-blip:
## Summary

Adds a new source for Gemeinde Ebbs (Tyrol, Austria), which publishes its
waste-collection calendar via the shared RiSKommunal platform. Also extends
the shared `RiSKommunalRetriever` with `zone_param`/`zone_typids` support,
for installs (like Ebbs) whose combined calendar merges both zones into one
table with no filterable zone column, but that publish a fixed `typids` set
per zone.

## Type of change

- [x] New source

## Checklist

- [x] N/A -- pipeline source generates its own doc
- [x] TEST_CASES use real, publicly accessible zones (not an address)

- [x] Cassette recorded (`tests/fixtures/ebbs_gv_at/zone_1.json`, `zone_2.json`).
- [x] Composes shared components: no local Retriever/Parser subclass, uses
      `RiSKommunalRetriever`/`RiSKommunalParser`.
- [x] Anything missing was added to the shared component
      (`zone_param`/`zone_typids` on `RiSKommunalRetriever`) rather than a
      source-local step.


bbr111 (approved):
Nice one — thanks for putting the zone handling into the shared RiSKommunal
component rather than the source module, and for the comment explaining why
the retriever's zone_typids and the parser's zone_param are different
mechanisms. That distinction is exactly the kind of thing the next person on
this platform will need.

I verified the provider independently: ebbs.gv.at/Muellabfuhrtermine does
publish precisely those two pre-filtered typids links, and querying each live
reproduces what you describe (zone 1 gets Gelber Sack, zone 2 does not, both
share Biomüllabfuhr). Both cassettes replay cleanly and WASTE_TYPES matches
the replayed union exactly.

I pushed two small tweaks:

1. `_zone_typids` passed `self.zone_param` (typed `str | None`) straight into
   `source.params.get()`, which fails the pinned pyright hook in CI with
   reportArgumentType. Narrowed it into a local once, which also tidies the
   repeated `or` fallbacks.
2. `VALID_ZONES` restated the `_ZONE_TYPIDS` keys, so a new zone would have
   needed editing in two places. It is now derived from the map.

Full offline suite is green (8025 passed) and pre-commit passes on all hooks.

Two things I deliberately left alone, just noting them: the source's `__init__`
now duplicates the zone validation the retriever does, but it matches
eggelsberg_at.py and gives a better config-flow error, so it stays. And
`zone_param`/`strasse_param` are documented as mutually exclusive while
`__call__` silently prefers `strasse_param` — worth a guard in the retriever
someday, but not in this PR.

Approving.

rr62wghzwp-blip:
Thanks for the review and for verifying it live, and for the cleanups — much appreciated!

PR #7425, opened 2026-09-15
fix: accept f_abfallarten from the UI config flow

thebino:
## Summary

Fixing a list from the config flow

> **New source?** Development focus is currently on the upcoming v3.0.0 release — please target this PR at the `release/3.0.0` branch instead of `master`. Bug fixes for the current stable release still go against `master`.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist

⚠️ The `AWB Limburg-Weilburg` returns a 401 in the test. It is unrelated to this PR. Api Key seems to not be valid anymore


bbr111 (approved):
Nice catch — this was genuinely broken. Without the annotation, `config_flow` fell back to `cv.string` for `f_abfallarten`, so a UI-entered `31,17,19,218` was iterated character by character in `fetch()` and produced twelve one-character `f_id_abfalltyp_*` fields. Annotating it as `list[str] | str | None` hits the exact `SUPPORTED_ARG_TYPES` key that maps to a multi-value `TextSelector`, and keeping the comma-string branch means existing config entries keep working. Both halves are right.

I checked the other new annotations for config-flow regressions: `str | int` maps to `cv.string` (same as the previous required-arg fallback) and `str | int | None` resolves through the union walk to `cv.string`, so nothing else changes shape.

I added one small thing: a `TEST_CASES` entry exercising the comma-separated form, since every existing case either omits `f_abfallarten` or passes a list — the new branch had no coverage. Verified live, it returns the same 18 entries as the equivalent list form.

Live-tested the whole source: no change in entry counts versus master, and `stadt_kerpen_de` (which subclasses this `Source`) still works. The `AWB Limburg-Weilburg` 401 you flagged is indeed unrelated — I reproduced it on master before applying your branch; it's the source's own error telling users to move to `abfall_io_graphql`.

One optional follow-up, not for this PR: `stadt_kerpen_de.py` re-declares `__init__` without annotations, so its `f_abfallarten` field still renders as plain text. It works now thanks to your shim, but the same annotation there would make the UI consistent.

Thanks!

thebino:
Great, feel free to merge

PR #7426, opened 2026-09-15
Preserve original waste labels lost to type_value_map collapsing

bbr111:
## Problem

Several pipeline sources map multiple distinct provider labels (bin-size or rhythm variants of the same waste stream) onto one canonical `WasteType`, e.g. `abfall_neunkirchen_siegerland_de`'s "Restmülltonne" / "Spartonne Restmüll" / "Container Restmüll" and `koppl_at`'s "Restabfall 14-tägig" / "Restabfall monatlich", all resolving to General Waste.

Two problems follow:
- The original distinction is lost — no way to tell them apart, or to hide the one that doesn't apply to a given address.
- Where the underlying schedules overlap on some dates (verified against both cassettes: Koppl's "monatlich" dates are an exact subset of its "14-tägig" dates; Neunkirchen's "Spartonne Restmüll" is a subset of "Restmülltonne"), canonicalisation turns two previously-distinguishable entries into visible duplicates on the same day.

## Fix

**Carrying the raw label (`transformers.py`):** `BaseTransformer` (all subclasses) gains `carry_raw_label`, off by default: when true and the source hasn't supplied its own `description`, the provider's raw label becomes the Collection's `description` instead of being discarded. Skipped when it would just repeat the canonical display name.

**User control over both the label and the dedup, per config entry, with sensible defaults:**
- `SourceShell.fetch()`'s dedup keys on `Collection._identity_key` (customize alias if set, else the canonical, locale-independent `WasteType.id`) instead of the displayed `.type`, and folds a discarded duplicate's `description` into the entry that's kept instead of dropping it — folded into the *existing* `ignore_duplicates` option (`CONF_IGNORE_DUPLICATES`, shipped v2.25.0) rather than adding a second mechanism. `BaseSource.IGNORE_DUPLICATES_DEFAULT` is a source's declared default for that option (class attribute for pipeline sources, module-level variable of the same name for legacy — same class-then-module precedence as `TITLE`/`DESCRIPTION`/`URL`), read the first time a config entry's options are opened; an explicit user choice always wins afterward, in either direction.
- New option `show_original_label` (`CONF_SHOW_ORIGINAL_LABEL`, default `True`) controls whether the `carry_raw_label` fallback is visible. `Collection` gained a provenance marker (`description_is_raw_label_fallback`, set only via `set_description_from_raw_label`) so turning this off clears *only* the auto-carried fallback text, never a source's genuine description (e.g. real ICS `DESCRIPTION` via `description_key`) — those were indistinguishable on the finished `Collection` before. Unlike `ignore_duplicates`, there's no per-source default to resolve here: a source either can produce the fallback or it can't, and showing it when it can is the sensible default the source already opted into — so this is a plain `True`/`False`.
- Both options are now also settable via **YAML** (`init_yaml.py`'s `SOURCE_CONFIG`), closing a gap for `ignore_duplicates` that predated this PR (UI-only since #6368/v2.25.0, no YAML equivalent at all). `WasteCollectionApi.add_source_shell()` forwards both through.

Neither mechanism lives inside `BaseSource.fetch()`/the transformer itself: a bare `Source().fetch()` call — what `tests/test_offline_fixtures.py`, `tests/dump_collections.py` and `test_sources.py -l` all do — has no HA config entry and therefore no user override to honour. Moving both to `SourceShell` (the layer that already owns customisation/filtering/day-offset) means a user's choice always reaches them. Trade-off: those bare-fetch tools show the raw, un-merged, always-labelled output for the two affected sources (verified: 44/46 entries with genuine same-day duplicates) — `SourceShell.fetch()`, the only thing an actual HA install ever runs, applies both options correctly in either direction (verified via cassette replay).

Also extends `source_shell._customize_keys()` to match `customize:` entries against a Collection's `description`, so a user can hide (or rename/re-icon) just the one label variant that doesn't apply to their address via existing `customize:` config.

## Applied to

Wired into `abfall_neunkirchen_siegerland_de` and `koppl_at`, the two sources that surfaced the issue. Left off by default for the ~24 other sources with the same `type_value_map`-collapsing pattern (RiSKommunal AT, various DE portals, `1coast_com_au`) — opting them in is a good follow-up, kept out of this diff to keep it reviewable.

Translations for `show_original_label` added in en/de/it/nl. `fr.json` has no `options` section at all pre-existing this PR — out of scope to backfill here.

## Testing

- `pytest tests/ -m "not live"`: 8000 passed, 8 skipped, 0 failed.
- `pytest tests/test_offline_fixtures.py -k "koppl_at or abfall_neunkirchen_siegerland_de"`, `test_declared_waste_types.py`, `test_config_flow.py`, `test_new_architecture.py`: passed.
- Manually verified via `SourceShell.create()` + cassette replay: source-declared `ignore_duplicates` default resolved and applied with no option set, explicit override wins in either direction; `show_original_label=True/False` toggles `description` on all entries without affecting the dedup count.
- Pre-commit (ruff, ruff format, codespell, bandit, mypy, pyright, check json): passed. mypy's pre-existing 19 `config_flow.py` warnings are unchanged by this diff.


bbr111:
**Follow-up fix:** \`show_original_label\` was rendered in every config entry's options form, regardless of whether that source's transform even declares \`carry_raw_label\` — a confusing, always-no-op checkbox for ~264 of ~266 pipeline sources and every legacy one, with wording ("some sources keep...") that gave no hint whether the source being configured was one of them.

Fixed: \`BaseTransformer.carries_raw_label\` (public property) + \`source_shell.source_supports_show_original_label(source_name)\` let the config flow check this per source, and the option is now omitted from the schema entirely unless it's actually relevant. Verified \`True\` for \`koppl_at\`/\`abfall_neunkirchen_siegerland_de\`, \`False\` for an unrelated pipeline source. Full suite still green (8000 passed, 8 skipped, 0 failed).

bbr111:
**Docs added:** \`doc/contributing_source.md\` now documents \`carry_raw_label\` (new subsection right after "Carrying event metadata") and \`IGNORE_DUPLICATES_DEFAULT\` (new row in the Metadata attributes table), for the next source author who hits this same type_value_map-collapsing pattern. Also added a \`CHANGELOG.md\` entry under \`[Unreleased] > Added\` consolidating the user-facing effect of this PR.

PR #7430, opened 2026-09-16
New source: Čisto mesto, Slovenia (cistomesto_si)

capreolusaux:
## Summary

Adds support for Čisto mesto Ptuj (Slovenia). The API requires an initial guest registration, so the script caches the user hash to reduce API calls. I've also added all 25 supported municipalities to `CONFIG_FLOW_TYPES` so users can select them from a dropdown during setup.

## Type of change

- [x] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


bbr111 (changes requested):
Thanks for this — nice clean source, and it runs green against all three test cases live (25 entries each for Majšperk, Cirkulane and Destrnik).

I've pushed a few small fixes to your branch to get CI green: removed the `sl` block from `PARAM_TRANSLATIONS`/`PARAM_DESCRIPTIONS` (the allowlist is currently `en/de/it/fr/nl`, and the structural test hard-fails on anything else — I'll open a separate issue about adding Slovenian properly), added an explicit `COUNTRY = "si"`, added the missing `doc/source/cistomesto_si.md`, and used an inline `codespell:ignore` marker so `Trnovska vas` stops tripping codespell.

One thing I'd like you to look at before we merge. I queried `getRegionsList` directly and it returns **26 regions but only 22 distinct titles** — Gorišnica, Markovci, Hajdina and Juršinci each appear **twice, with different IDs and genuinely different schedules**:

- Gorišnica `122` → 23 entries; Gorišnica `123` → 26 entries, different dates
- Hajdina `126` → first Papir 2026-09-28; Hajdina `127` → first Papir 2026-09-30

Because the lookup is first-match-wins and `break`s on title, a user who selects `Hajdina` from the dropdown silently gets whichever one the API listed first — potentially the wrong collection area, with no hint anything is off. And since the dropdown lists each name only once, there's currently no way to reach the second variant through the UI at all.

The API doesn't give us much to disambiguate with (`apartmentBuilding` is `0` for both in every pair; only the opaque `regionCode` differs), so I think this needs your local knowledge. Two options that would work: raise `SourceArgAmbiguousWithSuggestions` (it's already in `exceptions.py` and is meant for exactly this) listing the candidate IDs when a name matches more than one region, or add the four extra IDs to `CONFIG_FLOW_TYPES` with a label that distinguishes them. Happy with either.

capreolusaux:
Thanks for the thorough review and for pushing the CI fixes!
You were totally right about the duplicate names. I dug into the official [PDF schedules](https://cistomesto.si/urniki-odvoza-odpadkov/) on cistomesto.si and managed to map each specific region ID to the actual villages they serve on different collection days. 
Fun fact: while cross-referencing the schedules, I found out the API has a bug on their end. The region IDs for Markovci and Juršinci are actually swapped in the backend response (Markovci schedules return under Juršinci's IDs and vice versa). 
I went ahead and implemented both of your suggestions to make it as robust as possible:
1. I updated `CONFIG_FLOW_TYPES` to explicitly list all the distinct sub-regions with their corresponding villages and IDs (e.g., `Gorišnica 1 (Cunkovci...) (122)`). I also made sure to map Markovci and Juršinci to their correct physical IDs to bypass the API bug.
2. In `fetch()`, I added `SourceArgAmbiguousWithSuggestions`. If a user passes just `"Gorišnica"` in YAML, it will now throw an error with the exact choices.
3. Added a `.endswith(f"({r_id})")` check so the UI dropdown selection matches seamlessly.
I've just pushed the updated code. Let me know if you spot anything else!

bbr111 (approved):
This is exactly what was needed — thanks for digging through the official PDFs rather than guessing.

I verified the new mapping end to end against the provider's own schedules (`URNIK-<MUNICIPALITY>-{1,2}-DAN.pdf`) and the live API:

- the village lists in the dropdown labels match the PDFs verbatim,
- the collection weekdays match per ID: Hajdina 1 dan (Mon/Tue) = 126, Hajdina 2 dan (Wed/Thu) = 127, Markovci 1 dan (Wed/Thu) = 125, Markovci 2 dan (Mon/Tue) = 124, Juršinci 1 dan = 128, Gorišnica 1 dan = 122,
- so your inverted Markovci mapping is right, and all four ID pairs really do return different schedules,
- `region: "Hajdina"` now raises `SourceArgAmbiguousWithSuggestions` listing `['Hajdina (126)', 'Hajdina (127)']` instead of silently guessing, and all eight dropdown labels resolve correctly through the `endswith("(id)")` check.

I've pushed a few small things to your branch:

1. **The 401 retry wasn't doing anything useful.** The backend occasionally hands out a guest hash that `getRegionsList` rejects with `401 Not authenticated E0`, and retrying *that same hash* never recovers — I let it run 8 attempts over 8 seconds and it stayed 401. So roughly one fetch in ten was failing outright. Registration now happens inside the retry loop, so a 401 gets a fresh hash: 12/12 clean afterwards.
2. **codespell.** The new village lists trip it on `vas` -> `was`; added inline `# codespell:ignore vas` markers like the existing one on `Trnovska vas`.
3. Wrapped the long `if` condition, dropped a leftover `region_id = None`, mentioned the split municipalities in the `region` description, and expanded `doc/source/cistomesto_si.md` with the verified village/region-ID table so YAML users can look their area up.

`pytest tests/test_source_components.py` is green (42 passed) and all three test cases fetch live. Merging — thanks for the source and for the thorough follow-up!

PR #7431, opened 2026-09-16
Support recording an expected-exception cassette for pipeline sources

### LINKED ISSUE #7432: Adopt ERROR_TEST_CASES / failure-path cassettes across ArgumentGuard sources
### I propose a feature for:

The integration in general

#7431 adds the ability to record an offline cassette for a pipeline source's *expected-failure* path (e.g. an `ArgumentGuard` rejecting an out-of-area address), via an optional `ERROR_TEST_CASES` class attribute recorded by `record_fixtures.py` and replayed by `tests/test_offline_fixtures.py::test_offline_replay_expected_error`.

That PR only adds the mechanism — no source declares `ERROR_TEST_CASES` yet. This issue tracks going through pipeline sources that use `ArgumentGuard` (or another `SourceArgumentException`-raising guard) and adding an `ERROR_TEST_CASES` entry + recording its cassette (`python tests/record_fixtures.py <module>`), so the failure path is exercised offline in CI rather than only manually.

Not urgent or blocking anything — a reasonable first contribution for anyone already touching one of these sources. `wagga_nsw_gov_au` itself is a good first candidate (this is where the gap was originally noticed, in review).

CozyRocket:
`cassette.recording()` only ever persisted a cassette on a clean `fetch()`, so a source using `ArgumentGuard` (or any `SourceArgumentException`-raising guard) had no way to record its *failure* path — an out-of-area address, say. The interactions were captured but silently discarded the moment `fetch()` raised (raised during review of the wagga_nsw_gov_au PR).

- `tests/cassette.py`: `recording()` takes an optional `expect_exception`. When the wrapped `fetch()` raises a matching exception, the interactions captured up to that point are persisted (with an `expected_error` entry recording the exception's type and message) before the exception is re-raised. An unmatched exception still discards the recording, unchanged from today.
- `tests/record_fixtures.py`: new `record_errors()`, driven by an optional `ERROR_TEST_CASES` class attribute — same shape as `TEST_CASES`, for inputs a source expects to reject. Only a `SourceArgumentException` is recorded; anything else means the case doesn't exercise what it claims to, so it's reported and skipped rather than pinned as "expected".
- `tests/fixtures_support.py`: `error_fixture_path()` / `discover_error_fixtures()`. Cassettes land at `tests/fixtures/<module>/_error_<slug>.json`, prefixed like the existing `_choices.json` so `discover_fixtures()` keeps skipping them.
- `tests/test_offline_fixtures.py`: new `test_offline_replay_expected_error`, which discovers these cassettes and replays them offline, asserting the same exception type still fires.

No source declares `ERROR_TEST_CASES` yet — this adds the capability without changing any existing source's behaviour or cassettes. Follow-up issue tracks adopting it.

**Test plan**
- [x] `ruff check` / `ruff format --check` on the four changed files
- [x] `mypy --ignore-missing-imports --explicit-package-bases` on the four changed files
- [x] `bandit -c tests/bandit.yaml` on the four changed files
- [x] Full `pytest -m "not live"` (7952 passed, 9 skipped; one pre-existing unrelated Windows CRLF failure in `TestIcsConfiguredRetriever`, reproduces on `upstream/release/3.0.0` before this change too)


bbr111 (approved):
Nice gap to have spotted — that a guard's rejection path could be described but never verified offline was a real hole, and doing it as an opt-in ERROR_TEST_CASES with an underscore-prefixed cassette slots into the existing discovery cleanly. Full replay of all 682 cassettes plus every v3 gate stays green, so it really is additive.

I've pushed four small things rather than send it back:

- `record_errors()` printed "not recorded" in two branches that did leave a file behind. `recording()` counts a clean run as a success, so a case that has stopped raising got written out *without* an `expected_error` key — and the next CI run then fails on "has no recorded expected_error" for a file the recorder said it hadn't written. The wrong-exception branch had the mirror image: a stale recording from an earlier run survived and kept the replay green against an assertion the source no longer satisfies. Both now discard, the way `record()` and `record_choices()` already do.
- `SourceArgumentExceptionMultiple` is declared as a sibling of `SourceArgumentException` rather than a subclass, so a guard raising it would have been reported as "expected a SourceArgumentException, got …". Named both in an `EXPECTED_ERRORS` tuple.
- Added the `expect_exception` unit tests to `tests/test_cassette.py`. Worth flagging why: since no source declares `ERROR_TEST_CASES` yet, `_ERROR_FIXTURES` is empty and `test_offline_replay_expected_error` is parametrised over nothing — so as submitted the new harness code wasn't executed by any test in CI. The four new tests assert the persistence decision directly and issue no HTTP.
- Documented `ERROR_TEST_CASES` in the "Offline fixtures" section of `doc/contributing_source.md`, since that's where a contributor would look for it.

One thing I left alone, as a follow-up rather than a blocker: `test_pipeline_sources_ship_a_cassette` accepts `any(fixture_dir.glob("*.json"))` without filtering `_`-prefixed names, and `_unrecorded_cases()` returns early when `recorded` is empty. A source whose fixture directory held only `_error_*.json` would therefore pass both cassette gates with no fetch recording at all. It pre-exists via `_choices.json` and no directory is currently in that state, but this widens the set of files that can trigger it — worth a one-line filter in a separate PR.

PR #7433, opened 2026-09-16
Fix eastleigh_gov_uk: refresh stale Cloudflare impersonation fingerprint

### LINKED ISSUE #7424: [Source Defect]: Eastleigh Gov Uk - Error 403
### Source Name

Eastleigh Borough Council

### What's happening?

Within the last week or so Eastleigh Borough Council has started failing with a HTTP 403 error.

### Error Logs

```Shell
This error originated from a custom integration.

Integration: Waste Collection Schedule (documentation)
First occurred: 10:58:22 (2 occurrences)
Last logged: 10:58:43

fetch failed for source Eastleigh Borough Council: Traceback (most recent call last): File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source_shell.py", line 187, in fetch entries: Iterable[Collection] = self._source.fetch() ~~~~~~~~~~~~~~~~~~^^ File "/config/custom_components/waste_collection_schedule/waste_collection_schedule/source/eastleigh_gov_uk.py", line 39, in fetch r.raise_for_status() ~~~~~~~~~~~~~~~~~~^^ File "/usr/local/lib/python3.14/site-packages/curl_cffi/requests/models.py", line 241, in raise_for_status raise HTTPError(f"HTTP Error {self.status_code}: {self.reason}", 0, self) curl_cffi.requests.exceptions.HTTPError: HTTP Error 403:
```

### Source Configuration

```YAML
Default configuration.
```

### Checklist

CozyRocket:
Fixes #7424 — Eastleigh Borough Council fetches were failing with `HTTP Error 403`.

The source pinned `curl_cffi`'s impersonation to `"chrome124"`, which is now several major Chrome versions behind what `curl_cffi` currently ships (up to `chrome150`). Cloudflare's bot detection appears to have started flagging that stale, static TLS/HTTP2 fingerprint — the site itself is up, only the fetch was blocked.

Switched to the bare `"chrome"` target, which tracks curl_cffi's latest supported fingerprint instead of a pinned version that ages out. Same fix already applied elsewhere in this repo for the same reason: `south_ayrshire_gov_uk`, `southderbyshire_gov_uk`, `inverclyde_gov_uk`, `northumberland_gov_uk`.

**Caveat for the reviewer:** I couldn't reproduce a 403 locally — my requests with the *old* `chrome124` fingerprint succeeded too, so this is Cloudflare's detection being inconsistent/IP- or reputation-based rather than a hard block I could confirm is fixed end-to-end. The fix is the established, low-risk pattern for this exact symptom in this codebase; flagging in case the reporter still sees 403s after this merges, in which case a stronger measure (session warm-up, explicit headers) may be needed.

**Test plan**
- [x] `python test_sources.py -s eastleigh_gov_uk -l` against the live site — both TEST_CASES return real collection data
- [x] `ruff check` / `ruff format --check`
- [x] `pytest tests/test_source_components.py`


bbr111 (approved):
Thanks — approving as-is.

Verified against the live site: `test_sources.py -s eastleigh_gov_uk -l` returns 5 real entries for both TEST_CASES, `pytest tests/test_source_components.py` is 42 passed, and pre-commit (ruff, ruff format, codespell) is clean.

I can confirm your caveat: from here, curl_cffi 0.15.0 returns HTTP 200 with a byte-identical body for both `chrome124` and `chrome`, so the 403 in #7424 is IP/reputation-dependent rather than a hard fingerprint block. That doesn't change my view of the patch though — bare `chrome` tracks curl_cffi's newest supported fingerprint instead of ageing out, it's what 35 other sources in this repo already use, and it can't regress anywhere the pinned value worked.

If the reporter still sees 403s after this ships, we'll follow up with a stronger measure (session warm-up or explicit headers) on the issue.

Two notes for my own follow-up, nothing for you to do: this needs porting to release/3.0.0, which still pins chrome124, and there are ~15 other sources on stale pinned fingerprints that deserve a sweep.

PR #7434, opened 2026-09-16
Add source: Stadt Kufstein (stadt_kufstein_at)

### LINKED ISSUE #7422: [Source Request]:  Kufstein
### Municipality / Region

Kufstein.at

http://www.stadt.kufstein.at/system/web/kalender.aspx?bezirkonr=0&detailonr=218400983&typid=218400983&menuonr=218502408

### Example Address

Kinkstrasse 12

### Data Format

HTML webpage (dates visible on the page)

### Known Platform (if any)

Not sure / Unknown

### Additional Information

In Kufstein we have 6 Areas with different Dates. On webpage you can enter your adress and you see with area belongs to you

### Checklist

- [x] I have searched the [source list](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/README.md) and my provider is not already supported
- [x] I have provided a working example address (not my own) for testing
- [ ] I have tried to implement this myself (or with a free coding agent like Claude or Copilot, following [AGENTS.md](https://github.com/mampfes/hacs_waste_collection_schedule/blob/master/AGENTS.md))

CozyRocket:
Closes #7422

Kufstein (Tyrol, Austria) runs the same RiS-Kommunal (gem2go) platform already shared by 35 other AT sources, so this composes the existing `RiSKommunalRetriever`/`RiSKommunalParser` rather than adding anything to `service/`. The collection area (`typids`) is resolved per street + house number and is mandatory — omitting it merges all 6 of Kufstein's areas into one page of duplicated rows.

Closest template: `imst_at.py` (also Tyrol, also street + house-number → area lookup).

**Worth flagging for review:** the site is HTTP-only — `https://www.stadt.kufstein.at` presents a certificate for the wrong principal, so all URLs deliberately use `http://` (commented in the source so it doesn't get "fixed" later). Also `www.kufstein.at` without the `stadt.` prefix is an unrelated tourism site and 404s on this path.

**Test plan**
- [x] `ruff check` / `ruff format --check`
- [x] `mypy --ignore-missing-imports --explicit-package-bases`
- [x] `pytest tests/test_source_components.py`, `tests/test_declared_waste_types.py -k kufstein`, `tests/test_offline_fixtures.py -k kufstein`
- [x] Live `test_sources.py -s stadt_kufstein_at -l` — all 3 `TEST_CASES` (in 3 different collection areas, to actually exercise area resolution) return correct, non-empty, distinct schedules
- [x] Cassettes recorded per `TEST_CASES` entry (`python tests/record_fixtures.py stadt_kufstein_at`) and confirmed to pin the full request body (rejected under `-p tests.mutate_requests`)
- [x] Full `pytest -m "not live"` — 1508 passed, 4 skipped, no regressions


bbr111 (approved):
Thanks, this is a textbook pipeline source — nothing for me to change.

It composes the shared `RiSKommunalRetriever` / `RiSKommunalParser` / `ICSTransformer` with no source-local step and no hand-rolled retrieval, so the platform stays reusable. `WASTE_TYPES` matches what the cassettes actually produce, and the three recordings resolve three distinct `typids` (218400980 / 218400970 / 218400984), which means the per-house-number area lookup issue #7422 asks about is genuinely exercised rather than incidentally green. The recordings also pin their request bodies — `-p tests.mutate_requests` correctly fails all three cases.

I independently confirmed the `http://` decision: `https://www.stadt.kufstein.at` serves a certificate for a different principal (`no alternative certificate subject name matches target host name`), while the plain-HTTP calendar URL returns 200 with no redirect. Good call documenting that in the file so it does not get "fixed" later.

Checks on this branch: `tests/test_source_components.py` 42 passed; `test_new_architecture.py` + `test_declared_waste_types.py` + `test_offline_fixtures.py` 7557 passed, 8 skipped; `pre-commit` clean on all four files.

Approving as-is.

PR #7448, opened 2026-09-17
Config flow: collect a cascading_select one level per view, and migrate kiedysmieci_info onto it

kjanas:
## Summary

The problem: `config_flow` does not support lazy loading while user provides address by 4 dropdown lists. Currently users have to submit form 4 times to see correct values and find address.
Solution: For Poland/kiedysmieci_info flow, divide location setup into 4 steps - each step covers one dropdown populated with input from previous step. Final step shows all selected values.

Goal achieved: no free-text needed anymore, no guessing or spelling mistakes, dropdowns are being populated with entries served by source API, exactly the same lists as in the mobile apps and on the kiedysmieci.info web page.

This is the follow-up promised in #7419, and it comes in two commits: the config-flow change, which serves every `cascading_select` source, and the `kiedysmieci_info` migration that makes it reachable for this provider.

### Config flow: one level per view

A `cascading_select` lived entirely on the single argument form. Only the widest
level could be populated up front - every deeper level needs the choices above it
- so the rest rendered as free text and an incomplete submit, plus the validation
error it produced, was the only way to reveal the next dropdown. Measured on this
source, the first screen was:

```
calendar_title  text
voivodeship     dropdown
district        TEXT BOX
municipality    TEXT BOX
street          TEXT BOX
```

and is now four views of one dropdown each (16 / 9 / 1 / 311 options for the
address below), every one of them fetched for the levels already chosen.

A level whose `get_choices()` returns `[]` is skipped rather than shown empty, per
the `cascading_select` contract. A cascade that loses its provider mid-way falls
back to the ordinary form, keeping what was answered, rather than dead-ending.

### The closing view

The levels are dropped from the last form and shown as text. Re-offering them
there would let someone pick a different voivodeship on the final screen and
silently invalidate the district, municipality and street chosen underneath it;
they are merged back into the submission instead.

It uses a fixed step id, `cascade_confirm`, so its description can be the chosen
address alone rather than the generic "Configure your service provider" blurb and
the source's HOWTO, both of which answer questions the cascade has already
answered by that point. One step id means one translation serves every source
that reaches this view, rather than an entry per source.

### Translations in the diff

`translations/*.json` appears in the diff and this is not generated content. The
five files gain one hand-maintained base key, `cascade_confirm`, alongside the
existing `args` / `reconfigure` / `user` / ... entries. `update_docu_links.py`
only pops and rewrites keys prefixed `args_` / `reconfigure_`, so the new key is
preserved; I ran the generator to confirm it survives in all five languages, then
reverted its other output.

### kiedysmieci_info

Migrated to `BaseSource`, because the config flow reads `PARAMS` only from a
pipeline source and that declaration is what enables the wizard. The request flow
moved into `service/KiedySmieci.py`, which owns the proxy's response envelope, the
locations cascade behind `get_choices()`, and the walk that turns a failed fetch
into an error naming the argument at fault.

Waste types are kept verbatim rather than mapped onto canonical ones. Every label
this provider emits is Polish, `waste_types.SUPPORTED_LANGUAGES` has no `pl` to
resolve against, and mapping the five common streams would replace the label a
user reads today ("zmieszane") with an English one, renaming the calendars of
everyone already on this source. The types are built like `wt.preserved()` but
carry the icon the source picks for the stream, which the plain helper flattens to
`OTHER`'s. Happy to map them if you would rather have canonical types here.

Arguments and `TEST_CASES` are unchanged, so existing configurations keep working.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [x] Migration of an existing source to the 3.0.0 pipeline
- [ ] Documentation update
- [x] Other - config flow support for multi-step `cascading_select`

## Checklist

- [x] No generated files in diff - `translations/*.json` is in it for the one
      hand-maintained base key, see "Translations in the diff" above
- [x] `doc/source/<name>.md` created for new legacy sources (pipeline sources generate theirs) - n/a, and the HOWTO section it used to carry moved onto the class so the generator keeps it

- [x] **Cassette recorded.** `python tests/record_fixtures.py kiedysmieci_info`, committed under `tests/fixtures/kiedysmieci_info/` - the three TEST_CASES plus the `_choices.json` the recorder produced for the cascade.
- [x] **Composes shared components.** No `Retriever`/`Parser` subclass in the source module, no step override on `Source`, no module-level function issuing HTTP. `retrieve` / `parse` are `KiedySmieciRetriever()` / `KiedySmieciParser()` from the component.
- [x] **Anything the platform was missing was added to the shared component.** The envelope handling, the cascade and the argument-level diagnosis all live in `service/KiedySmieci.py`, so a second provider fronted by this proxy gets them.

`_raise_empty()` is overridden on the source. It is not a pipeline step - it is the
"nothing found" message - and the override exists because `BaseSource` blames the
first declared field, the voivodeship, when every level here was picked from a list
the API itself returned. The pair that can still be wrong together is the
municipality and the street.

## Testing

`python -m pytest tests/ -q` - 8122 passed, 9 skipped.

Live, against the provider:

```
Testing source kiedysmieci_info ...
  found 42 entries for Nadolany, podkarpackie, sanocki, Bukowsko
  found 18 entries for Kędzierz, podkarpackie, dębicki, Dębica
  found 19 entries for Parkowa, lubelskie, zamojski, Szczebrzeszyn
```

Unchanged from #7419, as are the error paths: a wrong voivodeship, district,
municipality or street each still report suggestions for their own level, and
lookups stay case-insensitive.

Five new tests in `tests/test_config_flow.py` cover the wizard against a synthetic
cascading source: one view per level, a level with no options skipped, an
unreachable provider falling back, the levels kept when it breaks mid-way, the
settled levels merged back into the submission, and the confirmation step present
in every shipped language.

Also installed on a live Home Assistant 2026.9.2 and walked through the UI: four
dropdown views, then a closing view showing the four values and the calendar title
field, with the entry created from it.


superrob:
This looks at a glance like functionality i was searching for myself.
For the source affaldonline_dk it would be possible to allow the user to input their address and have the source find the correct "values" for them, instead of having the user be tied into manually finding the correct "values" themselves by looking into network traffic in their browser.

It seems like a config flow like this, could allow the source to act on information and help the user find the "values" needed. The current config flow, would only allow for the user to input their address, and then on each reload of the integration have to "search" for the "values", instead of effectively caching them.

kjanas:
@superrob 
Yes, that is exactly the case this is meant to cover, and `affaldonline_dk` looks like a good fit.

The mechanism is `config_params.cascading_select(...)` plus a `get_choices()` classmethod on the source. The config flow asks for one level per view and calls `get_choices(field, selections)` with everything chosen so far, so the source decides at config time what the next dropdown contains. Two details are the ones you are after:
- an option may be a `(label, value)` pair, so the dropdown can show `Nørregade 1, 8382 Hinnerup` while the config entry stores the opaque `Nørregade|1||||8382|Hinnerup|6443|108156|0`. The lookup runs once, during setup; `fetch()` afterwards is the single call it is today. That is the caching you describe - it falls out of the contract rather than needing anything new.
- a level whose `get_choices()` returns `[]` for the current selections is skipped rather than shown empty, so the municipalities without a `client` (Sorø) do not have to dead-end the wizard.

For your source that would be roughly `municipality -> street -> house number` in one `cascading_select`, with `get_choices()` replaying the same requests the kalender page makes. `abfall_io.py` and `app_abfallplus_de.py` are existing sources that already do this, and `kiedysmieci_info` in this PR is a four-level one.

One caveat worth checking before you plan on it: this is a cascade of dropdowns, not a search box. It needs each level to be *enumerable* - the API must be able to return the streets for a municipality, and the house numbers for a street. If the only thing behind `showInfo.php` is prefix autocomplete over typed text, a pure cascade will not express it; `multi_value_lookup` (user types an address, the source resolves it into several stored params) is the closer primitive, and the two can be combined. Worth poking at the endpoint to see which one you have.

Note also that `PARAMS` is only read from a `BaseSource` pipeline source, so `affaldonline_dk` would need the 3.0.0 migration first - that is what the second commit here does for `kiedysmieci_info`, if it helps as a reference.

bbr111:
nice idea. will need to take a closer look.

bbr111:
Interim review notes (not a final review yet).

What I've checked so far:
- No generated files in the diff. The `cascade_confirm` key added to translations/*.json is a base key; update_docu_links.py only rebuilds `args_*` / `reconfigure_*`, so it survives.
- kiedysmieci_info follows the pipeline and reuse rules: retriever + parser live in service/KiedySmieci.py, the source module stays declarative, nothing was added to the exception/backlog lists.
- Cassettes: one per TEST_CASES entry plus the choices walk. They fail under `-p tests.mutate_requests`, so they pin the request payloads.
- Full suite green (8123 passed, 9 skipped), pre-commit clean.

One bug in the new config-flow code:
The closing `cascade_confirm` view drops the cascade levels from the form. A validation error naming one of those levels has no field to attach to, so it renders nowhere. This source's own `_raise_empty()` blames ("municipality", "street"), so the user would resubmit a form that never says what is wrong. `validate_location()` can hit the same case if the provider's lists change between the wizard reading them and the fetch.
Proposed fix: on such an error, re-open the levels (still as dropdowns) and go back to `args_<id>`, whose description carries {invalid_arg_message}. I have a fix and a regression test ready and will push them if maintainer edits are enabled.

Follow-ups, not blockers:
- service/KiedySmieci.py uses module-level `requests.get` instead of `source.session`.
- The bare `raise Exception(...)` for a not-ok envelope could be a narrower type.

Still open:
The config-flow change is generic, so it also changes the flow for the other cascading_select sources (abfall_io, app_abfallplus_de, aw_harburg_de, beachwood_oh_us, stadt_kerpen_de). I have not yet run them through the new flow and will do that before a final verdict.
Separately, once this lands, gemeinde24_at and the wizard/ helpers may be able to move onto cascading_select. That would be a follow-up issue, not part of this PR.

kjanas:
Thanks — the bug is real and I've fixed it, along with both follow-ups. Details below; the last section is a finding from your still-open item that changed the config-flow code again, so it's worth a look before you run the other sources.

Maintainer edits are on. I've written a fix anyway — if yours differs, push it over mine.

## The cascade_confirm bug

Confirmed, and the cause is exactly as you describe: the levels are not in that form's schema, so an error naming one binds to a field that isn't there and renders nowhere.

I took the shape you proposed. On an error whose key is one of the omitted levels, the cascade stops being settled and the flow returns to `args_<id>`:

```python
if not errors.keys().isdisjoint(omit_fields):
    self._cascade_complete = False
    omit_fields = self._cascade_omitted()
    step_id = f"args_{self._id}"
    description_placeholders = self._get_description_placeholders(self._id)
```

Re-opening is the half that matters most: the user needs a field to *change* the wrong answer, not only to read about it. `_finish_cascade` already parks the picks in `_extra_info_default_params`, so the re-opened form comes back populated, and `__get_dependent_options` repopulates each level as a dropdown rather than free text.

One small correction: `args_<id>`'s description doesn't carry `{invalid_arg_message}` — the `invalid_arg` *error* string does (`"Argument is invalid: \"{invalid_arg_message}\""`). The placeholder still has to survive the step switch, which is what the test asserts.

Two regression tests. I checked the first actually catches it — reverting the fix gives `assert 'cascade_confirm' == 'args_cascade_stale'`. The second pins the other direction: a `base` / `fetch_error` renders where it is and must *not* throw the settled levels back at the user.

Both of your paths are covered by it: `_raise_empty()`'s ("municipality", "street"), and `validate_location()` when the lists move between the wizard reading them and the fetch — that one lands as `SourceArgumentNotFoundWithSuggestions`, so the re-opened level comes back as its suggestions dropdown.

## Follow-ups

**`source.session`.** Done. `request()` takes an optional session and the retrieve path passes `source.session`; `_raise_empty()` passes `self.session` too, so the fetch and the cascade walk that diagnoses its failure share one connection. It stays optional because `choices()` is reached through the `cascading_select` *classmethod* contract, where there is no instance to take a session from — that call keeps plain `requests`.

The cassettes replay unchanged across the stack switch (the matcher is method + URL + body, and both stacks route through the patched `Session.request`), and they still fail under `-p tests.mutate_requests`.

**Bare `raise Exception`.** Now a module-level `KiedySmieciError`, following `IntraMapsError` / `ArcGisError` / `PoziError`. Narrow on purpose: it means the envelope came back `ok: false`, which is neither a transport error nor a wrong address (this API doesn't report one).

## Your still-open item — I ran it, and it found a gap in my own change

Since the config-flow change is generic I drove the wizard against the real sources, replaying each one's recorded `get_choices` walk (`tests/fixtures/<source>/_choices.json`).

`kiedysmieci_info`, `aw_harburg_de` and `beachwood_oh_us` walk correctly. **`abfall_io` and `app_abfallplus_de` did not.** Both key every lookup on a param that is not a cascade level — the service `key` and the `app_id` — and both get it from the region the user picked, not from the cascade. The wizard was passing `get_choices` only the cascade selections, so every level returned `[]`, every level was skipped as "not applicable", and the wizard silently handed back the old single form.

Not a regression (those users saw exactly what they see on master), but it means the change delivered nothing for 41 + 145 regions, which is most of what it claims to serve. Fixed by passing the non-level pre-filled params through as context, and pre-selecting a level a region already answered so picking the region still means something.

Measured against the providers:

```
abfall_io         f_id_kommune  {}                       -> 0 options
abfall_io         f_id_kommune  {key: ...}               -> 25 options
app_abfallplus_de bundesland    {app_id: ..., city: ...} -> 0   (not applicable)
app_abfallplus_de landkreis     {app_id: ..., city: ...} -> 0   (not applicable)
app_abfallplus_de city          {app_id: ..., city: ...} -> 6 options
```

app_abfallplus_de's two empty upper levels are the contract working as intended: that app fixes them, so the wizard steps over both and opens on the city.

Three tests cover it, one per shape: the recorded one-level-per-view walk for the three sources that need no context, abfall_io with its key, and abfall_io *without* it — the last asserts the wizard asks nothing, which is what it did before and why the context has to be passed.

## Two pre-existing things I noticed, both in files this PR doesn't touch

- `stadt_kerpen_de.get_choices` returns 0 options live, despite its `PINNED_PARAMS` carrying the key and kommune. So its cascade never populates on the old `dependent_options` path either — the wizard just makes it more obvious by having nothing to open with.
- `abfall_io` emits `('Bitte auswählen...', '0')` as a real dropdown option. It was easy to ignore as one field among five; as a whole view of its own it reads as a choice.

Separate issues rather than scope for this PR, unless you'd rather I fold them in.

## Testing

`python -m pytest tests/ -q` — 8126 passed, 9 skipped (8122 before these changes; +4 new tests). `ruff check` / `ruff format` clean on the changed files.

bbr111 (changes requested):
Thanks for the quick turnaround - I re-ran c1292af8 and the earlier findings are confirmed fixed: the full suite is green (8160 passed, 9 skipped), pre-commit is clean, and the cassettes still fail under `-p tests.mutate_requests`, so they pin the payloads. The abfall_io / app_abfallplus_de context fix is a good catch.

Before this can merge, three things in the closing view (`cascade_confirm`). I read the code for each and they reproduce from it:

1. Skipped levels come back as free text.
   `_cascade_omitted()` hides only the levels in `_cascade_selections`. A level with no options is skipped and never recorded there, so it stays in the closing form's schema as an ordinary input. For app_abfallplus_de that means `bundesland` / `landkreis`, which the wizard deliberately stepped over, reappear as empty text fields on the last screen. A level the cascade visited should count as settled whether it was answered or skipped: hide it too, and merge back whatever value the region pre-filled for it (or nothing) when the form is submitted.

2. The summary shows internal values, not names.
   `_cascade_summary` prints `self._cascade_selections[field]`, which is the option's value (`opt[1]`). For abfall_io that is an ID rather than the municipality's name, so the "what you picked" list is not readable. Keep the option label alongside the value when a level is answered and show the label.

3. Extra fields have no labels in the closing step.
   `cascade_confirm.data` in the translations only defines `calendar_title`. Any other field on that form gets no label; stadt_kerpen_de's waste-type selection is the case that shows it. The closing step needs to label the non-cascade fields the way `args_<id>` does. Note that the `args_*` sections are generated, so this has to come from a hand-maintained key or from passing the labels in, not from editing those.

I'd like these three fixed here since they are what the user sees on the last screen for most of the regions this change is meant to serve (abfall_io and app_abfallplus_de). A test for each would be enough: a skipped level is not offered on the closing form, the summary shows labels, and an extra field has a label.

Not blocking, better as separate issues:
- stadt_kerpen_de.get_choices returns 0 options live, and abfall_io offers `('Bitte auswählen...', '0')` as a real option (both from your notes).
- An HTTP error such as 503 is reported as a provider error without the status. I have not verified whether this predates the PR.
- affaldonline_dk needs a search-by-address flow, which this mechanism does not cover.

Requesting changes for the three items above; happy to look again as soon as they are in.

kjanas:
All three are fixed in 522f9796, and the branch is rebased onto 749aa48c (the only conflict was `tests/test_config_flow.py`, where your default-sensors block and my cascade block both append to the end of the file).

**1. Skipped levels came back as free text.** Levels are now settled by having been *visited*, answered or skipped, rather than by appearing in `_cascade_selections`. A skipped one contributes whatever the region pre-filled for it, or nothing — the levels are individually optional, so a level that does not apply is simply absent from the submission.

There is a trap underneath this one worth knowing about. Hiding every visited level breaks the case where *all* of them return `[]`, which is what a source looks like before its lookup key is known — abfall_io without `key`, the exact case the context fix was about. Hiding those leaves a form with no address fields at all. So a cascade that answered nothing now stays incomplete and the ordinary form keeps every level editable. `test_an_inert_cascade_leaves_every_level_editable` pins it; it is the one new test that passes on the old code, because it guards a regression the fix itself would otherwise have introduced.

**2. Summary showed internal values.** The label of the option chosen at each level is kept alongside the value, and the summary shows the label. The value is still what reaches the entry. Skipped levels are left out rather than listed blank, since a blank line reads as something the user failed to fill in.

**3. Unlabelled fields on the closing step.** `app_id`, `f_abfallarten` and `key` are labelled in all five languages, reusing the exact strings the generated `args_*` sections already use so the two screens agree. `f_abfallarten` also carries its `data_description`.

Rather than leave that as a list that rots, `test_the_closing_step_labels_every_field_it_can_show` computes the set from the sources themselves — every non-cascade param of every `cascading_select` source — and fails with a message naming the missing fields and saying the labels have to be hand-maintained because `args_*` is generated. A new cascading source that adds a param cannot quietly lose its label.

Four new tests beyond that one, and I checked each fails on the pre-fix tree: a skipped level is not offered on the closing form and contributes nothing, a skipped level still carries a region's pre-filled value, and the summary shows labels rather than stored ids.

`python -m pytest tests/ -q` — 8176 passed, 9 skipped. pre-commit clean.

On the non-blocking items: I will open separate issues for `stadt_kerpen_de.get_choices` and the `Bitte auswählen...` option. The 503 one predates this PR — `request()` parses the envelope before looking at the status, deliberately, because this API reports its errors in the body with a 4xx/5xx alongside, but it does mean a status-only failure loses the code; the pre-migration source did the same. Worth a separate fix that keeps the envelope path and adds the status when there is no message. affaldonline_dk is superrob's thread and wants `multi_value_lookup` rather than a cascade, so I would keep it out of this one.

bbr111 (approved):
Thanks - all three are fixed in 522f9796 and I re-ran it on a fresh checkout: the full suite is green (8176 passed, 9 skipped), pre-commit is clean on the changed files, and the kiedysmieci cassettes still fail under `-p tests.mutate_requests`, so they keep pinning the payloads.

I went through the config_flow.py diff for each point:
- Skipped levels: settled by being visited rather than by being answered, so they no longer come back as free text, and a region's pre-filled value is still carried. The exception you added for a cascade that answered nothing (abfall_io before its `key` is known) is the right call; without it the closing form would have no address fields at all.
- Summary: it shows the chosen option's label, the stored value is unchanged.
- Labels: `app_id`, `f_abfallarten` and `key` are labelled in all five languages, and the test that computes the set from the cascading_select sources means a new source can't quietly lose a label.

I did not drive the wizard live against abfall_io / app_abfallplus_de myself; I'm relying on your recorded-walk tests for those.

Agreed on keeping these out of this PR, as separate issues:
- stadt_kerpen_de.get_choices returning 0 options live, and abfall_io's `Bitte auswählen...` option showing up as a real choice.
- The 503 case: `request()` parses the envelope before the status, so a status-only failure loses its code. A follow-up that keeps the envelope path and adds the status when there is no message sounds right.
- affaldonline_dk, which wants `multi_value_lookup` rather than a cascade.

Approving. It targets release/3.0.0, so it goes into the 3.0.0 batch.

PR #7450, opened 2026-09-17
Make device removalbe

smoki3:
## Summary

This makes a device removable. It necessary, because if you do a reconfiguration every time you change the URL for example a new device is created. And you end them with many devices which have no function

## Type of change

- [ ] New source
- [ x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


bbr111:
@smoki3 how do you code this?

if use vscode or other tools you can run this checks local before commit.

bbr111:
@smoki3 
do you need help with this?

smoki3:
I hoped that I dont need to setup the build environment. I did it quick n dirty just in github. But I will setup one

bbr111:
no problem, i can help out if you want.

smoki3:
on my local now its working

bbr111:
if just checked in my installation.

Testcase:

setup 2 WCS integration.
got 2 devices with sensors.

if i delete on of the integration
the device is also gone.

sure we need this?

bbr111 (changes requested):
Thanks for tackling this — the orphaned-device problem is real. When a user reconfigures an entry with different arguments (e.g. a new address), the device identifier is recalculated from the new source args, so the old device is left behind in the registry with no way to remove it through the UI. The PR correctly identifies this gap.

The one issue is that `async_remove_config_entry_device` currently returns `True` unconditionally, which also allows deleting the *active* device of a live entry. When that happens HA drops the device and its entity-registry entries, while the coordinator still advertises the same identifier and recreates the device — so the user loses all their entity customisations.

Could you add a guard so only stale devices can be removed — i.e. devices whose identifier no longer matches the live coordinator?

```python
async def async_remove_config_entry_device(
    hass: HomeAssistant,
    config_entry: ConfigEntry,
    device_entry: dr.DeviceEntry,
) -> bool:
    coordinator = hass.data.get(const.DOMAIN, {}).get(config_entry.entry_id)
    if coordinator is None:
        return True
    active_identifier = (const.DOMAIN, str(coordinator.shell.unique_id))
    return active_identifier not in device_entry.identifiers
```

smoki3:
Best think would be if you can fix the recalculating of the device identifier. Because the reconfigure also leads to an other issue. As soon a new device is created and the old one is still existing all your entities get new IDs. Mean afterwards you need to delete the old device and rename all the entities

bbr111:
> Best think would be if you can fix the recalculating of the device identifier. Because the reconfigure also leads to an other issue. As soon a new device is created and the old one is still existing all your entities get new IDs. Mean afterwards you need to delete the old device and rename all the entities

Can you explain how I can reproduce it? i didn't get this problem. if i remove the integration-configuration my device was deleted. wenn setup a new, it gets a new device with the same names.

smoki3:
https://github.com/user-attachments/assets/ad61f210-5bec-44c7-83ac-3fdaa0889536

Made a short video: changed the URL of the device. Then you see a new device is created and the new entities get the suffix "_2" because they are still available on the old device

bbr111:
@smoki3 are you on discord? maybe we can there discuss possible solutions for this problem and report it back here.

bbr111:
we found a different solution PR #7462

PR #7462, opened 2026-09-18
Do not create new ids on reconfigure

smoki3:
## Summary

Use existing unique IDs for reconfiguring 

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


bbr111 (changes requested):
Thanks for tackling this — the diagnosis is spot on. `shell.unique_id` is derived from the source args, so reconfiguring to a new address rebuilt every entity's unique_id and orphaned the originals. Re-basing entity identity on `config_entry.entry_id` is the right target state and matches HA's own guidance.

The problem is the blast radius. This doesn't only affect entries that get reconfigured: it changes the unique_id of every sensor and calendar, plus the device identifier, on **every existing UI-configured install**, the moment users upgrade. HA won't recognise the old registry entries, so everyone gets duplicate entities (`sensor.x_2`), lost history and statistics, broken automations and dashboards, and a stale orphaned device. That's the same failure mode the PR is fixing, just relocated from "on reconfigure" to "on upgrade, for everyone".

Could you add an entity- and device-registry migration before this lands? The repo already has the hook — `async_migrate_entry` in `init_ui.py`, keyed on `CONFIG_VERSION` / `CONFIG_MINOR_VERSION` (currently 2 / 13). The old ids are fully reconstructible from data still on the entry, so it's mechanical:

```python
# behind a bump to CONFIG_MINOR_VERSION = 14
old_base = calc_unique_source_id(new_data["name"], new_data["args"])

@callback
def _migrate_unique_id(entry: er.RegistryEntry) -> dict | None:
    if entry.unique_id.startswith(old_base):
        return {"new_unique_id": entry.unique_id.replace(old_base, config_entry.entry_id, 1)}
    return None

await er.async_migrate_entries(hass, config_entry.entry_id, _migrate_unique_id)

dev_reg = dr.async_get(hass)
if device := dev_reg.async_get_device(identifiers={(DOMAIN, old_base)}):
    dev_reg.async_update_device(device.id, new_identifiers={(DOMAIN, config_entry.entry_id)})
```

Two smaller things while you're in there:

1. `calc_unique_calendar_id` (`calendar.py:202`) now has no callers. Please don't delete it — it's exactly what the migration needs to compute the old calendar ids, and it'd be nice to keep the YAML fallback branch at line 141 calling it so the two spellings of the id can't drift apart.
2. Entity identity now depends on `coordinator.config_entry`, but `WCSCoordinator` never passes `config_entry=` to `super().__init__()`, so it relies on HA's implicit `current_entry` contextvar and is typed `ConfigEntry | None`. It works today, but it's deprecated and `calendar.py:141` only guards on `coordinator`, not on `coordinator.config_entry`. Worth passing the entry explicitly from `init_ui.py`.

The `config_flow.py` half looks right to me: moving `async_set_unique_id()` into the non-reconfigure branch is the correct HA pattern, and dropping `unique_id=config_entry.unique_id` from `async_update_reload_and_abort` is equivalent since an omitted value leaves the entry untouched. As a possible follow-up (not this PR): once entity ids no longer track the args, it becomes safe to refresh the *entry's* unique_id on reconfigure and collision-check it against other entries — right now a reconfigured entry keeps its original key, so a duplicate entry for the same address wouldn't be caught.

Finally, this is a breaking change under `doc/versioning.md` (entity-ID + device-identifier change). `release/3.0.0` is the right target, but it needs its own CHANGELOG entry — ideally one that says the migration makes it transparent.

Everything currently passes locally (`tests/test_source_components.py`, `tests/test_config_flow.py`, full `pre-commit`), so it's purely the migration that's blocking.

smoki3:
I used some copilot proposal. Can you check this?

bbr111 (approved):
Thanks for the follow-up — the migration is the right shape now, and re-keying both registries means existing installs carry over transparently. I've pushed a few fixes on top:

1. **`async_update_device()` would have raised.** The device half passed `identifiers=`, which isn't a parameter of `DeviceRegistry.async_update_device`. The keyword is `new_identifiers`. As written, every existing UI-configured install would have hit `TypeError` during migration and landed in migration-error, which is the failure this PR exists to prevent. Switched to `new_identifiers={(DOMAIN, entry_id)}`, and looked the device up with `dr.async_get_device(identifiers=...)` rather than scanning the registry.
2. **Entity re-keying now goes through `er.async_migrate_entries`.** The loop mutated `entity_registry.entities` while iterating it, and walked every entity in the instance to filter by `config_entry_id`. `_migrate_registry_unique_ids` is now async and awaited.
3. **`_get_source_unique_id` calls `calc_unique_source_id`** instead of re-spelling `name + str(sorted(args.items()))`. That format is owned by `source_shell.py`.
4. **`calc_unique_calendar_id` is back in use.** It now takes an optional `base_id`, and both branches (UI = entry id, YAML = `shell.unique_id`) go through it, so the platform's id and the migration's rewrite can't drift.
5. **`WCSCoordinator` takes `config_entry` explicitly.** It's passed from `init_ui.py` and assigned after `super().__init__()`, because the `config_entry=` kwarg on `DataUpdateCoordinator` only exists from HA 2024.11 and we still declare `homeassistant>=2024.4.0`.

`tests/test_source_components.py` (42 passed), `tests/test_config_flow.py` and `tests/test_fetch_retry.py` (14 passed) and a full `pre-commit run --all-files` are green.

Two notes, neither blocking:

- **CHANGELOG:** this repo writes `CHANGELOG.md` at release time, so I've left it out of the branch and will fold a "Fixed" entry into the next 3.0.0 beta notes.
- **Manual verification:** we have no in-repo HA test harness, so the migration path isn't covered automatically. Before this ships in a beta I'd like a smoke test on a real install with an existing v2 entry — no `_2` duplicates, sensor history survives, the device keeps its area and custom name, and a subsequent reconfigure to a different address reuses the same entities.

Approving on that basis. Good catch on the root cause, and thanks for pushing the migration through.

PR #7465, opened 2026-09-18
fix(publidata_fr): merge schedules per garbage type instead of overwriting them

klsx0:
## Summary

`_sanitize_response()` kept only the last "single"-sectorized hit per `garbage_type`, overwriting any earlier one instead of merging. Publidata sometimes returns two such hits for the same type — one schedule currently in effect, another taking over at a future date — and whichever came later in the API response won, silently dropping the other. For some addresses this made near-term collection dates for a given waste type disappear entirely from the calendar.

Found while debugging a live install (Agglo d'Agen, instance_id 1264): "Emballages" had a schedule valid through 2027-01-03 and one starting 2027-01-04, and the future one clobbered the current one — the calendar showed zero upcoming Emballages collections.

## Changes

- Append schedules per `garbage_type` instead of replacing them.

## Test plan

- [x] `ruff check` / `ruff format` — clean
- [x] `python -m pytest tests/test_source_components.py -q` — 42 passed
- [x] `python test_sources.py -s publidata_fr -l` — CASGBS, Le Pecq goes from 652 to 794 entries (recovered dates, same waste types, no duplicates); the other 7 test cases unchanged


bbr111 (approved):
The change correctly accumulates schedules from multiple single-sectorized hits sharing the same garbage type, preventing a later API hit from discarding an earlier active schedule. The fallback value is also corrected from an empty mapping to an empty list, matching the expected schedules shape. The PR changes no generated files, and the existing source contract, supported language keys, canonical icons, lowercase country code, live upstream integration, and test cases remain intact. All reported CI checks pass. Approved.

PR #7466, opened 2026-09-19
update sunderland form scraper

gvasilakiss:
## Summary

Fix the Sunderland City Council bin collection source to work with the council's current `BINCOLLECTIONCHECKERNEWV3` postcode/address lookup flow.

The source now:

- Submits the postcode through Sunderland City Council's current bin collection form.
- Resolves the selected address and UPRN from the postcode results.
- Submits the selected address/UPRN to retrieve collection data.
- Decodes the returned `BINCOLLECTIONCHECKERNEWV3FormData`.
- Extracts future collection dates from `DATARETURNED.result.schedules[].futureJobs`.
- Returns collection dates as `datetime.date` values expected by the source tests.

Tested using real, publicly accessible Sunderland addresses.

## Type of change

- [ ] New source
- [x] Bug fix / source fix
- [ ] Documentation update
- [ ] Other

## Checklist


bbr111 (changes requested):
Thanks for updating the Sunderland scraper to the council’s new form flow. This is close, but it cannot be merged yet because `fetch()` ignores the configured `address` and always selects the first property returned for a postcode. That can silently return another household’s collection schedule.

Please match the configured address against the returned options and raise `SourceArgumentNotFoundWithSuggestions` when it cannot be found. The no-address result should likewise use `SourceArgumentNotFound` for the postcode. The serialized-data decoder should catch only the expected Base64/Unicode exceptions, and terminal redirect responses should call `raise_for_status()`.

The source should also retain its existing canonical icons for known bin types. Finally, Ruff formatting is required; both current CI validation jobs fail because Ruff reformats the `__init__` signature.

gvasilakiss:
@bbr111 these have been updated now

bbr111 (approved):
Thanks for working through the previous round — the address matching, the argument-specific errors and the narrowed decode exceptions all look good now, and Ruff is clean.

I've pushed a couple of small tweaks rather than send this back again:

- Restored the canonical icons. The rewrite dropped `ICON_MAP` and the `Icons` import, so collections were coming through with no icon at all, which is a regression against the current source. Since the new flow labels collections from the API's free-text `binName` rather than the rendered page labels, I kept the three known labels and added an ordered keyword fallback, so an unrecognised name just falls back to no icon (the behaviour you had).
- `_follow_redirects` now calls `raise_for_status()` on the terminal response, and the redirect chain is capped at 10 hops instead of `while True` — an unbounded server-driven loop would hang the HA update thread.
- The `SourceArgumentNotFoundWithSuggestions` calls that passed `""` as the value now pass the real postcode or address, so the HA UI shows something useful.

Two things I couldn't check myself, if you could confirm in the thread:

1. Could you paste the actual `schedules[].binName` strings you saw? My `ICON_MAP` keys are carried over from the old rendered-page labels and may not match; the keyword fallback covers it either way, but I'd rather pin the real values.
2. `status != "IN SERVICE"` is an exact-string filter, and a non-match silently produces zero collections where the old code raised. Worth keeping in mind if empty-calendar reports show up.

Note also that the waste-type labels now come from the API rather than the three fixed strings, so existing users may see their labels change — I'll call that out in the CHANGELOG.

The provider Cloudflare-challenges my IP, so I couldn't run the TEST_CASES live; I'm trusting your report that they pass against real addresses. `pytest` (160 passed) and `pre-commit` are green. Approving.

gvasilakiss:
Confirmed from the live `test_sources.py -s sunderland_gov_uk -l` run against all three TEST_CASES: the API-backed collection labels were `General Waste 240L`, `Recycling 240L`, and `Garden Waste 240L`. These are the `schedules[].binName` values surfaced as collection types, so the legacy `ICON_MAP` keys do not match them directly; the keyword fallback is what maps them to the canonical icons.

The status handling is also as noted: only the exact normalized value `IN SERVICE` is included, and any other status is skipped, which can result in an empty calendar without an explicit error. The live cases passed with 4, 4, and 6 entries respectively. Cloudflare did not block the runner here; the structural pytest suite also passed. I will call out the API-provided label change in the changelog.

PR #7468, opened 2026-09-19
fix(sjobo_se): read day from its own div when the cell carries a week label

Darkslayer3324j:
## Summary

`sjobo_se` fails for every address: `ValueError: invalid literal for int() with base 10: '\xa0v.25'`.

The first day of each week carries a week-number label in the same cell:

```html
<td class="styleDayHit" colspan="2">
  <div style="float:left;...">v.2</div>
  <div class="styleInteIdag">5</div>
</td>
```

`cell.text` joins the two, giving `v.25` for Monday the 5th, so `int(day)` raises and the whole fetch fails. Since every month has week-start days with collections, the source never returns data.

The fix reads the day from the div that holds only digits, falling back to the old `cell.text`.

## Type of change

- [x] Bug fix / source fix

## Checklist

- [x] `python -m pytest tests/test_source_components.py -q` passes (sjobo_se)
- [x] TEST_CASES unchanged (existing public address)

Verified live: `test_sources.py -s sjobo_se` went from failing to 234 entries for the existing test case. Added `tests/test_sjobo_se.py` (offline HTML fixture); both tests fail on master and pass with the fix.

Found by running every source's TEST_CASES against the live sites. No existing issue or PR for this. Drafted with AI assistance (Claude Code) as this repo's AGENTS.md invites; I reproduced the failure and ran the checks above.


bbr111 (approved):
Thanks for this -- nice catch, and a well-written diagnosis.

I reproduced it against the live provider: the first cell of each week really is `<div>&nbsp;v.2</div><div class="styleInteIdag">5</div>`, so `cell.text` gives `'\xa0v.25'` and the whole fetch aborted for every address. Your fix also covers the `style="styleDayHit"` workaround cells, which carry the same inner structure. `test_sources.py -s sjobo_se` now returns 234 entries where it previously raised.

One small tweak from me: I removed `tests/test_sjobo_se.py`. Per-source unit-test files aren't this project's coverage model -- `TEST_CASES` plus a live `test_sources.py -s <name> -l` run is what we check, and with ~600 sources a file per source doesn't scale (see CLAUDE.md, "What NOT to do"). Worth knowing that the file was never actually running either: root `pytest.ini` pins `python_files = test_source_components.py test_fetch_retry.py test_ics_timezone.py`, so anything else under `tests/` is skipped by both `pytest tests/` and CI. Your existing `TEST_CASES` entry is what caught this bug to begin with, so it's already the right regression guard.

The source change is untouched. `pytest tests/test_source_components.py` (42 passed) and `pre-commit run --all-files` are green. Approving.

PR #7469, opened 2026-09-19
fix(lindau_ch): read plain date and name fields now that *-sort fields are obfuscated

Darkslayer3324j:
## Summary

`lindau_ch` fails for every address: `ValueError: Invalid isoformat string: '#1713171f...'`.

lindau.ch now returns the `*-sort` fields of its table data as opaque hex strings (`"_anlassDate-sort": "#1713171f05..."`, `"name-sort": "#382a2e3e..."`). The source parsed `_anlassDate-sort` with `datetime.fromisoformat` and used `name-sort` as the icon key, so the fetch always raised.

The plain fields carry the same data and are unchanged:

- date: `_anlassDate` contains `dd.mm.yyyy` (first match, so entries with a time like `23.09.2026<br>7.00 Uhr` still work)
- icon: the accent-stripped, lower-cased `name`, matched against the existing `ICON_MAP` keys (`grungut` now matches `Biogene Abfälle (Grüngut)`)

## Type of change

- [x] Bug fix / source fix

## Checklist

- [x] TEST_CASES unchanged

Verified live: `test_sources.py -s lindau_ch` went from failing to 33 entries for the existing `Tagelswangen` case. Added `tests/test_lindau_ch.py` (offline fixture); all 3 tests fail on master and pass with the fix.

Found by running every source's TEST_CASES against the live sites; no existing issue or PR. Drafted with AI assistance (Claude Code) as this repo's AGENTS.md invites; I reproduced the failure and ran the checks above.


bbr111 (approved):
Thanks for tracking this down — I reproduced it against the live site and confirmed lindau.ch now serves `_anlassDate-sort` / `name-sort` as opaque hex, so `datetime.fromisoformat` was raising on every entry. Reading the plain `_anlassDate` / `name` fields is the right fix, and both TEST_CASES pass live again (33 entries for Tagelswangen, 32 for Grafstal).

I made three small tweaks while merging:

- Swapped the raw `"mdi:trash-can"` fallback for `Icons.GENERAL_WASTE` — same value, but keeps the module free of raw `mdi:` strings (see #2813).
- Added `"sonderabfall": Icons.HAZARDOUS`; it's one of the six types the feed actually emits and was the last one falling back to the generic bin icon.
- Removed `tests/test_lindau_ch.py`. It's nothing against the test itself, but root `pytest.ini` pins `python_files` to a three-file allowlist, so that file is never collected by `pytest tests/` or by CI — it looks like coverage without being any. With ~600 sources we rely on `TEST_CASES` plus a live `test_sources.py -s <name> -l` run instead; that's documented in CLAUDE.md under "What NOT to do".

One non-blocking note for the future: the date regex requires zero-padded `dd.mm.yyyy`, and an unpadded date would be silently dropped by the `continue`. All 39 current live entries are padded, so it's theoretical — `\d{1,2}\.\d{1,2}\.\d{4}` would just make it bulletproof.

Approving with those tweaks pushed to your branch. Thanks again!

PR #7470, opened 2026-09-19
fix(reso_gmbh_de): strip the duplicated time suffix the feed adds to timestamps

Darkslayer3324j:
## Summary

`reso_gmbh_de` fails for every address: `Cannot access 'dt' on broken property 'CREATED' (expected 'vDDDTypes'): Expected datetime, date, or time. Got: '20260101T000000ZT000000Z'`.

The provider's feed appends a second time component to its timestamp properties:

```
CREATED:20260101T000000ZT000000Z
LAST-MODIFIED:20260101T000000ZT000000Z
```

Current `icalendar` no longer tolerates that value when `icalevents` reads the property. The source now removes the duplicated `T000000Z` suffix from `CREATED` / `LAST-MODIFIED` / `DTSTAMP` lines before handing the feed to the ICS service. Well-formed feeds are untouched.

## Type of change

- [x] Bug fix / source fix

## Checklist

- [x] TEST_CASES unchanged

Verified live: `test_sources.py -s reso_gmbh_de` went from failing to 23 entries for the existing `Reichelsheim Kerngemeinde` case. Added `tests/test_reso_gmbh_de.py` (offline fixture with the malformed feed shape); it fails on master and passes with the fix.

Found by running every source's TEST_CASES against the live sites; no existing issue or PR. Drafted with AI assistance (Claude Code) as this repo's AGENTS.md invites; I reproduced the failure and ran the checks above.


bbr111 (approved):
Thanks for this — the diagnosis is spot on, and I reproduced the exact `BrokenCalendarProperty` you describe. I've made a couple of changes before merging.

The fix has moved from the source module into the shared ICS service (`service/ICS.py`), next to the three sanitisers already there for EXDATE, truncated DTSTART/DTEND and TZID-on-VALUE=DATE lines. The malformed value is rejected inside `ICS.convert()`, so that's where it belongs: your version left `convert_events()` and every other ICS-based source exposed to the identical defect, and any other provider hitting it would need the same patch copied again. With the service fixed, `reso_gmbh_de.py` needs no change at all, so it's back to its original state.

I also tightened the pattern slightly — `\s*$` under `(?m)` can swallow line breaks and depends on backtracking to land on a line end, so it's now a `(?=\r?$)` lookahead, and property names match case-insensitively.

On the test: I removed `tests/test_reso_gmbh_de.py`. Per-source test files aren't the convention here (see CLAUDE.md / CONTRIBUTING) — TEST_CASES plus a live `test_sources.py -s <name> -l` run is what we check. More importantly it was never running: `pytest.ini`'s `python_files` allowlist only collects `test_source_components.py`, `test_fetch_retry.py` and `test_ics_timezone.py`, so the file would have sat there looking like coverage while executing nowhere.

Since this is now a shared-service fix, it does deserve a guard, so I've added `tests/test_ics_malformed_timestamps.py` (covering both `convert()` and `convert_events()`, CRLF and LF, plus a well-formed-feed case) and registered it in `python_files`. I confirmed it fails without the fix and passes with it.

`pytest tests/` is green at 165 passed and `pre-commit run --all-files` is clean. Nice catch — thanks for chasing it down and for the clear write-up.

PR #7471, opened 2026-09-19
fix(cidiu_it): read schedules from the Junker app now that cidiu-processer.php is gone

Darkslayer3324j:
## Summary

`cidiu_it` returns `404 Client Error` for every address:

```
404 Client Error: Not Found for url: https://cidiu.it/wp-content/themes/icelander/integrazione-lista/cidiu-processer.php?func=getrecollection&comune=COLLEGNO&...
```

cidiu.it was rebuilt (the `icelander` theme is gone, the site now runs `cindy`) and the calendar endpoint the source scrapes no longer exists. The new site sends people to the Junker app instead (https://cidiu.it/cidiu/junker-app/), and Junker already lists the CIDIU towns: `differenziata.junkerapp.it/collegno/calendario` etc.

## What changed

The source keeps its arguments (`city`, `street`, `street_number`) so existing configurations continue to work, but now:

1. asks the existing `junker_app` service for the town's zone list,
2. picks the zone whose name covers the street and number, and
3. fetches that zone through the same service.

Junker zone names are free text (`VIA CONDOVE da civico 2 a civico 124 e da civico 1 a civico 123`, `CORSO SUSA pari da 2 a 314 dispari da 17 a 315`, `... (tranne civico 51)`, or just `Viale Antonio Gramsci`), so the matching handles number ranges, `pari`/`dispari`, `tranne` exceptions, and falls back to matching on the words given when Junker spells a street out in full (`VIALE GRAMSCI` -> `Viale Antonio Gramsci`). If no zone or more than one zone matches it raises `SourceArgumentNotFoundWithSuggestions` / `SourceArgAmbiguousWithSuggestions` listing the candidates instead of guessing.

`doc/source/cidiu_it.md` is rewritten (its example also used `street_name`, which is not the argument name).

## Type of change

- [x] Bug fix / source fix

## Checklist

- [x] TEST_CASES use real, publicly accessible addresses (the existing three)

Verified live: `test_sources.py -s cidiu_it` went from three 404s to 819 (Collegno), 516 (Grugliasco) and 504 (Rivoli) entries. Added `tests/test_cidiu_it.py` (16 offline tests covering ranges, parity, exceptions, name fallback, ambiguity and the fetch flow).

Caveats: collection type names now come from Junker (`General waste collection`, `Organic waste`, `Paper`, `Plastic`, `Glass/Cans`) instead of the old Italian column headers, so users who customised by type name will need to update. Zone matching uses number ranges only; I could not check every one of the ~1,700 zones by hand.

Found by running every source's TEST_CASES against the live sites; no existing issue or PR for this. Drafted with AI assistance (Claude Code) as this repo's AGENTS.md invites; I reproduced the failure and ran the checks above.


bbr111:
Thanks for the migration: `cidiu-processer.php` is indeed gone and Junker is the right replacement. While reviewing I found a few problems, which I fixed directly in a commit on your branch (`dba50667`):

- **Zone matcher:** zones dedicated to a single house number (`Viale Bruno Radich 11`), streets with `da` or a number in their name (`Via Leonardo da Vinci`, `Piazza 66 Martiri`) and `civici pari` / `dispari` zones (`Via Roma`) were matched wrongly or not at all. `Grugliasco / Viale Bruno Radich / 11` silently returned a completely different calendar. The most specific zone now wins, and I checked this against the live zone lists of Collegno, Grugliasco and Rivoli.
- **No breaking change:** Junker's types are mapped back to the previous labels (`Indifferenziato`, `Organico`, `Carta`, `Plastica`, `Vetro e lattine`). Junker does not publish `Sfalci` (garden waste); the docs now say so.
- **Smaller things:** removed the generic `ValueError` and the unreachable `except` branch, cut the requests per update from three to two, and moved the tests into `tests/test_source_components.py` (the standalone test file was never collected by CI).

Please have a quick look and let me know if this works for you.

Darkslayer3324j:
Thanks for reviewing this and fixing the matcher directly. I checked out `dba50667` and re-ran it:

- `pytest tests/test_source_components.py`: 47 passed.
- `test_sources.py -s cidiu_it` against the live site still gives Collegno 819, Grugliasco 516 and Rivoli 504 entries.
- Spot checks: `Viale Bruno Radich` 11 and 13 now resolve to different calendars (516 vs 501 entries), `Via Roma` 5 and 6 (odd/even) both resolve, and `Via Leonardo da Vinci` 3 works. Types come back as the previous labels (`Indifferenziato`, `Organico`, `Carta`, `Plastica`, `Vetro e lattine`).

You caught real cases my range-only matching got wrong (single-number zones, `da` in a street name, pari/dispari); I had only checked the three test addresses and said I couldn't review every zone. Nothing further from me, this looks good to merge.

(This reply was drafted with AI assistance, as was the original change; the checks above are what I ran.)

Issue #7102, opened 2026-08-02
v3: cassette replay silently accepts a changed POST body


markvp:
Found while reviewing a v3 cleanup campaign migration. This is about the verification method the campaign depends on, so it matters more than any individual source.

## The hole

`tests/cassette.py`'s `_find()` matches a request in two passes:

1. **Exact key** — `f"{METHOD} {url} {body_hash}"`. Correct and strict.
2. **Fallback: method + url only**, with the comment "body may not reproduce byte-for-byte".

A recorded interaction stores `key`, `method`, `url`, `status`, `headers`, `content_b64`, `encoding`. **It does not store the request body.**

So when a refactor changes a POST body, the key changes, pass 1 misses, and pass 2 matches on method and URL alone. The replay serves the recorded response and the test goes green, having verified nothing about the request that was actually sent.

## Why it matters now

The campaign's entire safety argument is:

> The recorded requests and the parsed output must be identical before and after. If replay changes, the refactor changed behaviour and is wrong.

For GET-only sources that holds. For POSTs it does not: the request half of that sentence is unverified whenever the body differs.

Current exposure across the fixture set:

- 247 cassette directories
- 2510 recorded interactions
- **859 POST interactions across 92 sources**

Many campaign migrations move POST-building logic (form state, wizard steps, multipart vs urlencoded) out of a source and into a shared retriever. That is exactly the code path this fallback stops checking.

To be clear: no known migration is wrong because of this. One agent caught it by instinct and hash-compared the outgoing bodies by hand, finding them field-for-field identical. The problem is that the harness does not require that, so the next one may not.

## Suggested fix

Record the request body (or its hash) in the interaction, and make the method+url fallback assert the body hash matches, failing loudly when it does not. If a genuinely non-reproducible body needs tolerating (a nonce, a timestamp), that should be opt-in per source and visible, not the silent global default.

A cheaper interim step: keep the fallback but emit a warning naming the source and URL whenever pass 2 is used, so a body change is at least visible in the test output.

## Related

`tests/cassette.py` gained a third fallback pass in the Athos batch (folding `params` into the URL, for a refactor that moves a hand-built query string onto `params=`). That one is a genuine HTTP equivalence and does not widen this hole, but it is worth reviewing alongside this issue since it touches the same matcher.

markvp:
Two updates from the v3 cleanup campaign, one correcting the scope of this issue.

## It is not POST-only

[...]

PR #7474, opened 2026-09-19
Update Bath and North East Somerset to use BaseSource

trvrnrth:
## Summary

Updates Bath and North East Somerset to use BaseSource.

Adds a new SplitByFields preprocessor to facilitate splitting records that contain collections for multiple dates as I couldn't see a way of doing this with those already available.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other

## Checklist


bbr111:
Small bug in `_extract_uprn`: `source.params("postcode")` will raise a `TypeError` because `source.params` is a plain dict. It should be `source.params.get("postcode")`. This only triggers when a postcode returns no addresses, so none of the TEST_CASES hit it.

bbr111:
@trvrnrth and one thing for the future. set to draft if not ready for review.

trvrnrth:
> Small bug in `_extract_uprn`: `source.params("postcode")` will raise a `TypeError` because `source.params` is a plain dict. It should be `source.params.get("postcode")`. This only triggers when a postcode returns no addresses, so none of the TEST_CASES hit it.

Good catch. I meant for that to be `source.params["postcode"]`.

> @trvrnrth and one thing for the future. set to draft if not ready for review.

I thought it was ready for review until the CI failed ;)

bbr111:
@trvrnrth no prob.

trvrnrth:
Should be good now. Let me know if I've missed any new best practices.

On a side-topic I'd be somewhat keen to drop support for the UPRN lookup as it adds extra complexity that it would be nice to do away with. Are there any more general plans to do that? It feels a bit awkward that various sources separately maintain their own lookup implementations.

bbr111 (approved):
Looks good — approving as-is, no changes needed from my side.

A few things I checked:

- `source.params["postcode"]` in `_extract_uprn` is safe: `alternatives()` builds a non-required ConfigParam and `apply_defaults` guarantees every declared field is a key in `self.params` (None when unset). The function only runs on the lookup path anyway, so `postcode` is populated whenever that line executes.
- `SplitByFields` went into the shared `preprocessors.py` rather than the source module, which is exactly where the reuse rule wants it. It's genuinely new capability too — `DateFields` would drop `featureType` here.
- All three cassettes are present (one per TEST_CASE) and are post-#7102 recordings that pin the request body: running the suite under `-p tests.mutate_requests` correctly fails all three, so the replay actually checks what the source sends.
- `COUNTRY = "uk"` is a nice bonus — the legacy module had no `COUNTRY` at all, so this un-orphans the source from the README / sources.json listings.
- `schedule_url=lambda found_uprn, **_` sidesteps the duplicate-kwarg collision with `**source.params`; good catch.

Verified locally: full `pytest tests/` (8112 passed), `pre-commit run --all-files` all green, and CI is green on the PR.

One note for the release manager rather than for you: this is a legacy→pipeline migration, so it changes the waste-type labels (`Rubbish` → General Waste, etc.). That's a breaking change and it's correctly targeted at `release/3.0.0` to be batched into the major.

PR #7494, opened 2026-09-20
kiedysmieci_info: keep the HTTP status when the envelope reports a failure

### LINKED ISSUE #7493: kiedysmieci_info: a failed envelope is reported without its HTTP status
## What happens

`kiedysmieci_info` reports a failed response without its HTTP status, so a provider outage and a provider refusal are told to the user the same way — and when the body carries no message, the error names neither.

The proxy reports failures inside an `{"ok": ..., "data": ...}` envelope *with* a 4xx/5xx status alongside, which is why the source parses the body before looking at the status. It then discards the status entirely:

```python
if not payload.get("ok"):
    raise Exception(
        payload.get("message") or "Unexpected response from kiedysmieci.info"
    )
```

Measured against each response shape:

| response | reported as | |
|---|---|---|
| 503 + HTML body | `HTTPError: 503 Server Error` | correct |
| 503 + envelope with message | `Service unavailable` | status lost |
| 503 + envelope without message | `Unexpected response from kiedysmieci.info` | **names nothing at all** |
| 200 + envelope without message | `Unexpected response from kiedysmieci.info` | correct — nothing to name |

The third row is the one that matters: a user whose provider is down gets a generic "unexpected response" in the config flow, with nothing to distinguish it from a wrong address or a parsing problem, and nothing to search for.

Note the first row already works: a status-only failure (an error page rather than an envelope) reaches `raise_for_status()` and reports the code. The gap is specifically the envelope path.

## Reproduce

```python
class Resp:
    status_code = 503
    def json(self): return {"ok": False}
    def raise_for_status(self): raise requests.HTTPError("503 Server Error")

# -> Exception("Unexpected response from kiedysmieci.info"), no status
```

## Scope

Present in the shipped source (`source/kiedysmieci_info.py`, `Source._get`) on `master` and `release/3.0.0`, and carried unchanged into `service/KiedySmieci.py` by the pipeline migration in #7448.

Also: the envelope failure raises a bare `Exception`, which the contributing guide asks sources not to do. #7448 already narrows that to a `KiedySmieciError`; on the legacy source it is still `Exception`.

## Suggested fix

Keep both pieces of information, since they answer different questions — what went wrong, and whether the provider failed or refused:

```
503 + "Service unavailable"  ->  "Service unavailable (HTTP 503)"
503 + no message             ->  "HTTP 503"
200 + "Nieznany adres"       ->  "Nieznany adres"
200 + no message             ->  "Unexpected response from kiedysmieci.info"
```

A non-JSON body should keep going to `raise_for_status()`, which already names the status; wrapping an envelope error around it would bury that.

kjanas:
Fixes #7493.

> **Stacked on #7448 — do not review the diff as it stands.** `service/KiedySmieci.py` is created by that PR, so this branch is built on it and the file list here shows all of #7448's commits. The change that belongs to *this* PR is the last commit alone, `kiedysmieci_info: keep the HTTP status when the envelope reports a failure`. I will rebase onto `release/3.0.0` as soon as #7448 merges, after which the diff will be two files. Opening it now so the issue has a fix attached; happy to close and re-open cleanly instead if you would rather not carry a stacked PR.

## The change

The proxy reports failures inside its `{"ok": ..., "data": ...}` envelope with a 4xx/5xx alongside, so `request()` parses the body before looking at the status — and then discarded the status. A 503 carrying a message was reported as the message alone, and a 503 carrying none became a bare "unexpected response" naming neither what went wrong nor that the provider was down.

The message and the status answer different questions — *what* failed, and whether the provider *failed or refused* — so both are kept when both are there:

```
503 + "Service unavailable"  ->  "Service unavailable (HTTP 503)"
503 + no message             ->  "HTTP 503"
200 + "Nieznany adres"       ->  "Nieznany adres"
200 + no message             ->  "Unexpected response from kiedysmieci.info"
```

A non-JSON body still goes to `raise_for_status()`, which already names the status; wrapping an envelope error around that would only bury it.

## Type of change

- [x] Bug fix / source fix

## Testing

`tests/test_kiedysmieci_client.py`, a service-client test in the shape of the existing `test_ecoharmonogram_client.py`: a fake session per response shape, six cases covering the four rows above plus the non-JSON path and a successful envelope. Two of them fail on the current code, four pin behaviour that must not change.

It is added to `python_files` in `pytest.ini`. That list is an allowlist, not a pattern — a new file left out of it is collected by nothing and silently never runs.

`python -m pytest tests/ -q` — 8182 passed, 9 skipped. `pre-commit run` clean on the changed files.

https://claude.ai/code/session_012thbQYtbYEmZH7VS7fj7y9


bbr111 (approved):
Thanks — the fix itself is right, and the four-row table in the issue made it easy to check: message and status answer different questions, a 200 refusal keeps its message alone, and a non-JSON body still goes to raise_for_status() rather than being wrapped. Merging that as-is.

I have dropped `tests/test_kiedysmieci_client.py` and its `pytest.ini` allowlist entry, though. The project's expected coverage for a source or service fix is TEST_CASES plus the recorded cassettes, and a per-service test file does not scale to ~600 sources — `test_ecoharmonogram_client.py` is the exception rather than the pattern, and it earns that by pinning a BOM-only response needing a retry, which no cassette can produce and no reading of the code makes obvious. Six lines of status formatting are not in that class, and the file also hand-rolls a FakeResponse/FakeSession pair beside `tests/cassette.py`, which already models a response's status, `json()` and `raise_for_status()`.

You are right that a cassette cannot reach the failed-envelope path — it records a successful TEST_CASE. That gap applies to every source's error path today, so if we want to close it, the shape is one shared test driving the cassette response object over a table of service clients, as its own change rather than a file per client. Worth opening if you fancy it.

Also: you were right to flag the stacking, and it has sorted itself out — #7448 merged, so the diff here is already the one file. No rebase needed.

Full suite green after the removal: 8176 passed, 9 skipped.

bbr111:
@kjanas thx for you help. if you have time please pick one of the other issues. Will try to release an new 3.0.0 beta today.

Issue #7492, opened 2026-09-20
Audit which sources can move onto cascading_select (one level per view)


bbr111:
#7448 makes the config flow ask for a cascading_select one level per view. Other sources with cascading lookups may benefit, and the wizard/ helpers may be replaceable by it:
- source/gemeinde24_at.py defines `get_choices`, but I did not check whether it declares `cascading_select`.
- waste_collection_schedule/wizard/: abfall_io_graphql, citiesapps_com, cmcitymedia_de, wokingham_uk.

Task: for each, check whether its lookup is a chain of dependent levels that `get_choices(field, selections)` can express, and open a per-source conversion where it is. Anything that needs a search rather than a walk (see the affaldonline_dk issue) stays out.

Sources already on cascading_select: abfall_io, app_abfallplus_de, aw_harburg_de, beachwood_oh_us, stadt_kerpen_de (plus kiedysmieci_info after #7448). Known problems in two of them are tracked in #7488 and #7489.

Related: #7448.

costajohnt:
I went through the list against `release/3.0.0`:

- **gemeinde24_at**: already uses `dependent_select("gemeinde", "strasse")` with `get_parent_choices` / `get_choices(parent_value)`, so it has a working two-level cascade. Moving it to `cascading_select` would only change the declaration, so I'd leave it unless you want to retire `dependent_select`.
- **abfall_io_graphql**: a good fit. `cities(query: "")` returns the provider's full city list (I got 45, 1 and 26 cities for three of the registered keys; a real search like `"a"` returns 2, 0 and 0), so the city step is a walk, not a search. After that it's district (only when `districtsCount > 0`), street, then house number, which matches the "a level that returns `[]` is skipped" rule. Today users have to type an opaque `idHouseNumber` found with the wizard. A cascade can store that same id as the last level's value, so existing configs keep working.
- **citiesapps_com**: the wizard needs an account login (email/phone + password) before it can list calendars, so this doesn't fit a public cascade.
- **cmcitymedia_de**: there's no list of `hpid`s. The wizard brute-forces ids 0 to 1000 against the API, so the first level can't be served by `get_choices`. Realm and district after it would be walkable, but only with a static hpid list.
- **wokingham_uk**: starts from a free-text postcode, so it's a search and stays out, like affaldonline_dk.

[...]

PR #7498, opened 2026-09-21
abfall_io_graphql: pick the address in the config flow with cascading_select

costajohnt:
## Summary

Refs #7492.

The config flow now walks `abfall_io_graphql` addresses as a cascade (city, district, street, house number) instead of asking for a free-text `idHouseNumber` that users had to find with the CLI wizard.

- `cities(query: "")` returns the provider's full city list, so the first level is a list rather than a search. Every registered key returns cities once the region's Origin is sent.
- The platform attaches `idHouseNumber` to whichever entity is deepest for an address: a city, a district, a street, or a house number of a street. Once a level carries it, the levels below return `[]` (skipped by the flow) and the `idHouseNumber` level offers that one id. The cascade therefore always ends with the value the retriever already reads, so existing `key` + `idHouseNumber` configurations keep working unchanged.
- The lookup lives in the shared `service/AbfallIOGraphQL.py` as `list_choices`, next to the retriever, like `service/AbfallIO.py`. The source only declares the `cascading_select` and delegates `get_choices`.
- KELL and Landkreis Böblingen answer 401 without an Origin. The retriever's Origin lookup moves to module level, so the lookups and the retriever share it (retriever behaviour unchanged).
- A GraphQL `errors` payload in a lookup raises instead of reading as an empty level, so the flow falls back to the editable form rather than silently skipping the rest of the cascade. A leftover `idDistrict` that doesn't belong to the chosen city is ignored.
- Since the cascade makes every level optional in the form, the retriever now raises `SourceArgumentRequired` when `idHouseNumber` is missing.
- `wizard/abfall_io_graphql.py` is removed. The config flow does its job now, and on this branch it no longer imported (`SERVICE_MAP` moved out of the service module). Happy to put it back if you'd rather keep it.
- `cascade_confirm.data` in each translations file gets a `wasteTypes` label (reusing the file's `f_abfallarten` wording). `test_the_closing_step_labels_every_field_it_can_show` requires it, and that section is hand-maintained, as in #7495.

Checked live against all 11 registered providers: each walks to an `idHouseNumber` and `fetch()` returns collections. Between them they cover every branch: the city settles the address (Landkreis Göttingen), the district settles it (KELL, ASG Nordsachsen), the street settles it (Märkisch-Oderland, Böblingen), and a real house-number pick (Graz, Essen, Duisburg, Reutlingen, Osterholz).

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Migration of an existing source to the 3.0.0 pipeline
- [ ] Documentation update
- [x] Other (config-flow cascade for an existing pipeline source)

## Checklist

- [ ] No generated files in diff (README.md, info.md, sources.json, translations/*.json, CI regenerates these post-merge). The only translations change is the hand-maintained `cascade_confirm.data` label described above.
- [x] `doc/source/<name>.md` created for new legacy sources (pipeline sources generate theirs)

- [x] **Cassette recorded.** A new TEST_CASE (Osterholz, Lübberstedt, Kampstraße 5) carries all four levels. I recorded its cassette plus `_choices.json` for it only, so the existing cassettes are untouched. The offline fixture covers the four-level path; the other branches were checked live as listed above.
- [x] **Composes shared components.** No retriever or parser in the source module.
- [x] **Anything the platform was missing was added to the shared component**: `list_choices` in `service/AbfallIOGraphQL.py`.

Full `pytest -m "not live"` and `pre-commit` pass locally (in a venv, not the devcontainer).


bbr111:
@costajohnt Thank for you help. will have a closer look later.
but one thing. you don't need to edit the lang-files.json. they are auto-gen by update_docu_links.py
if you need some translations here, we must enable update_docu_links.py to generate it.

bbr111 (approved):
Nice piece of work — this is exactly the shape the `cascading_select` widget was meant to take, and it lines up with how `abfall_io` does it: the lookup lives in the shared `service/` module next to the retriever, and the source just declares the levels and delegates `get_choices`. No retriever or parser in the source module, nothing added to a reuse allowlist.

One correction to the earlier comment about the translation files — you were right to change them. `update_docu_links.py` only rewrites the `args_*` and `reconfigure_*` step keys; `config.step.cascade_confirm` is hand-maintained, and `test_the_closing_step_labels_every_field_it_can_show` requires a label for every non-cascade param a cascading source can put on that closing form. Your cascade ends on `cascade_confirm` showing `calendar_title`, `key` and `wasteTypes`, so without your `wasteTypes` label CI fails and the field renders as its raw name. Please leave it in.

Verified locally: `pytest tests/test_source_components.py` (47 passed), the config-flow / new-architecture / offline-fixture suites (7695 passed, 9 skipped), and `pre-commit run --all-files` all green. Running the offline replay with `-p tests.mutate_requests` fails all 14 `abfall_io_graphql` cases, which confirms both new recordings pin their request bodies rather than matching on the URL alone.

Removing the wizard is fine — it imported `SERVICE_MAP`, which no longer exists in the service module, so it could not import on this branch.

Backward compatibility looks right too: the retriever still reads only `key` + `idHouseNumber`, and the new TEST_CASE is appended after the two-arg one, so the generated `doc/source/abfall_io_graphql.md` keeps showing the minimal YAML example.

Approving. (`sl.json` also lacks `cascade_confirm`, but that predates this PR; tracked separately.)

Issue #7491, opened 2026-09-20
affaldonline_dk: let users enter an address instead of looking up the parameter values by hand


bbr111:
Requested in the discussion on #7448 (@superrob): for affaldonline_dk the user currently has to find the correct parameter values manually. It would be better to enter an address and have the source resolve the values.

This does not fit `cascading_select`: as far as I understand from that thread the flow is a search by address rather than a walk down dependent levels, and the author of #7448 suggests `multi_value_lookup` for it. Someone should confirm that against the provider's API before implementing.

Not part of #7448, which only covers cascades.

Related: #7448.

superrob:
What i would love to have is a feature in config_flow, where the Source can lookup a search field (Eg. Address) and save the results in the config.

The current way would be to let the user input their address (In for example a multi_value_lookup). This can work fine enough, but would result in a search on the providers API every time the user reloads HA or the integration, which seems like a bad idea. 
We don't want to give them a reason to block/restrict the API, which they easily could do.

I had created a discussion of this idea earlier: #7457

bbr111:
@superrob see answer in #7457

PR #7519, opened 2026-09-21
Affaldonline basesource reimp

superrob:
## Summary

This PR reworks the affaldonline_dk source into the new BaseSource structure.
To facilitate this the service file AffaldOnlineDk.py is added, which contains the retriever, parser and cascading_select logic.

The source no longer requires the user to fiddle around to manually find the affaldonline identifying values. Instead the source now implements the cascading_select flow as mentioned in #7491 

This is implemented by letting the user select their city, then street, finally the house number which carries the identifying affaldonline values.
The house number parameter is currently kept as "values" to not break existing installations completely.

Regarding the choice of not using WasteType.
Waste seperation in Denmark is mandetory to be atleast seperated into these 10 fractions:
Food Waste
Paper
Cardboard
Plastic
Food and drink cartons
Metal
Glass
Textiles
Hazardous waste
Residual waste

Only 6 out of the 10 types currently have a WasteType, the following are missing:
Cardboard, Plastic, Food and drink cartons, Metal

These 10 fractions are usually combined into a few bins for collection, with one or two compartments.
Some fractions are additionally allowed to be combined into the same compartment. As such some bins have 4 different fractions combined.

This makes the WasteType approach highly impractical, as some bins would end up with the same WasteType, even though it is a completely different bin and content.
For example take these two bins:
Metal, Glass, Paper & Cardboard
Plastic & Food and drink cartons

Both of these would under the current WasteType system end up as WasteType.RECYCLING, making for a confusing UX.

Instead the source currently uses the older direct Collection system, which offers more flexibility.
The API contains the exact fractions from each collection, and thus is able to dynamically name every collection based on the fraction content.

The user can if desired choose to split each collection into separate collections, one for each fraction. This is achieved by the boolean "split_bins" during config_flow.

## Type of change

- [ ] New source
- [ ] Bug fix / source fix
- [ ] Documentation update
- [x] Other - Reworked to BaseSource and Cascading select

## Checklist


bbr111:
please solve codeignore inline

superrob:
I am guessing you meant the **# type: ignore[attr-defined]**?
Was simply a leftover from the old scraping source. It looks to have effectively done nothing.

bbr111 (changes requested):
Thanks for this — the componentisation is exactly right: retrieval, parsing and the cascade lookup all live in `service/AffaldOnlineDk.py`, so both reuse gates pass with no allowlist entry, and you've shipped a cassette for every one of the 16 test cases. The cascading city → street → house-number flow is a big usability win over the old "open devtools and copy the values string" instructions.

I've pushed a few mechanical fixes to your branch:

- `COUNTRY = "dk"` was missing — without it all 15 regions are silently orphaned out of the README and `sources.json`.
- `regions.from_yaml(..., title_suffix="affaldonline")` — `title_suffix` names a *key in the YAML entry*, so that was a no-op, and with no `**param_fields` mapping every Region came out with `params={}`. Now `municipality="municipality"`, which pre-fills all 15 listings.
- `doc/regions/affaldonline_dk.yaml` had `Middelfart` capitalised, so it matched neither `CLIENT_ID_LOOKUP` nor the lowercase URL path segment.
- `tests/test_config_flow.py` was failing: `cascade_confirm.data` needs hand-added labels for `municipality` and `split_bins` in all five locales (that section is hand-maintained; only `args_*` is generated).
- The `aeroe` test case's `values` was stale from the old scraping source and isn't in the house-number list the recorded response returns, which failed `test_offline_choices`. Updated it (and the `_choices` fixture) to `…|4162588|448776|0`; the fetch cassette is unaffected because only the last two segments become the `addressId`. The `holbaek` case also had `city: "Broby"`.
- Exceptions: `ValueError` → `SourceArgumentException`, and the municipality guards now use `SourceArgumentRequiredWithSuggestions` / `SourceArgumentNotFoundWithSuggestions` so the UI lists the supported municipalities. Plus `Retriver` → `Retriever`, a sorted city dropdown, an `IndexError` guard in `classify()` when a bin has no fraction ids, and some dead code removed.
- Reverted `doc/source/affaldonline_dk.md`. For `BaseSource` sources that file is rendered wholesale from the class metadata post-merge (`doc_generator.render_source_doc`), so hand-edits get overwritten. See the `HOWTO` note below.

Two things I'd like you to take on before this can go in:

**1. The waste-type shape (currently 17 failing tests).** I follow your reasoning in the description, and you're right that there's no canonical `METAL`, `PLASTIC`, `CARDBOARD` or `CARTON`, so a combined Danish bin would collapse onto `RECYCLABLES`. But the legacy `Collection(t=..., icon=...)` shape isn't the way out of that inside a pipeline source: the compat shim in `collection.py` turns it into `WasteType(id=f"legacy_{t}", names={"en": t})`, so the Danish label is registered as the *English* name and nothing else, and the resulting `legacy_*` ids fail `tests/test_declared_waste_types.py` (all 16 cases) and `test_has_waste_types`.

The sanctioned escape hatch for a label the vocabulary can't express is `waste_types.preserved(label)`: it keeps your text verbatim in every locale, gives it a `preserved:` id, and the gate explicitly ignores those. `bsr_de.py` is the worked example.

What I'd suggest: resolve single-fraction bins to the canonical type where one exists (`Madaffald` → `FOOD_WASTE`, `Papir` → `PAPER`, `Glas` → `GLASS`, `Restaffald` → `GENERAL_WASTE`, `Tekstiler` → `TEXTILES`, `Farligt affald` → `HAZARDOUS`, `Elektronik` → `ELECTRONICS`, `Haveaffald` → `GARDEN_WASTE`, `Storskrald` → `BULKY_WASTE`), declare exactly those in `WASTE_TYPES`, and use `preserved("Plast, Kartoner og Metal")` for the combined bins where your dynamic naming genuinely carries information the vocabulary can't. `FRACTION_MAP` then becomes a fraction-id → `WasteType` map and the per-fraction `Icons` entries go away — a pipeline source takes its icon from the resolved type. The one cost is that `preserved()` borrows `OTHER`'s icon, so combined bins lose the bespoke icon; if that's a blocker for you, say so and let's discuss rather than you guessing.

**2. Add a `HOWTO`.** Since the doc page is generated, everything you wrote in `doc/source/affaldonline_dk.md` disappears on the next generation run. Move the per-municipality links and the "how to find the values string" guidance into a `HOWTO: ClassVar[dict]` on the class (see `bsr_de.py`); the generator renders it into the doc page *and* the config form shows it.

One small optional thing: `discover_choices()` builds its own `curl_cffi` session while the retriever uses `source.session` — they could share one.

Flagging for the release manager: this is a legacy → pipeline migration, so it changes waste-type labels and is a **breaking (major)** change. Correctly targeted at `release/3.0.0`.

bbr111:
Since this was the only thing blocking CI, I went ahead and implemented the waste-type fix myself so this doesn't need another round-trip:

**`FRACTION_MAP` now maps each fraction id to a canonical `WasteType`** instead of an `Icons` member. Turns out the "4 missing types" premise wasn't quite right: `PAPER`'s English name is already `"Paper & Cardboard"`, and `RECYCLABLES` covers the mixed plastic/metal/carton packaging stream - so all 18 known fraction ids resolve to one of 10 canonical types, no gaps.

**`classify()`:**
- When every fraction in a bin resolves to the *same* canonical type, it's used directly (no information lost, e.g. a pure "Plast, Kartoner og Metal" bin now cleanly becomes `recyclables`).
- When a bin combines fractions that resolve to *different* canonical types (e.g. "Restaffald og Madaffald" mixes `general_waste` and `food_waste`), it falls back to `waste_types.preserved()` with your original composite Danish label verbatim - collapsing that onto either type would misrepresent what's actually being collected.

`WASTE_TYPES` is now derived from `FRACTION_MAP` itself, so it can't drift from what `classify()` actually produces.

One trade-off from this, flagged in case it matters to you: single-fraction bins for "Plast", "Metal", "Kartoner" and "Drikkedåser" now all share `RECYCLABLES`'s generic icon (`mdi:recycle`) instead of their previous distinct icons - a pipeline source's icon always comes from the resolved `WasteType`, so per-fraction icons aren't available anymore. Say if you'd rather handle this differently.

Verified: all 4 previously-failing suites are green now (`test_declared_waste_types.py` 16/16, `test_new_architecture.py -k affaldonline` 21/21, `test_offline_fixtures.py` 17/17, `test_source_components.py` 47/47), plus a manual replay of all 16 cassettes to eyeball the actual output.

The `HOWTO` addition from my earlier review is still open and non-blocking - happy to take that on too if you'd like, or feel free to add it yourself.

superrob:
Hi! First of all thanks for looking into this and changing a few things around :)!

Unfortunately we are still missing the WasteTypes for a few fractions, as the current implementation sadly ends up being highly confusing in many cases.
Take my own bins as an example. I have 3 bins with these fractions in them:
General waste & Food waste
Metal, Glass, Cardboard & Paper
Plastic & Drink-cartons.

Under the latest commit these will now resolve into the following:
image
As you can see the Plastic & Drink cartons bin ends up simply being named "Recycling".
This is quite troublesome as the definition of the Recycling waste type is way too broad and can end up producing the "Recycling" WasteType and end up consolidating two different bins into the same sensor.

This is also a problem when using the split bins feature. In my own case Plastic, Metal and Drink-cartons now end up as "Recycling" under the same sensor, however they live in two different bins for collections, so the sensor is essentially useless at that point for knowing what bin to actually put outside.
image

"Recycling" is currently a quite broad catchall WasteType

The most optimal solution would be the addition of a Metal, Plastic and a Drink-cartons WasteType.
There is already presidence for this, as Glass is a separate type and could also have been put into Recycling as well. 

Paper and cardboard is a type that works just fine being consolidated for this source. I am not aware of any municipalities that choose to not combine those in the same bin/compartment. So functionally they are the same.

I hadn't noticed wt.preserved, however one note is that it ends up giving a mix of danish and english labels. This could be solved by adding a danish translation to the defined WasteTypes. Or translating into danish overall (I would not be against doing that)

Regarding **HOWTO** i hadn't had time to look into this yet :)!

bbr111:
Thanks for the detailed explanation — your example makes the problem with collapsing Plastic, Metal and Drink-cartons into the generic RECYCLABLES type quite clear.

I think adding dedicated WasteTypes for these fractions is worth discussing separately, as this would also be useful beyond this specific source.

For now, I'll convert this PR back to draft so there is no pressure to finish it immediately. Feel free to take your time with the remaining changes, including the HOWTO, and mark it ready for review again once you're happy with it.

No rush from my side — I'm happy to wait until you're ready.

superrob:
I have made a temporary stopgap change to use preserved types for the three missing types.
The ambition is to get them implemented into WasteType's. Doing so could also allow the source to use the WasteType names for the waste labels on the combined bins, which would allow for translation, instead of simply being in Danish.

HOWTO - I am unsure what to write here. The old **values** variable is no longer something the user will have to find themselves and the instructions are no longer needed. 
It could simply state that the user will have to select their city, street and then house number? But it seems quite self explanatory when going through the config flow.

The best outcome for this source is to get the three new WasteTypes added, along with adding danish to the integration. I might try and do this.

I don't know if the addition of the above is preferred before moving forwards with this pr?

bbr111:
Thanks, that makes sense.

Regarding the `HOWTO`: I agree with you. With the new cascading city → street → house number flow, the old instructions for manually finding `values` are no longer needed, and I don't think we should add a HOWTO just for the sake of having one. The config flow should be self-explanatory enough here.

For the missing WasteTypes, I think adding dedicated types for Plastic, Metal and Drink cartons would be the cleaner long-term solution. Since those types are not specific to AffaldOnline and could be useful for other sources as well, I would prefer to handle that in a separate PR.

If you would like to work on that, feel free to do so. We can keep this PR in draft in the meantime and then replace the temporary `preserved()` mappings once the new WasteTypes are available.

Adding Danish translations would also be a nice improvement, but I don't think we need to make that a hard requirement for this PR unless it becomes necessary for the WasteType changes.

So no need to rush this one — I'd rather get the underlying WasteTypes right than merge the temporary solution and change it again shortly afterwards.

superrob:
The goal is 100% to get dedicated WasteTypes implemented for the three types. That will be the most clean implementation.

I have already been working on the danish translation in #7549
With that translation in place, the source could instead of using hardcoded danish strings, use the WasteType localized name for the composite labels. That would greatly improve the overall finish, as currently a user could experience to bins with danish names and one with a english label.

That i would really like to also solve 😊

